PR #4, opened 2018-11-02
CORCI-502 checkPatch Groovy only for Production

JohnMalmberg:
* Add groovydoc
* Require env.DAOS_JENKINS_NOTIFY_STATUS for an actual update.
* Lookup the credentials for the GITHUB upddate instead of having
  them passed.
* Provide a default empty Map as a parameter.

### DIFF
--- a/vars/checkPatch.groovy
+++ b/vars/checkPatch.groovy
@@ -1,13 +1,33 @@
 // vars/checkPatch.groovy
 
+/**
+ * checkPatch.groovy
+ *
+ * checkPatch variable
+ */
+
 import com.intel.checkoutScm
 
-def call(Map config) {
+/**
+ * Method to check the status of a patch and notify GitHub of the results
+ * 
+ * @param config Map of parameters passed.
+ *
+ * config['jenkins_review'] is the optional revision number to use.
+ *
+ */
+def call(Map config = [:]) {
+    if (env.DAOS_JENKINS_NOTIFY_STATUS == null) {
+      println "Jenkins not configured to notify github of builds."
+      stepResult name: env.STAGE_NAME, context: "pre-build",
+                 result: "SUCCESS"
+      return
+    }
 
     def c = new com.intel.checkoutScm()
     c.checkoutScmWithSubmodules()
 
-    if(config['jenkins_review']) {
+    if (config['jenkins_review']) {
         rev_num = config['jenkins_review']
         branch = 'FETCH_HEAD'
         refspec = 'refs/changes/' + rev_num
@@ -35,12 +55,15 @@ def call(Map config) {
                  description: env.STAGE_NAME,
                  context: "pre-build" + "/" + env.STAGE_NAME,
                  status: "PENDING"
-
-    script = 'GH_USER=' + config['user'] + \
-             ' GH_PASS=' + config['password'] + \
-             ' jenkins/code_review/jenkins_github_checkwarn.sh'
-    rc = sh(script: script, returnStatus: true)
-
+    withCredentials([[$class: 'UsernamePasswordMultiBinding', 
+                    credentialsId: 'aa4ae90b-b992-4fb6-b33b-236a53a26f77',
+                    usernameVariable: 'GITHUB_USR',
+                    passwordVariable: 'GITHUB_PSW']]) {
+      script = 'GH_USER="' + GITHUB_USR + '"' + \
+               ' GH_PASS="' + GITHUB_PSW + '"' + \
+               ' jenkins/code_review/jenkins_github_checkwarn.sh'
+      rc = sh(script: script, returnStatus: true)
+    }
     // All of this really should be done in the post section of the main
     // Jenkinsfile but it cannot due to
     // https://issues.jenkins-ci.org/browse/JENKINS-39203
--- a/vars/checkPatch.groovy
+++ b/vars/checkPatch.groovy
@@ -65,11 +65,9 @@ def call(Map config = [:]) {
     if (config['review_creds']) {
       withCredentials([[$class: 'UsernamePasswordMultiBinding', 
                       credentialsId: config['review_creds'],
-                      usernameVariable: 'GITHUB_USR',
-                      passwordVariable: 'GITHUB_PSW']]) {
-        script = 'GH_USER="${GITHUB_USR}"' + \
-                 ' GH_PASS="${GITHUB_PSW}"' + \
-                 ' CHECKPATCH_IGNORED_FILES="' + ignored_files + '"' + \
+                      usernameVariable: 'GH_USER',
+                      passwordVariable: 'GH_PASS']]) {
+        script = 'CHECKPATCH_IGNORED_FILES="' + ignored_files + '"' + \
                  ' jenkins/code_review/jenkins_github_checkwarn.sh'
         rc = sh(script: script, returnStatus: true)
       }
--- a/vars/checkPatch.groovy
+++ b/vars/checkPatch.groovy
@@ -45,7 +45,6 @@ def call(Map config = [:]) {
     checkout([
         $class: 'GitSCM',
         branches: [[name: branch]],
-        doGenerateSubmoduleConfigurations: false,
         extensions: [[$class: 'RelativeTargetDirectory',
                               relativeTargetDir: 'jenkins']],
         submoduleCfg: [],
@@ -62,21 +61,20 @@ def call(Map config = [:]) {
                  status: "PENDING"
 
     rc = 1
+    script = 'CHECKPATCH_IGNORED_FILES="' + ignored_files + '"' + \
+             ' jenkins/code_review/jenkins_github_checkwarn.sh'
+
     if (config['review_creds']) {
       withCredentials([[$class: 'UsernamePasswordMultiBinding', 
                       credentialsId: config['review_creds'],
                       usernameVariable: 'GH_USER',
                       passwordVariable: 'GH_PASS']]) {
-        script = 'CHECKPATCH_IGNORED_FILES="' + ignored_files + '"' + \
-                 ' jenkins/code_review/jenkins_github_checkwarn.sh'
         rc = sh(script: script, returnStatus: true)
       }
     } else {
-      // Older method to be deprecated
-      script = 'GH_USER="' + usr + '"' + \
-        ' GH_PASS="' + psw + '"' + \
-        ' CHECKPATCH_IGNORED_FILES="' + ignored_files + '"' + \
-        ' jenkins/code_review/jenkins_github_checkwarn.sh'
+      // Older method using username/password
+      script = 'GH_USER="' + config['user'] + '"' + \
+        ' GH_PASS="' + config['password'] + '"' + script
       rc = sh(script: script, returnStatus: true)
     }
 
--- a/vars/checkPatch.groovy
+++ b/vars/checkPatch.groovy
@@ -8,14 +8,6 @@
 
 import com.intel.checkoutScm
 
-private def run_review(usr, psw, ignored_files) {
-  script = 'GH_USER="' + usr + '"' + \
-    ' GH_PASS="' + psw + '"' + \
-    ' CHECKPATCH_IGNORED_FILES="' + ignored_files + '"' + \
-    ' jenkins/code_review/jenkins_github_checkwarn.sh'
-    return sh(script: script, returnStatus: true)
-}
-
 /**
  * Method to check the status of a patch and notify GitHub of the results
  * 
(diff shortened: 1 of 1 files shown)


vars/checkPatch.groovy
```diff
@@ -1,13 +1,33 @@
...
+/**
+ * Method to check the status of a patch and notify GitHub of the results
+ * 
+ * @param config Map of parameters passed.
+ *
+ * config['jenkins_review'] is the optional revision number to use.
+ *
+ */
+def call(Map config = [:]) {
+    if (env.DAOS_JENKINS_NOTIFY_STATUS == null) {
```
brianjmurrell: So we should set the global property before we land this, yes?

vars/checkPatch.groovy
```diff
@@ -1,13 +1,33 @@
...
+      println "Jenkins not configured to notify github of builds."
+      stepResult name: env.STAGE_NAME, context: "pre-build",
+                 result: "SUCCESS"
```
brianjmurrell: If we don't want to notify GitHub why even call `stepResult` here?

vars/checkPatch.groovy
```diff
@@ -35,12 +55,15 @@ def call(Map config) {
...
                  context: "pre-build" + "/" + env.STAGE_NAME,
                  status: "PENDING"
-
-    script = 'GH_USER=' + config['user'] + \
-             ' GH_PASS=' + config['password'] + \
-             ' jenkins/code_review/jenkins_github_checkwarn.sh'
-    rc = sh(script: script, returnStatus: true)
-
+    withCredentials([[$class: 'UsernamePasswordMultiBinding', 
+                    credentialsId: 'aa4ae90b-b992-4fb6-b33b-236a53a26f77',
```
brianjmurrell: I'm not sure I like embedding the `credentialsId` in here vs. passing the username/password.  It feels like it's removing some of the generic/portability of this function.

Yes, we do have the Gerrit credentials above also, but that is just temporary until we get the `code_review/` scripts out onto GitHub.

vars/checkPatch.groovy
JohnMalmberg: Not sure.
In the Jenkinsfile I am also calling stepResult.
Maybe we need some logic in stepResult also.
There should be ways like environment variables to tell us how the run was triggered.

vars/checkPatch.groovy
JohnMalmberg: Yes on Jenkins-3 for now and then switch it for production.
I think that all build agents need a restart to pick up the change.

vars/checkPatch.groovy
brianjmurrell: My point is that this is a generic library function.  There is no reason that this function needs to know anything about Jenkins credentials.  It simply needs the GitHub account credentials which can be from a Jenkins credential or otherwise.

Using a Jenkins credential here makes this otherwise generic function very specific to the way we are using Jenkins.

vars/checkPatch.groovy
brianjmurrell: I've added that to _jenkins-3_.

vars/checkPatch.groovy
brianjmurrell: This seems unresolved at this point.

vars/checkPatch.groovy
JohnMalmberg: And my point is that if we resolve fetch the credentials in the Jenkinsfile, particularly as environment variables, they have a larger chance of accidentally leaking into log file.

So either this method needs to have a credential ID with a generic name stored in it, or it needs to be called with a credential ID as a parameter.

We should not ever be passing in usernames and passwords to library methods.

Since you do not want to put a specific credential ID in this library, that leaves passing the credential ID.

brianjmurrell: But we already know that Jenkins masks these account values.

```
[checkpatch] Running script: GH_USER="****" GH_PASS="****" CHECKPATCH_IGNORED_FILES="code_review/checkpatch.pl::src/control/vendor/*" jenkins/code_review/jenkins_github_checkwarn.sh
[Pipeline] [checkpatch] sh
[checkpatch] [daos-stack-org_daos_PR-13-XSYVVAKVRRCIRZBUBQLEPYMVXQHYAQV52HHPQR55EURLM5NXUHMA@2] Running shell script
[checkpatch] + GH_USER=****
[checkpatch] + GH_PASS=****
[checkpatch] + CHECKPATCH_IGNORED_FILES='code_review/checkpatch.pl::src/control/vendor/*'
[checkpatch] + jenkins/code_review/jenkins_github_checkwarn.sh
```

JohnMalmberg: Decoding the credentials into global environment variables is a bad practice.

They are masked there in the Jenkins log, but the environment variables are globally accessible to all steps in the Jenkinsfle and could get dumped in logs/artifacts that Jenkins does not know how to censor.

Decoded credentials should be restricted only to location where they are used as much as possible.

If a library method needs a password or equivalent token, we should pass it the credential id for it to get the access.

JohnMalmberg: I just noticed that the 'daos-jenkins-commit-status' ID is also present in this module.

Does the  'daos-jenkins-commit-status' credential ID used to update the status also have the ability to annotate the code?  If so, then there is a simple solution just to use it.

brianjmurrell: But there is no need for the library function to need to understand or work with Jenkins credentials rather than the more generic form that a credential will yield.  Using the credential id makes the library function less flexible to other applications.

Have you tested an "env" dump to a(n artifact) file to see if the values are masked or not?

JohnMalmberg: No, but that is just one of many ways that a password could accidentally leak out.  It could be difficult to verify all cases, so isolation is the best practice.

With Jenkins traditional jobs, you have to make the environment variables with passwords visible to the build steps.

The Jenkins pipeline provides the method that is designed to only make the sensitive environment variables visible where they are used, so it makes sense to use it and use it that way.

vars/checkPatch.groovy
```diff
@@ -41,11 +61,24 @@ def call(Map config) {
...
+      withCredentials([[$class: 'UsernamePasswordMultiBinding', 
+                      credentialsId: config['review_creds'],
+                      usernameVariable: 'GH_USER',
+                      passwordVariable: 'GH_PASS']]) {
+        script = 'CHECKPATCH_IGNORED_FILES="' + ignored_files + '"' + \
+                 ' jenkins/code_review/jenkins_github_checkwarn.sh'
+        rc = sh(script: script, returnStatus: true)
+      }
+    } else {
+      // Older method to be deprecated
```
brianjmurrell: Or can be left for cases where somebody doesn't want to use a Jenkins credential.

vars/checkPatch.groovy
```diff
@@ -41,11 +61,24 @@ def call(Map config) {
...
+      script = 'GH_USER="' + usr + '"' + \
+        ' GH_PASS="' + psw + '"' + \
```
brianjmurrell: Where are usr and psw defined?

vars/checkPatch.groovy
```diff
@@ -41,11 +61,24 @@ def call(Map config) {
...
-             ' jenkins/code_review/jenkins_github_checkwarn.sh'
-    rc = sh(script: script, returnStatus: true)
+    rc = 1
+    if (config['review_creds']) {
+      withCredentials([[$class: 'UsernamePasswordMultiBinding', 
+                      credentialsId: config['review_creds'],
+                      usernameVariable: 'GH_USER',
+                      passwordVariable: 'GH_PASS']]) {
+        script = 'CHECKPATCH_IGNORED_FILES="' + ignored_files + '"' + \
+                 ' jenkins/code_review/jenkins_github_checkwarn.sh'
```
brianjmurrell: This should be defined outside of this block, up before line 63, perhaps, then it can be used here without modification and can be re-used in lines 76-79 by simply prefixing it with the two additional env. variables.

vars/checkPatch.groovy
```diff
@@ -41,11 +61,24 @@ def call(Map config) {
                  context: "pre-build" + "/" + env.STAGE_NAME,
                  status: "PENDING"
 
-    script = 'GH_USER="' + config['user'] + '"' + \
-             ' GH_PASS="' + config['password'] + '"' + \
-             ' CHECKPATCH_IGNORED_FILES="' + ignored_files + '"' + \
-             ' jenkins/code_review/jenkins_github_checkwarn.sh'
-    rc = sh(script: script, returnStatus: true)
+    rc = 1
```
brianjmurrell:
```suggestion
    script = 'CHECKPATCH_IGNORED_FILES="' + ignored_files + '"' + \
￼              ' jenkins/code_review/jenkins_github_checkwarn.sh'
    rc = 1
```

vars/checkPatch.groovy
```diff
@@ -41,11 +61,24 @@ def call(Map config) {
...
+    if (config['review_creds']) {
+      withCredentials([[$class: 'UsernamePasswordMultiBinding', 
+                      credentialsId: config['review_creds'],
+                      usernameVariable: 'GH_USER',
+                      passwordVariable: 'GH_PASS']]) {
+        script = 'CHECKPATCH_IGNORED_FILES="' + ignored_files + '"' + \
```
brianjmurrell:
```suggestion
```

vars/checkPatch.groovy
brianjmurrell:
```suggestion
```

vars/checkPatch.groovy
```diff
@@ -41,11 +61,24 @@ def call(Map config) {
...
+                 ' jenkins/code_review/jenkins_github_checkwarn.sh'
+        rc = sh(script: script, returnStatus: true)
+      }
+    } else {
+      // Older method to be deprecated
+      script = 'GH_USER="' + usr + '"' + \
```
brianjmurrell:
```suggestion
      script = 'GH_USER="' + config['user'] + '"' + \
```

vars/checkPatch.groovy
brianjmurrell:
```suggestion
        ' GH_PASS="' + config['password'] + '"' + \
```

vars/checkPatch.groovy
```diff
@@ -41,11 +61,24 @@ def call(Map config) {
...
+        ' GH_PASS="' + psw + '"' + \
+        ' CHECKPATCH_IGNORED_FILES="' + ignored_files + '"' + \
```
brianjmurrell:
```suggestion
        ' ' + script
```

vars/checkPatch.groovy
```diff
@@ -41,11 +61,24 @@ def call(Map config) {
...
+        ' jenkins/code_review/jenkins_github_checkwarn.sh'
```
brianjmurrell:
```suggestion
```

brianjmurrell:
As for the _suggested changes_ beta feature I used in my review above, it seems that [changes to multiple lines at once is one of the identified short comings they are planning to fix](https://blog.github.com/2018-11-01-suggested-changes-update/).

vars/checkPatch.groovy
JohnMalmberg: Bug I introduced when I was trying some different refactoring.

vars/checkPatch.groovy
JohnMalmberg: Done on latest patch pushed.

vars/checkPatch.groovy
```diff
@@ -41,11 +60,23 @@ def call(Map config) {
...
+
+    if (config['review_creds']) {
+      withCredentials([[$class: 'UsernamePasswordMultiBinding', 
+                      credentialsId: config['review_creds'],
+                      usernameVariable: 'GH_USER',
+                      passwordVariable: 'GH_PASS']]) {
+        rc = sh(script: script, returnStatus: true)
+      }
+    } else {
+      // Older method using username/password
```
brianjmurrell:
```suggestion
      // Alternate method using username/password
```

vars/checkPatch.groovy
```diff
@@ -41,11 +60,23 @@ def call(Map config) {
...
+      script = 'GH_USER="' + config['user'] + '"' + \
+        ' GH_PASS="' + config['password'] + '"' + script
```
brianjmurrell: I think you need a space before `+ script`
```suggestion
        ' GH_PASS="' + config['password'] + '" ' + script
```

vars/checkPatch.groovy
```diff
@@ -25,7 +45,6 @@ def call(Map config) {
     checkout([
         $class: 'GitSCM',
         branches: [[name: branch]],
-        doGenerateSubmoduleConfigurations: false,
```
brianjmurrell: Why did you need to remove this?

JohnMalmberg: Simplification.
https://issues.jenkins-ci.org/browse/JENKINS-43308

vars/checkPatch.groovy
JohnMalmberg: Fixed in next push.

vars/checkPatch.groovy
JohnMalmberg: Changed in next push.

PR #9, opened 2018-12-03
CORCI-518 Parameterized sconsBuild step.

JohnMalmberg:
* Add groovydoc
* Require env.DAOS_JENKINS_NOTIFY_STATUS for an actual update.
* Provide a default empty Map as a parameter.
* Allow commonly used scons_local paramaters to be specified.
* Default to failing build for declarative pipeline and optionally
  returning a status.


JohnMalmberg:
Scons-local PR 1  build 14 tested this PR.

vars/sconsBuild.groovy
```diff
@@ -2,49 +2,201 @@
...
-                    echo "$SCONS_ARGS failed"
-                    cat config.log || true
-                    exit \$rc
-                fi'''
-    rc = sh(script: script, returnStatus: true)
-
+                '''
+    script += clean_cmd
+    script += 'SCONS_ARGS="' + scons_args +'"\n'
+    script += '''# the config cache is unreliable so always force a reconfig
```
brianjmurrell: Is there any particular reason to keep adding to `script` instead of just continuting the multi-line string (albeit with """ instead of ''' so that you get variable interpolation)?

vars/sconsBuild.groovy
```diff
@@ -2,49 +2,201 @@
...
+                 # with "--config=force"
+                 if ! scons --config=force $SCONS_ARGS; then
+                     rc=\${PIPESTATUS[0]}
+                     echo "$SCONS_ARGS failed"
+                     cat config.log || true
```
brianjmurrell: This doesn't work for projects that use an _arch_.  I have a patch to pipeline-lib for _arch_-using projects and a patch for daos to make it an _arch_-using project but it seems the concenus is actually in the other direction, to remove the _arch_ support from iof and cart.  I'm just not there yet.

vars/sconsBuild.groovy
```diff
@@ -2,49 +2,201 @@
...
-    githubNotify credentialsId: 'daos-jenkins-commit-status',
-                 description: env.STAGE_NAME,
-                 context: "build" + "/" + env.STAGE_NAME,
-                 status: "PENDING"
+def call(Map config = [:]) {
+  /**
+   * sconsBuild step method
+   *
+   * @param config Map of parameters passed
+   * @return Integer status of script
```
brianjmurrell: Trailing whitespace.

vars/sconsBuild.groovy
```diff
@@ -2,49 +2,201 @@
...
+   * config['returnStatus'] return a status code.  Default false.
+   * config['scm'] are the parameters for the scm to use.
+   * config['sconstruct'] Path to sconstruct file if specified.
+   * config['scons_args'] are optional additional arguments for scons
+   *  after all other options are set.
+   * config['scons_local_replace'] replace scons_local in config['target']
+   *  with a symlink.
+   * config['SRC_PREFIX'] Directories to find the prefetched source code in.
+   *  Defaults to config['target'] if config['target'] is specified.
+   * config['target'] Target if only one component is being built.
```
brianjmurrell: Trailing whitespace.

vars/sconsBuild.groovy
```diff
@@ -2,49 +2,201 @@
 
 import com.intel.checkoutScm
```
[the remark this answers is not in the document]
JohnMalmberg: Should not be.

vars/sconsBuild.groovy
JohnMalmberg: Was not sure about getting string interpolation exactly right with quoting.

vars/sconsBuild.groovy
JohnMalmberg: I do not recall making a change that would affect projects using an arch except for adding a wildcard for the cleanup.

vars/sconsBuild.groovy
JohnMalmberg: Will be fixed in next update.

vars/sconsBuild.groovy
JohnMalmberg: Will be fixed in the next update.

vars/sconsBuild.groovy
```diff
@@ -2,49 +2,201 @@
...
+    }
+    if (config['REQUIRES']) {
+        scons_args += " REQUIRES=${config['REQUIRES']}"
+    } else if (config['target']) {
+        scons_args += " REQUIRES=${config['target']}"
+    }
+    //scons -c is not perfect so get out the big hammer
+    clean_files = "_build.external* install build"
+    if (config['clean']) {
+        clean_files = config['clean']
```
brianjmurrell: This implementation of `config['clean']` is conflicting with an existing use of `config['clean']` in [sconsBuild-clean](https://github.com/daos-stack/pipeline-lib/commit/62dd57e7becdd6845cf7a0d65ecfac09006a9225)[1] which existing pipelines are using.  To be backwards compatible with that existing use and not need a flag-day change of existing pipelines:

```suggestion
        clean_files = config['clean'] + 'install build'
```

[1] I didn't put that into a PR and land it yet as it was planned to be obsolete by harmonzing the use (or lack thereof as it now appears it will be) of the _arch_ in the build across daos, iof and cart.

JohnMalmberg: Done.

vars/sconsBuild.groovy
```diff
@@ -2,49 +2,201 @@
...
+    script += 'SCONS_ARGS="' + scons_args +'"\n'
+    script += '''# the config cache is unreliable so always force a reconfig
+                 # with "--config=force"
+                 if ! scons --config=force $SCONS_ARGS; then
+                     rc=\${PIPESTATUS[0]}
+                     echo "$SCONS_ARGS failed"
+                     cat config.log || true
+                     exit \$rc
+                 fi'''
+    full_script = "#!/bin/bash\n" + set_cwd + prebuild + prefix_1 + script
```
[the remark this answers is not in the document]
JohnMalmberg: fixed.

vars/sconsBuild.groovy
```diff
@@ -1,7 +1,5 @@
 // vars/sconsBuild.groovy
 
-import com.intel.checkoutScm
-
```
brianjmurrell: Removing this leaves src/com/intel/checkoutScm.groovy and vars/checkoutScmWithSubmodules.groovy as dead code yes?  Shouldn't this patch remove them also then?

vars/sconsBuild.groovy
```diff
@@ -1,50 +1,201 @@
...
+    } else if (config['target']) {
+        scons_args += " SRC_PREFIX=\${WORKSPACE}}"
+    }
+    if (config['REQUIRES']) {
+        scons_args += " REQUIRES=${config['REQUIRES']}"
+    } else if (config['target']) {
+        scons_args += " REQUIRES=${config['target']}"
+    }
+    //scons -c is not perfect so get out the big hammer
+    clean_files = "_build.external* install build"
```
brianjmurrell: Your earlier comment made me wonder if we should be this aggressive (i.e. dangerous) with an `rm -rf` or if we should be more targeted:
```suggestion
    clean_files = "_build.external{,-Linux} install build"
```

particularly considering this is going to be `clean_files = "_build.external install build"` very soon, once the _arch_ stuff is removed from iof and cart.

I'm on the fence but do like to be conservative and targeting when using destructive commands like `rm -rf`.

vars/sconsBuild.groovy
brianjmurrell: Lines 152-158 right above: https://github.com/daos-stack/pipeline-lib/blob/fcc29777463e0fb09ac60a45caca01b04885209d/vars/sconsBuild.groovy#L152-L158 provides a pretty good example I think.

If you decide you want to change it.

vars/sconsBuild.groovy
brianjmurrell: Yeah.  Probably alright to just leave it as is for now since it will correct for all projects once _arch_ support is removed from iof and cart.

vars/sconsBuild.groovy
JohnMalmberg: Change made for next update.

vars/sconsBuild.groovy
JohnMalmberg: For now I would like to leave it the same.

PR #25, opened 2019-03-06
CORCI-607 Cancel old builds

brianjmurrell:
A function to cancel old builds of the currently running job.

### DIFF
--- a/src/com/intel/cancelPreviousBuildsInternal.groovy
+++ b/src/com/intel/cancelPreviousBuildsInternal.groovy
@@ -5,7 +5,7 @@ package com.intel
 /**
  * cancelPreviousBuildsInternal.groovy
  *
- * Routine to canel old builds in progress
+ * Routine to cancel old builds in progress
  */
 
 
@@ -25,7 +25,7 @@ def cancelPreviousBuildsInternal(Map config = [:]) {
   /* Iterating over the builds for specific job */
   for (def build : currentJob.builds) {
     /* If there is a build that is currently running and it's not current build */
-    if (build.isBuilding() && build.number.toInteger() != buildNumber) {
+    if (build.isBuilding() && build.number.toInteger() < buildNumber) {
       print "Stopping currently running build #${build.number}"
       /* Than stopping it */
       build.doStop()
--- a/src/com/intel/cancelPreviousBuildsInternal.groovy
+++ b/src/com/intel/cancelPreviousBuildsInternal.groovy
@@ -1,0 +1,35 @@
+// src/com/intel/cancelPreviousBuildsInternal.groovy
+
+package com.intel
+
+/**
+ * cancelPreviousBuildsInternal.groovy
+ *
+ * Routine to canel old builds in progress
+ */
+
+
+def cancelPreviousBuildsInternal(Map config = [:]) {
+  /**
+   * Cancel old builds method.
+   *
+   * @param config Map of parameters passed (currently none)
+   * @return Nothing
+   */
+
+  def jobName = env.JOB_NAME
+  def buildNumber = env.BUILD_NUMBER.toInteger()
+  /* Get job name */
+  def currentJob = Jenkins.instance.getItemByFullName(jobName)
+
+  /* Iterating over the builds for specific job */
+  for (def build : currentJob.builds) {
+    /* If there is a build that is currently running and it's not current build */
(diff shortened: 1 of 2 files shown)


brianjmurrell:
Tested in daos-stack/daos#239.  Build [#5](https://build.hpdd.intel.com/job/daos-stack/job/daos/job/PR-239/5) cancelled build [#4](https://build.hpdd.intel.com/job/daos-stack/job/daos/job/PR-239/4) which previously cancelled [#3](https://build.hpdd.intel.com/job/daos-stack/job/daos/job/PR-239/3), etc.

brianjmurrell:
It hasn't been decided that we will utilise this in any of our pipelines yet but it was worthwhile writing the code so that it's available in the future if not currently.

src/com/intel/cancelPreviousBuildsInternal.groovy
```diff
@@ -0,0 +1,35 @@
+// src/com/intel/cancelPreviousBuildsInternal.groovy
+
+package com.intel
+
+/**
+ * cancelPreviousBuildsInternal.groovy
+ *
+ * Routine to canel old builds in progress
```
ashleypittman: canel

src/com/intel/cancelPreviousBuildsInternal.groovy
```diff
@@ -0,0 +1,35 @@
...
+
+  def jobName = env.JOB_NAME
+  def buildNumber = env.BUILD_NUMBER.toInteger()
+  /* Get job name */
+  def currentJob = Jenkins.instance.getItemByFullName(jobName)
+
+  /* Iterating over the builds for specific job */
+  for (def build : currentJob.builds) {
+    /* If there is a build that is currently running and it's not current build */
+    if (build.isBuilding() && build.number.toInteger() != buildNumber) {
```
ashleypittman: I feel sure there should be a < rather than a != in here somewhere.

If the intention is that this should be called early on in the pipeline then that should be added as a comment.

brianjmurrell: I'm not sure there is anything requiring that it be called early in the pipeline.  For the use-case I (seems we) have in mind, it would be done early and that's the use-case I tested.

But I think if somebody had a use-case where it only happen after passing certain stages, say, Build for example, then I don't think there would be any harm in doing it there instead.

brianjmurrell:
> This is for old builds of the same PR I assume?

Yes.

brianjmurrell:
> What happens if a commit to master is made whilst the nightly run is still? Can this have a check for the job being triggered by a PR?

At least until somebody convinces me otherwise, I think that's for the caller to decide and not cancel old builds (i.e. conditionalize calling `cancelPreviousBuilds()`) if he doesn't want non-PR builds cancelled rather than baking that limitation into this generic function.

Conditionalizing by the caller can be done with: `when { changeRequest() }` in the step added to cancel previous builds such as:

```
        stage('Cancel Previous Builds') {
            when { changeRequest() }
            steps {
                cancelPreviousBuilds()
            }
        }
```

ashleypittman:
> > What happens if a commit to master is made whilst the nightly run is still? Can this have a check for the job being triggered by a PR?
> 
> At least until somebody convinces me otherwise, I think that's for the caller to decide and not cancel old builds (i.e. conditionalize calling `cancelPreviousBuilds()`) if he doesn't want non-PR builds cancelled rather than baking that limitation into this generic function.
> 
> Conditionalizing by the caller can be done with: `when { changeRequest() }` in the step added to cancel previous builds such as:
> 
> ```
>         stage('Cancel Previous Builds') {
>             when { changeRequest() }
>             steps {
>                 cancelPreviousBuilds()
>             }
>         }
> ```

This would be great as a comment in the code.

src/com/intel/cancelPreviousBuildsInternal.groovy
```diff
@@ -0,0 +1,35 @@
+// src/com/intel/cancelPreviousBuildsInternal.groovy
+
+package com.intel
+
+/**
```
ashleypittman: I've just noticed the Doxygen style comments, are we using them throughout?

brianjmurrell: Not yet.

PR #31, opened 2019-03-26
Post an error() if post-provisioning configuration fails

brianjmurrell:


brianjmurrell:
@ashleypittman Could review this one when you get a minute?  Thanks.

vars/provisionNodes.groovy
```diff
@@ -121,10 +121,17 @@ def call(Map config = [:]) {
                                              python-debuginfo python2-aexpect libcmocka          \
                                              python-pathlib python2-numpy git                    \
                                              golang-bin'''
-    sh script: 'set -x; rm -f ci_key*; ssh-keygen -N "" -f ci_key;' +
-               ' pdcp -R ssh -l root -w ' + nodeString +
-               ' ci_key* /tmp/;' +
-               ' pdsh -R ssh -l root -w ' + nodeString +
-               ' "' + provision_script + '" 2>&1 | dshbak -c'
+    def rc = 0
+    rc = sh(script: 'set -x; rm -f ci_key*; ssh-keygen -N "" -f ci_key;' +
```
ashleypittman: I find it much clearer to read the cart/Jenkinsfile script where """ is used rather than '.  Even if this way is preferred, you could use ' pdcp -R ssh -l root -w $nodeString' + to do variable substitution.

vars/provisionNodes.groovy
```diff
@@ -121,10 +121,17 @@ def call(Map config = [:]) {
...
+                    ' pdcp -R ssh -l root -w ' + nodeString +
+                    ' ci_key* /tmp/;' +
+                    ' pdsh -R ssh -l root -w ' + nodeString +
+                    ' "' + provision_script + '" 2>&1 | dshbak -c',
```
ashleypittman: This will return the return code of dshbak, not the provision_script which I doubt it what you expect or want.

brianjmurrell: Hoist with his own petard.

Well, kinda sorta.  I wasn't quite bitten by `$?` vs `${PIPESTATUS[0]}` but this is a pitfall I point out frequently.

vars/provisionNodes.groovy
brianjmurrell: Yeah.  It can just sometimes lead down escaping hell, particularly when you have shell script that needs escaping inside of groovy script that needs escaping, etc.

In this case I don't think it's so bad, so I have changed to multi-line strings using `"""` and '`'''` as needed.

PR #36, opened 2019-04-02
DAOS-2240 Add Jenkins RPM project repos to provisioned nodes

brianjmurrell:


vars/provisionNodes.groovy
```diff
@@ -113,14 +119,24 @@ def call(Map config = [:]) {
                               chmod 600 /localhome/jenkins/.ssh/{authorized_keys,id_rsa*}
                               chown -R jenkins.jenkins /localhome/jenkins/.ssh
                               echo \\"jenkins ALL=(ALL) NOPASSWD: ALL\\" > /etc/sudoers.d/jenkins
+                              yum -y install yum-utils
+                              pkgs=\\"${inst_rpms}\\"
```
ashleypittman: Why use a local variable here, rather than just using inst_rpms on L139?

brianjmurrell: `$pkgs` is used on more than just L139. Yes, I could interpolate `inst_rpms` through all the places where `$pkgs` is used but it seems cleaner (and easier to lift this script to somewhere else if needed) to have a single interface point between the script and it's caller.

ashleypittman: Ah, I only saw the one instance.

ashleypittman:
Does this still apply?  I can approve but it looks like it's probably been superseded now.

brianjmurrell:
@ashleypittman This will be needed when we migrate to testing using the built RPMs but since it already has a conflict that needs resolving before it can be landed, let's just put it on the back-burner until we are ready to move to testing using RPMs.

PR #37, opened 2019-04-04
CORCI-640 Create python3 symlink

brianjmurrell:
python34-3.4.9-3.el7.x86_64 seems to have removed the
python3 -> python3.4 symlink that existed previous to it.  This is
breaking our usages of python3.

Create the symlink if it doesn't exist.

### DIFF
--- a/vars/provisionNodes.groovy
+++ b/vars/provisionNodes.groovy
@@ -120,7 +120,11 @@ def call(Map config = [:]) {
                                              python2-avocado-plugins-varianter-yaml-to-mux       \
                                              python-debuginfo python2-aexpect libcmocka          \
                                              python-pathlib python2-numpy git                    \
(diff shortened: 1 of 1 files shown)


brianjmurrell:
```
$ rpm -qvl python34-3.4.9-3.el7.x86_64 | grep bin
-rwxr-xr-x    1 root    root                       78 Mar  7 15:51 /usr/bin/pydoc3.4
-rwxr-xr-x    2 root    root                    11384 Mar  7 15:51 /usr/bin/python3.4
-rwxr-xr-x    2 root    root                    11384 Mar  7 15:51 /usr/bin/python3.4m
-rwxr-xr-x    1 root    root                      230 Mar  7 15:51 /usr/bin/pyvenv-3.4
```

PR #40, opened 2019-04-09
Only diplay tee error when teeing

brianjmurrell:
Also remove unneeded escaping of $ inside single quotes.


brianjmurrell:
Fixed in another PR.

PR #42, opened 2019-04-10
Post an error to the job if runTest stage fails

brianjmurrell:
This is what signals the pipeline to not continue beyond the run_test.sh
state if it fails for example.


jolivier23:
It's saying there is a conflict that needs to be resolved

PR #47, opened 2019-05-03
CORCI-673 Get pdsh's exit status

brianjmurrell:
So that it can be reported to the caller and the caller can exit
accordingly.


vars/provisionNodes.groovy
```diff
@@ -120,14 +120,26 @@ def call(Map config = [:]) {
...
+                                                  python36-paramiko            \
+                                                  python2-avocado              \
+                                                  python2-avocado-plugins-output-html \
+                                                  python2-avocado-plugins-varianter-yaml-to-mux \
+                                                  python-debuginfo             \
+                                                  python2-aexpect libcmocka    \
+                                                  python-pathlib python2-numpy \
+                                                  git golang-bin; then
+                                  rc=\\${PIPESTATUS[0]}
+                                  for file in /etc/yum.repos.d/*.repo; do
```
ashleypittman: You could replace this whole block with "grep . /etc/yum.repos.d/*.repo"

brianjmurrell: Good point.  I have used that idiom in the past.

PR #50, opened 2019-05-09
CORCI-539 build: provisionNodes fix

JohnMalmberg:
ProvisionNodes needs to be able to be called multiple times in a step.

The Jenkins account being added to the CI nodes may already exist.

Need to sync the file system after provisioning to make sure data
persists after a power cycle.

### DIFF
--- a/vars/provisionNodes.groovy
+++ b/vars/provisionNodes.groovy
@@ -108,10 +108,16 @@ def call(Map config = [:]) {
         error "One or more nodes failed to provision!"
       }
     }
+
     // Prepare the node for daos/cart testing
-    def provision_script = '''set -ex
-                              groupadd -g 1101 jenkins || true
-                              useradd -b /localhome -g 1101 -u 1101 jenkins || true
+    def provision_script = "set -ex\n" +
+                           "my_uid=" + env.UID + "\n" +
+                           '''if ! grep ":$my_uid:" /etc/group; then
+                                groupadd -g $my_uid jenkins
+                              fi
+                              if ! grep ":$my_uid:$my_uid:" /etc/passwd; then
+                                useradd -b /localhome -g $my_uid -u $my_uid jenkins
+                              fi
                               mkdir -p /localhome/jenkins/.ssh
                               cat /tmp/ci_key.pub >> /localhome/jenkins/.ssh/authorized_keys
                               mv /tmp/ci_key.pub /localhome/jenkins/.ssh/id_rsa.pub
--- a/vars/provisionNodes.groovy
+++ b/vars/provisionNodes.groovy
@@ -112,11 +112,11 @@ def call(Map config = [:]) {
     // Prepare the node for daos/cart testing
     def provision_script = "set -ex\n" +
                            "my_uid=" + env.UID + "\n" +
-                           '''if ! grep ":$my_uid:" /etc/group; then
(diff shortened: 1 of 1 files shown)


vars/provisionNodes.groovy
```diff
@@ -110,8 +110,8 @@ def call(Map config = [:]) {
     }
     // Prepare the node for daos/cart testing
     def provision_script = '''set -ex
-                              groupadd -g 1101 jenkins
-                              useradd -b /localhome -g 1101 -u 1101 jenkins
+                              groupadd -g 1101 jenkins || true
+                              useradd -b /localhome -g 1101 -u 1101 jenkins || true
```
brianjmurrell: These will plow through an actual error to add the group or user.

Can we apply a more positive check for skip, like grepping the group and password files for the entry and not adding them if they exist?

JohnMalmberg: I can do that.

vars/provisionNodes.groovy
```diff
@@ -108,10 +108,16 @@ def call(Map config = [:]) {
...
       }
     }
+
     // Prepare the node for daos/cart testing
-    def provision_script = '''set -ex
-                              groupadd -g 1101 jenkins
-                              useradd -b /localhome -g 1101 -u 1101 jenkins
+    def provision_script = "set -ex\n" +
+                           "my_uid=" + env.UID + "\n" +
+                           '''if ! grep ":\\$my_uid:" /etc/group; then
```
brianjmurrell: Why not just `grep` for `jenkins` and skip all of the machinations to get the `uid`?

```suggestion
                           '''if ! grep "^jenkins:" /etc/group; then
```

ashleypittman: "getent passwd jenkins" would be the traditional way.

ashleypittman: In fact, "getent passwd jenkins || useradd -b /localhome -g 1101 -u 1101 jenkins" should check the user, try and add it if not found, and fail correctly if it can't be added.

JohnMalmberg: The most likely cause of a failure is that the GID or UID is in use for another name, and the grep report from the current patch will show that in the log.

It is the gid/uid that needs to match the uid of the build agent as seen in the env.UID variable that must be generated into the script, as that information is not available when the script is actually run on the remote node.

brianjmurrell: Thinking more about @ashleypittman's comment, I wonder if we want to be aware when we'd be masking an NSS entry by only checking for a local username/group.

JohnMalmberg: The "jenkins" username is only referenced ever inside the provisionNodes step.
If any tests care what the username is, than that is a different issue that needs to be addressed.

The account added to the provisioned nodes must match the env.UID and its assumed GID that is present in the procedure for NFS file serving to work.

brianjmurrell: Yes, but at some point in the future, if a _jenkins_ user is added to the network `passwd` database (i.e. IPA) to be used in some way (as of yet unknown of course) and we go adding one here that masks that, that could be a nasty issue to try to uncover.

I think generally speaking, we should try to avoid masking network database entries and without knowing what could be added in the future, we should be ensuring that we are not masking an entry each time we create one and error out if some day we do start trying to mask one.

PR #55, opened 2019-06-14
DAOS-1316 Test: Adding clustershell python package for running control plane command in cluster.

ravalsam:
Signed-off-by: Samir <samir.raval@intel.com>

### DIFF
--- a/vars/provisionNodes.groovy
+++ b/vars/provisionNodes.groovy
@@ -161,6 +161,7 @@ EOF'''
                                                   python-debuginfo             \
                                                   libcmocka python-pathlib     \
                                                   python2-numpy git            \
+                                                  clustershell                 \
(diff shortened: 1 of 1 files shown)


vars/provisionNodes.groovy
```diff
@@ -161,6 +161,7 @@ EOF'''
                                                   python-debuginfo             \
                                                   libcmocka python-pathlib     \
                                                   python2-numpy git            \
+                                                  clustershell                 \
```
brianjmurrell: Do we actually need clustershell or will python2-clustershell suffice?  We are using the python API of it yes?

ravalsam: It can work with python2-clustershell too. Let me change it to Python2.

PR #56, opened 2019-06-28
DAOS-623 build: Fix the filename from daos_m to daos

jolivier23:
We recently removed the _m from the config file so the cleanup
needs to remove the new file instead.

### DIFF
--- a/vars/sconsBuild.groovy
+++ b/vars/sconsBuild.groovy
@@ -134,7 +134,7 @@ def call(Map config = [:]) {
     if (config['clean']) {
         clean_files = config['clean']
     }
-    clean_files += ' install build {daos_m,iof,cart-Linux}.conf'
+    clean_files += ' install build {daos,iof,cart-Linux}.conf'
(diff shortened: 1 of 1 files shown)


vars/sconsBuild.groovy
```diff
@@ -134,7 +134,7 @@ def call(Map config = [:]) {
     if (config['clean']) {
         clean_files = config['clean']
     }
-    clean_files += ' install build {daos_m,iof,cart-Linux}.conf'
+    clean_files += ' install build {daos_m,daos,iof,cart-Linux}.conf'
```
brianjmurrell: FYI, just a bit of shell-geek-ery:

```suggestion
    clean_files += ' install build {daos{,_m},iof,cart-Linux}.conf'
```

:grin:

brianjmurrell: To be clear, not at all suggesting you make this change.  Was totally just a FYI.

jolivier23: thanks.  I think I'll pass.  We don't need daos_m for very long anyway

PR #65, opened 2019-09-10
CORCI-768 build: Convert to new repository format

JohnMalmberg:
Convert to new repository format to separate group repositories from
local repositories.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -25,7 +25,7 @@
 // Then a second PR submitted to comment out the @Library line, and when it
 // is landed, both PR branches can be deleted.
 
-// @Library(value="pipeline-lib@my_pr_branch") _
+@Library(value="pipeline-lib@corci-768") _
 
 pipeline {
     agent { label 'lightweight' }
@@ -59,7 +59,7 @@ pipeline {
                                   env.BUILD_ID,
                         format: 'yum',
                         maturity: 'test',
-                        tech: 'el7',
+                        tech: 'el-7',
                         repo_dir: 'artifact/artifacts/centos7',
                         download_dir: 'artifact/download',
                         test: true)
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -25,7 +25,7 @@
 // Then a second PR submitted to comment out the @Library line, and when it
 // is landed, both PR branches can be deleted.
 
-@Library(value="pipeline-lib@corci-768") _
+@Library(value="pipeline-lib@corci-768a") _
 
 pipeline {
     agent { label 'lightweight' }
@@ -93,7 +93,8 @@ pipeline {
                                    node_count: 1,
                                    snapshot: true,
                                    inst_rpms: "slurm slurm-example-configs" +
-                                              " slurm-slurmctld slurm-slurmd"
+                                              " slurm-slurmctld slurm-slurmd" +
+                                              " ipmctl"
                             runTest(
                         script: '''NODE=${NODELIST%%,*}
                                    ssh $SSH_KEY_ARGS jenkins@$NODE "set -x
@@ -130,7 +131,7 @@ pipeline {
                                    distro: 'sles12sp3',
                                    node_count: 1,
                                    snapshot: true,
-                                   inst_rpms: "slurm"
+                                   inst_rpms: "slurm ipmctl"
                             runTest(
                         script: '''NODE=${NODELIST%%,*}
                                    ssh $SSH_KEY_ARGS jenkins@$NODE "set -x
--- a/vars/provisionNodes.groovy
+++ b/vars/provisionNodes.groovy
@@ -94,6 +94,20 @@ def call(Map config = [:]) {
   if (distro.startsWith("sles") || distro.startsWith("leap")) {
       distro_type = 'suse'
   }
+  def gpg_key_urls = []
+  if (env.DAOS_STACK_REPO_SUPPORT != null) {
+     gpg_key_urls.add(env.DAOS_STACK_REPO_SUPPORT + 'RPM-GPG-KEY-CentOS-7')
+     gpg_key_urls.add(env.DAOS_STACK_REPO_SUPPORT +
+                      'RPM-GPG-KEY-CentOS-Debug-7')
+     gpg_key_urls.add(env.DAOS_STACK_REPO_SUPPORT +
+                      'RPM-GPG-KEY-CentOS-Testing-7')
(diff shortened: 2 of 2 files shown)


Jenkinsfile
```diff
@@ -25,7 +25,7 @@
 // Then a second PR submitted to comment out the @Library line, and when it
 // is landed, both PR branches can be deleted.
 
-// @Library(value="pipeline-lib@my_pr_branch") _
+@Library(value="pipeline-lib@corci-768") _
```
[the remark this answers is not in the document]
brianjmurrell: What if you have both uncommented but in an `if`/`else` type of conditional?

vars/provisionNodes.groovy
```diff
@@ -220,6 +256,8 @@ EOF'''
       if (inst_rpms) {
          provision_script += '''\nyum -y erase ''' + inst_rpms
       }
+      provision_script += '\nrpm --import https://download.fedoraproject.org/' +
+                                         'pub/epel//RPM-GPG-KEY-EPEL-7'
```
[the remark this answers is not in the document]
brianjmurrell: But this is not necessary on vanilla installations.  All that is necessary is to do on any EL7 machine is:

```
# yum install epel-release
```

and EPEL is ready to go.  That's why this has worked for the many months that it has without having to do this key import.

So the question is why is it not working now?  Can you paste (a link, if you prefer, to) the error you get without this explicit key addition?  My concern is that something else has subtly broken which we'd be better to uncover now rather than later.

Jenkinsfile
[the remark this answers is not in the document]
JohnMalmberg: According to the Jenkins documentation, the @Library statement is processed before the groovy code is run so it can only accept string constants available at that time.  The documentation shows using job parameters to control the library, but that will not help with triggered builds.

The dynamic syntax also will not allow replacing the default loaded library.  It does not error out the job, it just writes out a message in the log that the original library specification will be used.

vars/provisionNodes.groovy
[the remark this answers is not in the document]
JohnMalmberg: It appears to be a side effect of using yum add-repo with a Nexus group repository.
Installs fail on validating the repository until any the rpm --import /etc/pki/rpm-gpg/* is run.
If any additional GPG keys not already installed on the system are needed, then they also need to be imported with the rpm command.

I will be adjusting this PR to also verify that the ipmctl libraries install as they need their own GPG key.

vars/provisionNodes.groovy
```diff
@@ -220,6 +256,10 @@ EOF'''
       if (inst_rpms) {
          provision_script += '''\nyum -y erase ''' + inst_rpms
       }
+      provision_script += '\nrpm --import /etc/pki/rpm-gpg/*'
```
brianjmurrell: Why do we trust these keys?  We should only install keys that come from a trustable source, be that the secure web server of a known and trustable entity or a web server controlled by us that we have populated the keys into only after verifying their authenticity.

These keys are the linchpin to the trustability of the entire system since they are what are validating the RPMs being installed.  If we fail to be diligent at this point, the whole system comes crumbling down.

vars/provisionNodes.groovy
```diff
@@ -254,10 +294,26 @@ EOF'''
                               ln -s python3.6 /usr/bin/python3
                           fi'''
     } else if (distro_type == "suse") {
-      if (repository != '') {
-        provision_script += '\nzypper --non-interactive ar ' +
-                            '--gpgcheck-allow-unsigned -f ' +
-                            repository + ' daos-stack-group-repo'
+      // Needed for sles-12.3/leap-42.3 only
+      provision_script += '\nrpm --import http://download.opensuse.org/' +
```
brianjmurrell: Lack of SSL here is insecure as it's subject to a MitM attack.

vars/provisionNodes.groovy
```diff
@@ -254,10 +294,26 @@ EOF'''
...
+                                 'repositories/science:/HPC/' +
+                                 'openSUSE_Leap_42.3/repodata/repomd.xml.key'
+      provision_script += '\nrpm --import http://download.opensuse.org/' +
```
brianjmurrell: Ditto.

vars/provisionNodes.groovy
JohnMalmberg: Changed to import keys from local site that we control.

vars/provisionNodes.groovy
JohnMalmberg: Fixed.  The http: URL is what is stored in the repository .repo file.

vars/provisionNodes.groovy
JohnMalmberg: Fixed.

vars/provisionNodes.groovy
```diff
@@ -256,8 +270,12 @@ EOF'''
...
          provision_script += '''\nyum -y erase ''' + inst_rpms
       }
-      provision_script += '\nrpm --import https://download.fedoraproject.org/' +
-                                         'pub/epel//RPM-GPG-KEY-EPEL-7'
+      gpg_key_urls.each { gpg_url ->
+        provision_script += '\nrpm --import ' + gpg_url
+      }
+      provision_script += '\nrpm --import ' +
+              'https://copr-be.cloud.fedoraproject.org/results/jhli' +
+              '/ipmctl/pubkey.gpg'
```
brianjmurrell: Given that this is project (i.e. DAOS) specific, it would be good to figure out a way to push this out to that project, at some point in the future and keep this routine as generic as possible.  Realising of course we are already down the road of non-genericness.  But probably much of that can be eliminated with the `inst_repos` and `inst_rpms` support added subsqeuently (to the non-genericness being added) or something like it for full repo paths and keys.

vars/provisionNodes.groovy
```diff
@@ -293,9 +311,12 @@ EOF'''
                           fi'''
     } else if (distro_type == "suse") {
       // Needed for sles-12.3/leap-42.3 only
-      provision_script += '\nrpm --import http://download.opensuse.org/' +
+      provision_script += '\nrpm --import https://download.opensuse.org/' +
                                  'repositories/science:/HPC/' +
                                  'openSUSE_Leap_42.3/repodata/repomd.xml.key'
+      provision_script += '\nrpm --import https://download.opensuse.org/' +
+                                 'repositories/home:/jhli/SLE_15/' +
+                                 'repodata/repomd.xml.key'
```
brianjmurrell: Ditto for these.

PR #71, opened 2019-09-17
Don't publish from PRs

brianjmurrell:


vars/publishToRepository.groovy
```diff
@@ -35,6 +35,12 @@ def call(Map config = [:]) {
         println('No REPOSITORY_URL environment variable configured.')
         return
     }
+
+    // Don't publish from PRs
+    if (env.CHANGE_ID) {
```
JohnMalmberg: There are some cases where we want to publish from a PR, usually into a test repository, such as the self test for this module.

brianjmurrell: Hrm.  How did the tests pass then?

JohnMalmberg: Tests are actually only run / graded if the test parameter is passed to the publishToRepositorySystem if that routine exists and the repository is not an APT repository.

brianjmurrell: But the tests were run for this PR.

PR #76, opened 2019-09-24
Support defaults for package name and distros

brianjmurrell:
Allows a `Jenkinsfile` to simply call:

```
packageBuildingPipeline()
```

to build on all supported distros but allows projects to only build on a subset of distros if desired.

### DIFF
--- a/vars/packageBuildingPipeline.groovy
+++ b/vars/packageBuildingPipeline.groovy
@@ -49,7 +49,7 @@ def call(Map pipeline_args) {
         distros = pipeline_args['distros']
     } else {
         distros = ['centos7', 'sles12.3', 'leap42.3',
-                   'leap15', 'ubuntu18.04', 'ubuntu_rolling']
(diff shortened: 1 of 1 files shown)


vars/packageBuildingPipeline.groovy
```diff
@@ -42,6 +42,20 @@
...
 
 def call(Map pipeline_args) {
+    if (!pipeline_args) {
+        pipeline_args = [:]
+    }
+    if (pipeline_args['distros']) {
+        distros = pipeline_args['distros']
+    } else {
+        distros = ['centos7', 'sles12.3', 'leap42.3',
+                   'leap15', 'ubuntu18.04', 'ubuntu_rolling']
```
JohnMalmberg: I think we should drop ubuntu18.04 from the default distros.

brianjmurrell: Done

PR #78, opened 2019-09-30
CORCI-786 build: Allow rpmlint checks to fail.

JohnMalmberg:
vars/packageBuildingPipeline.groovy:
  The Slurm and Meson pckages currently do not pass lint.
  This allows a Jenkinsfile to supresss the lint checks.

### DIFF
--- a/vars/packageBuildingPipeline.groovy
+++ b/vars/packageBuildingPipeline.groovy
@@ -98,9 +98,11 @@ def call(Map pipeline_args) {
                             }
                         }
                         steps {
-                            sh 'make ' +
-                               pipeline_args.get('make args', '') +
-                               ' rpmlint'
+                            sh script: 'make ' +
+                                       pipeline_args.get('make args', '') +
+                                       ' rpmlint',
+                               returnStatus: !pipeline_args.get('rpmlint_check',
+                                                                 true)
                         }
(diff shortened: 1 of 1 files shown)


vars/packageBuildingPipeline.groovy
```diff
@@ -98,9 +98,11 @@ def call(Map pipeline_args) {
                             }
                         }
                         steps {
-                            sh 'make ' +
-                               pipeline_args.get('make args', '') +
-                               ' rpmlint'
+                            sh script: 'make ' +
+                                       pipeline_args.get('make args', '') +
+                                       ' rpmlint',
+                               returnStatus: !pipeline_args.get('rpmlint_check',
```
brianjmurrell: Not following why you want the `!` in there.  I'm taking it that `rpmlint_check` is supposed to default to `true` and so can be passed as `false` to allow it to fail which directly correlates to whether you want `returnStatus` to be `true` or `false`, doesn't it?  So why the `!`?

JohnMalmberg: Return status = True causes lint_check failures to not fail the build.  The lint output can still be set in the log.

That is the opposite value for the rpmlint_check boolean parameter to disable the lint check.

PR #80, opened 2019-10-03
DAOS-3268 build: Add scons_exe variable so alternative can be used

jolivier23:
Need to use scons-3, for instance on some platforms

Signed-off-by: Jeff Olivier <jeffrey.v.olivier@intel.com>

### DIFF
--- a/vars/sconsBuild.groovy
+++ b/vars/sconsBuild.groovy
@@ -16,6 +16,7 @@ def call(Map config = [:]) {
    * @return Integer status of script
    *
    * config['build_deps'] Setting for --build-deps.  Default 'yes'.
+   * config['scons_exe'] Name of scons executable.  Default 'scons'.
    * config['clean'] Space separated filenames to be removed after
    *  the prebuild option.
    *  Defaults _build.external* install build'.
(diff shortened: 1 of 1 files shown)


vars/sconsBuild.groovy
```diff
@@ -92,8 +93,12 @@ def call(Map config = [:]) {
         set_cwd = "cd ${config['target']}\n"
     }
 
+    def scons_exe = 'scons'
```
brianjmurrell: Can we do some autodetection here, like check for the existence of a priority-descending list of alternatives.  Perhaps `scons-3`, followed by `scons` to start.

jolivier23: Not until we have build supporting scons-3.   For now, we have to be explicit.  Otherwise, daos (which can't currently build with 3) will fail.

vars/sconsBuild.groovy
```diff
@@ -180,7 +185,7 @@ def call(Map config = [:]) {
     script += 'SCONS_ARGS="' + scons_args + '"\n'
     script += '''# the config cache is unreliable so always force a reconfig
                  # with "--config=force"
-                 if ! scons --config=force $SCONS_ARGS''' +
+                 if ! ''' + scons_exe + ''' --config=force $SCONS_ARGS''' +
```
brianjmurrell: Technically, only need `'` here:

```suggestion
                 if ! ''' + scons_exe + ' --config=force $SCONS_ARGS' +
```

but most certainly not worth pushing a new one just for that.  I'll keep a mental note for the next time I'm in there.

vars/sconsBuild.groovy
brianjmurrell: Just so I'm clear, we're making this optional at this point because CaRT will build with `scons-3` but DAOS not yet?

jolivier23:
>Just so I'm clear, we're making this optional at this point because CaRT will build with scons-3 but DAOS not yet?

Yes, it's optional.   Even in the cart patch I posted, I only use it on CentOS.

PR #83, opened 2019-10-21
Abandon the OBS build tool and use mock instead

brianjmurrell:
The OBS build tool has depdency resolution limitations that make it
unable to build some packages. In particular it only recognizes a
single verison of a given package and so is quite unable to handle
our multi-package-version repos.

Since we are using mock for all RPM builds, we might as well just use
Fedora as the build platform to get much more recent versions of mock.

Signed-off-by: Brian J. Murrell brian.murrell@intel.com

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -121,7 +121,10 @@ pipeline {
             stage('provisionNodes_with_slurm_sles12') {
                 when {
                     beforeAgent true
-                    expression { env.NO_CI_TESTING != "true" }
+                    allOf {
+                        expression { false }
+                        expression { env.NO_CI_TESTING != "true" }
+                    }
                 }
                 agent {
                     label 'ci_vm1'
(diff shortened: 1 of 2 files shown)


brianjmurrell:
Coordinated in what way, other than this needs to land and then daos-stack/packaging#20 need to rebase on this repos master?

JohnMalmberg:
Any repository doing packaging that does not have the new Dockerfile.mockbuild will start failing as soon as this patch is merged with master.

To prevent fails, all the repositories doing packaging first need to be updated to add the Dockerfile.mockbuild before removing the obsoleted RPM based Dockerfiles from them.

brianjmurrell:
Well, not exactly.  Projects using (a copy of) this repo don't get built for any reason other than an update by a developer.  Any package which gets an update for some reason will have to update the packaging, per our usual process and will then get the needed updates.  Or is there an aspect that I'm not thinking of?

PR #84, opened 2019-10-23
Add a packageBuildingPipeline variant for DAOS

brianjmurrell:
That builds DAOS with the package being updated as a test against the
packge being updated.

### DIFF
--- a/vars/packageBuildingPipelineDAOS.groovy
+++ b/vars/packageBuildingPipelineDAOS.groovy
@@ -1,0 +1,621 @@
+#!/usr/bin/groovy
+/* Copyright (C) 2019 Intel Corporation
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted for any purpose (including commercial purposes)
+ * provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice,
+ *    this list of conditions, and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions, and the following disclaimer in the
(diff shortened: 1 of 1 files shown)


vars/packageBuildingPipelineDAOS.groovy
```diff
@@ -0,0 +1,621 @@
...
+                                additionalBuildArgs '--build-arg UID=$(id -u)' +
+                                                    ' --build-arg JENKINS_URL=' +
+                                                    env.JENKINS_URL
+                            }
+                        }
+                        steps {
+                            sh label: "Build package",
+                               script: '''rm -rf artifacts/leap15/
+                                          mkdir -p artifacts/leap15/
+                                          make CHROOT_NAME="opensuse-leap-15.1-x86_64" ''' +
```
JohnMalmberg: Why is this specifically leap 15.1 and not leap 15 to be Leap current?

How do we specify which Leap and which CentOS release to build?

brianjmurrell: [ As discussed on Skype, but just to answer the questions for the record...]

> Why is this specifically leap 15.1 and not leap 15 to be Leap current?

Because unlike CentOS, where [mock](https://github.com/rpm-software-management/mock) understands it my [major version, 7](https://github.com/rpm-software-management/mock/blob/devel/mock-core-configs/etc/mock/epel-7-x86_64.cfg) it knows leap by it's [major and minor](https://github.com/rpm-software-management/mock/blob/devel/mock-core-configs/etc/mock/opensuse-leap-15.1-x86_64.cfg).  Arguably a [bug](https://github.com/rpm-software-management/mock/issues/386).
> 
> How do we specify which Leap and which CentOS release to build?

It's in the `CHROOT_NAME`, if I'm understanding your question.

PR #85, opened 2019-10-23
DCO-8879 build: Fixup SSH access.

JohnMalmberg:
Allow CI to ssh access to the root account of CI nodes.

### DIFF
--- a/vars/provisionNodes.groovy
+++ b/vars/provisionNodes.groovy
@@ -218,6 +218,7 @@ EOF'''
                           fi
                           mkdir -p /localhome/jenkins/.ssh
                           cat /tmp/ci_key.pub >> /localhome/jenkins/.ssh/authorized_keys
+                          cat /tmp/ci_key.pub >> /root/.ssh/authorized_keys
                           mv /tmp/ci_key.pub /localhome/jenkins/.ssh/id_rsa.pub
                           mv /tmp/ci_key /localhome/jenkins/.ssh/id_rsa
                           mv /tmp/ci_key_ssh_config /localhome/jenkins/.ssh/config


JohnMalmberg:
The chef code that recreates the /root/.ssh directory runs after the /etc/rc.d/local is run.

Developers know this is only a temporary solution and they need to find a different solution long term.

amjustin13:
Yes, treating this as a short term solution. We should keep https://jira.hpdd.intel.com/browse/DCO-8878 open to come up with a better solution (possibly the suggestion the Brian had to add -mca plm_rsh_args "-i ci_key -l root" arguments to the orterun command) for the long term when we will be having to handle tests rebooting the nodes.

brianjmurrell:
@amjustin13 This PR is the provisioning side of my `-mca plm_rsh_args "-i ci_key -l root"`.  When this PR is landed, you need to implement those `orterun` arguments.  To be clear that solution and this PR work together and so are an **AND** not an **OR**.

amjustin13:
> @amjustin13 This PR is the provisioning side of my `-mca plm_rsh_args "-i ci_key -l root"`. When this PR is landed, you need to implement those `orterun` arguments. To be clear that solution and this PR work together and so are an **AND** not an **OR**.

Ok, thanks @brianjmurrell . I'm a bit confused however, I though adding the `-mca plm_rsh_args "-i ci_key -l root"` was specific for resolving the issue of the nodes rebooting. Now that the keys are added in I was under the impression that keeping the command how it is: `sudo orterun --allow-run-as-root... daos_server...` should also work if not rebooting. 
Wouldn't these two commands below essentially be doing the same thing? With the exception that command 2. is additionally passing in the key to use, which would only be needed to be passed in after reboot since the authorized keys get wiped out:
1. `sudo orterun --allow-run-as-root... daos_server...` and 
2. `orterun... -mca plm_rsh_args "-i ci_key -l root" daos_server...`

JohnMalmberg:
Please move the discussion to DCO-8778 ticket.

PR #86, opened 2019-10-25
Erase all previously installed jenkins repos

brianjmurrell:
Not just previous incarnations of repos we want to add.

This is a poor man's node-reuse (rather than reprovision).


brianjmurrell:
I suppose we will have to, if we don't have real provisioning before we start running tests on Leap/SLES.  I have no idea where Zypper keeps this data at this point though.

PR #87, opened 2019-11-01
DCO-8741 build: Add CentOS 8 stage

JohnMalmberg:
Add a CentOS 8 Stage needed for building SLURM.

### DIFF
--- a/vars/packageBuildingPipelineDAOS.groovy
+++ b/vars/packageBuildingPipelineDAOS.groovy
@@ -203,12 +203,81 @@ def call(Map pipeline_args) {
                             }
                         }
                     } //stage('Build on CentOS 7')
+                    stage('Build on CentOS 8') {
+                        when {
+                            beforeAgent true
+                            allOf {
+                                expression { distros.contains('centos8')}
+                            }
+                        }
+                        agent {
+                            dockerfile {
+                                filename 'packaging/Dockerfile.mockbuild'
+                                label 'docker_runner'
+                                args  '--group-add mock' +
+                                      ' --cap-add=SYS_ADMIN' +
+                                      ' --privileged=true'
+                                additionalBuildArgs '--build-arg UID=$(id -u)' +
+                                                    ' --build-arg JENKINS_URL=' +
+                                                    env.JENKINS_URL
+                            }
+                        }
+                        steps {
+                            sh label: "Build package",
+                               script: '''rm -rf artifacts/8/
+                                          mkdir -p artifacts/centos8/
+                                          make CHROOT_NAME="epel-8-x86_64" ''' +
+                                       pipeline_args.get('make args', '') + ' chrootbuild ' +
+                                       pipeline_args.get('add_make_targets', '')
+                        }
+                        post {
+                            success {
+                                sh label: "Collect artifacts",
+                                   script: '''(cd /var/lib/mock/epel-8-x86_64/result/ &&
+                                              cp -r . $OLDPWD/artifacts/centos8/)\n''' +
+                                              pipeline_args.get('add_archiving_cmds', '') +
+                                             '\ncreaterepo artifacts/centos8/'
+                                publishToRepository product: package_name,
+                                                    format: 'yum',
+                                                    maturity: 'stable',
+                                                    tech: 'el-8',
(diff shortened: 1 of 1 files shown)


vars/packageBuildingPipelineDAOS.groovy
```diff
@@ -203,12 +203,81 @@ def call(Map pipeline_args) {
...
+                            }
+                        }
+                    } //stage('Build on CentOS 8')
                     stage('Build on SLES 12.3') {
                         when {
                             beforeAgent true
                             allOf {
                                 environment name: 'SLES12_3_DOCKER', value: 'true'
-                                expression { false }
+                                expression { pipeline_args.get('sles12_build', false) }
```
brianjmurrell: Isn't this in fact just duplicating the functionality `distros` argument more or less?  Perhaps the more accurate thing to do here is removing `sles12.3` and `leap42.3` from the default `distros` list:
https://github.com/daos-stack/pipeline-lib/blob/0f96842adea881fd0166f682a69e0384c22e7a8d/vars/packageBuildingPipelineDAOS.groovy#L50-L53

vars/packageBuildingPipelineDAOS.groovy
```diff
@@ -203,12 +203,81 @@ def call(Map pipeline_args) {
                             }
                         }
                     } //stage('Build on CentOS 7')
+                    stage('Build on CentOS 8') {
+                        when {
+                            beforeAgent true
+                            allOf {
+                                expression { distros.contains('centos8')}
```
brianjmurrell: If you end up having to push another commit:
```suggestion
                                expression { distros.contains('centos8') }
```
to make the whitespace consistent.  And could you update the other instances of this particular inconsistency in this file while you are at it?

vars/packageBuildingPipelineDAOS.groovy
JohnMalmberg: I considered that, however we were listing the sles-12.3 and leap42.3 for some projects where we did not need to build all distros.  So that change would potentially enable builds that are now disabled.

And I wanted a way to allow building sles-12.3 and leap42.3 if needed with out having to also have to have a pipeline-lib PR also needed.

vars/packageBuildingPipelineDAOS.groovy
JohnMalmberg: Pushing updated commit with change.

vars/packageBuildingPipelineDAOS.groovy
brianjmurrell: Not sure I'm following.  For builds where we are explicitly listing `distros`, there would be no change in behaviour by reducing the default list.  For builds where we are relying on the default value, they would just lose the sles 12 and leap 42 builds, which is no great issue at this point in any case.

JohnMalmberg: It is possible that there are builds that we have SLES-12.3 and Leap 42.3 enabled in the Jenkinsfile that are using this groovy file which is force disabling SLES-12.3 and Leap42.3.

So removing SLES-12.3 and Leap42.3 from the default  would enable the builds where they are currently being disabled.

The only way to make sure would be to inspect each and every packaging job to see if they are calling this script and how they are calling it.

vars/packageBuildingPipelineDAOS.groovy
```diff
@@ -48,8 +48,7 @@ def call(Map pipeline_args) {
     if (pipeline_args['distros']) {
         distros = pipeline_args['distros']
     } else {
-        distros = ['centos7', 'sles12.3', 'leap42.3',
-                   'leap15', 'ubuntu_rolling']
+        distros = ['centos7', 'leap15', 'ubuntu_rolling']
```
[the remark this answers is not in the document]
JohnMalmberg: I had assumed that script would be removed when the migration using mockbuild for SUSE was complete, so was leaving it alone.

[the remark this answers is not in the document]
brianjmurrell: No.  `packageBuildingPipeline()` will remain for building projects that don't want to test against building DAOS.  I.e. non-DAOS affecting projects, actually which I would say the slurm project is.

PR #88, opened 2019-11-04
DCO-8741 build: Add CentOS 8 part 2

JohnMalmberg:
Update packageBuildingPipeline.groovy for CentOS 8 and
retire SLES-12/Leap-42 for default builds.


brianjmurrell:
The similar changes to `packageBuildingPipelineDAOS()` are/were still worthwhile I think.  We will need them eventually.  Why not combine these changes with that PR?

JohnMalmberg:
That PR has already been merged.

PR #90, opened 2019-11-06
CORCI-819 build: Allow bootstrap of RPMs

JohnMalmberg:
vars/packageBuildingPipelineDAOS.groovy
When boot-strapping a new Jenkins the test stage can fail when not all
the RPMs are present.

Ignore test failures caused by missing packages.

### DIFF
--- a/vars/packageBuildingPipelineDAOS.groovy
+++ b/vars/packageBuildingPipelineDAOS.groovy
@@ -631,7 +631,7 @@ def call(Map pipeline_args) {
                                           git checkout ''' + pipeline_args.get('daos_test_branch',
                                                                                'origin/master') + '''
                                           git submodule update --init
-                                          if make PR_REPOS="''' + env.JOB_NAME.split('/')[1] +
+                                          if ! make PR_REPOS="''' + env.JOB_NAME.split('/')[1] +
                                                '@' + env.BRANCH_NAME + ':' + env.BUILD_NUMBER +
                                                ' ' + pipeline_args.get('test_repos', '') + '''" \
                                                CHROOT_NAME=epel-7-x86_64                        \
--- a/vars/packageBuildingPipelineDAOS.groovy
+++ b/vars/packageBuildingPipelineDAOS.groovy
@@ -631,11 +631,14 @@ def call(Map pipeline_args) {
                                           git checkout ''' + pipeline_args.get('daos_test_branch',
                                                                                'origin/master') + '''
                                           git submodule update --init
-                                          make PR_REPOS="''' + env.JOB_NAME.split('/')[1] +
+                                          if make PR_REPOS="''' + env.JOB_NAME.split('/')[1] +
                                                '@' + env.BRANCH_NAME + ':' + env.BUILD_NUMBER +
                                                ' ' + pipeline_args.get('test_repos', '') + '''" \
                                                CHROOT_NAME=epel-7-x86_64                        \
-                                               -C utils/rpms chrootbuild'''
+                                               -C utils/rpms chrootbuild; then
+                                            grep 'No matching package to install'               \
+                                                 /var/lib/mock/epel-7-x86_64/result/root.log
+                                            fi'''
                         }
                         post {
                             always {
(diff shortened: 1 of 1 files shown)


JohnMalmberg:
Tested with https://build-stage.hpdd.intel.com/job/daos-stack/job/libfabric/view/change-requests/job/PR-17/

vars/packageBuildingPipelineDAOS.groovy
```diff
@@ -631,11 +631,14 @@ def call(Map pipeline_args) {
                                           git checkout ''' + pipeline_args.get('daos_test_branch',
                                                                                'origin/master') + '''
                                           git submodule update --init
-                                          make PR_REPOS="''' + env.JOB_NAME.split('/')[1] +
+                                          if make PR_REPOS="''' + env.JOB_NAME.split('/')[1] +
```
brianjmurrell:
```suggestion
                                          if ! make PR_REPOS="''' + env.JOB_NAME.split('/')[1] +
```

vars/packageBuildingPipelineDAOS.groovy
```diff
@@ -631,11 +631,14 @@ def call(Map pipeline_args) {
...
                                                '@' + env.BRANCH_NAME + ':' + env.BUILD_NUMBER +
                                                ' ' + pipeline_args.get('test_repos', '') + '''" \
                                                CHROOT_NAME=epel-7-x86_64                        \
-                                               -C utils/rpms chrootbuild'''
+                                               -C utils/rpms chrootbuild; then
+                                            grep 'No matching package to install'               \
```
brianjmurrell: I'm concerned about false positives here, where a package should be found but isn't due to a bug.

I understand the problem we are trying to solve, but just wonder if there is an alternate way to solve it.

What about global Jenkins variable called BOOTSTRAP (or something aptly named) which allows this stage to fail, or not even be run perhaps.

JohnMalmberg: A missing package other than the package being built is not something that should block a PR for or a build of that package.

Right now, I need to run a new set of builds on the master branch, and this test is preventing it because the other RPMs are not present.

brianjmurrell: It should be if it's got a `Requires:` for something that cannot be found because the `Requires:` is wrong.

JohnMalmberg: A more comprehensive sanity check that still allows a bootstrap condition can be added later.

This should be good enough to get the packages built.

vars/packageBuildingPipelineDAOS.groovy
```diff
@@ -631,11 +631,16 @@ def call(Map pipeline_args) {
...
+                                          if ! make PR_REPOS="''' + env.JOB_NAME.split('/')[1] +
                                                '@' + env.BRANCH_NAME + ':' + env.BUILD_NUMBER +
                                                ' ' + pipeline_args.get('test_repos', '') + '''" \
                                                CHROOT_NAME=epel-7-x86_64                        \
-                                               -C utils/rpms chrootbuild'''
+                                               -C utils/rpms chrootbuild; then
+                                            grep 'No matching package to install'               \
+                                                 /var/lib/mock/epel-7-x86_64/result/root.log
+                                          else
+                                            exit ${PIPESTATUS[0]}
```
brianjmurrell: Doesn't this factor out to:
```suggestion
                                            exit 0
```
since this code-path only gets here if the make is successful?  But then I don't see why the `else` branch is even needed:

```
$ (if ! true; then exit 2; fi); echo $?
0
$ (if ! false; then exit 2; fi); echo $?
2
```

But most importantly, I think a comment around this code would be super useful for somebody coming and looking at this for the first time or in a long time, and forgetting what the use-case is that is being covered here.  It could be very easy to imagine somebody not considering the bootstrap situation and deciding that this check could lead to false positives and want to remove it due to not understanding why it's here in the first place.  I could see myself very easily doing that.

JohnMalmberg: The else step only was needed for the inverted logic test.  I have removed it and added a comment as to why the grep is present.

PR #91, opened 2019-11-11
CORCI-750 build: Add coverity to sconsBuild

JohnMalmberg:
Add coverity support to sconsBuild

### DIFF
--- a/vars/sconsBuild.groovy
+++ b/vars/sconsBuild.groovy
@@ -183,6 +183,22 @@ def call(Map config = [:]) {
                     fi\n'''
     script += clean_cmd
     script += 'SCONS_ARGS="' + scons_args + '"\n'
+
+    if (config['coverity']) {
+        sh 'rm -rf ./cov_analysis ./cov-int'
+        Map cov_config = [:]
+        cov_config['project'] = config['coverity']
+        cov_config['tool_path'] = './cov_analysis'
+        try {
+            coverityToolDownloadSystem(cov_config)
+        } catch (java.lang.NoSuchMethodError e) {
+            println('Could not find a coverityToolDownloadSystem step in' +
+                    ' a shared groovy library')
+            return
+        }
+        script += "PATH+=:${WORKSPACE}/cov_analysis/bin\n"
+        scons_exe = "cov-build --dir cov-int " + scons_exe
+    }
     script += '''# the config cache is unreliable so always force a reconfig
                  # with "--config=force"
                  if ! ''' + scons_exe + ''' --config=force $SCONS_ARGS''' +


vars/sconsBuild.groovy
```diff
@@ -183,6 +183,22 @@ def call(Map config = [:]) {
...
+        sh 'rm -rf ./cov_analysis ./cov-int'
+        Map cov_config = [:]
+        cov_config['project'] = config['coverity']
+        cov_config['tool_path'] = './cov_analysis'
+        try {
+            coverityToolDownloadSystem(cov_config)
+        } catch (java.lang.NoSuchMethodError e) {
+            println('Could not find a coverityToolDownloadSystem step in' +
+                    ' a shared groovy library')
+            return
```
brianjmurrell: Doesn't this mean that `sconsBuild` returns a success without having done the build?

JohnMalmberg: Yes.  If the site has not configured a coverityToolDownloadSystem step in a shared library, then that should not fail a build.

This means we do not need to have a test in the Jenkinsfile to check if Coverity scans are enabled for that instance.

jolivier23:
looks like it has conflicts or some such

vars/sconsBuild.groovy
JohnMalmberg: The coverity code path only runs if coverity is passed as an optional parameter with the coverity project name.

This option is not used for creating artifacts for other stages or CI.

We will probably limit Coverity builds to the master branch.

brianjmurrell: So should it be an error then if the `coverty` option is given but the download fails rather than returning _success_ without actually doing a build?

I'm just having a hard time understanding how `sconsBuild` should be able to return success without actually doing any building.  Do you have a PR of another pipeline that uses `sconsBuild` in this way that might make it more clear why `sconsBuild` can return success without actually building anything?

vars/sconsBuild.groovy
```diff
@@ -183,6 +183,22 @@ def call(Map config = [:]) {
...
+        }
+        script += "PATH+=:${WORKSPACE}/cov_analysis/bin\n"
+        scons_exe = "cov-build --dir cov-int " + scons_exe
```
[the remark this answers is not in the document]
JohnMalmberg: I am going by the Coverity examples for using it with an scons project.

vars/sconsBuild.groovy
JohnMalmberg: This is the PR I am using to test this code:
https://github.com/daos-stack/daos/pull/1395

Eventually that stage will be in the parallel with the testing since it takes at least 20 minutes and we don't want to delay testing for it to run.  And the Coverity scans would be changed to only run on the master branch.

A site has to have configured a coverityToolDownloadSystem step in order to have the tools needed to do a Coverity scan.

The intent here is for the stage to always try to run the Coverity Scan on the master branch, so I do not need logic in the Jenkinsfile to know if a site can run Coverity or not.

If the site has not configured a coverityToolDownloadSystem, then they will not get Coverity artifacts and when they look at the logs for the Coverity stage it will tell them why.

vars/sconsBuild.groovy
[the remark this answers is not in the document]
jolivier23: Nevermind, I'm confusing it with bullseye

vars/sconsBuild.groovy
brianjmurrell: > This is the PR I am using to test this code:
> [daos-stack/daos#1395](https://github.com/daos-stack/daos/pull/1395)

Ahhh.  This is the missing link.  I can see now that the Coverty stage is separate and so returning success from it even when it does nothing doesn't do anything to harm the job.

What about silent failures though?  What if a site has Coverty, wants the scans but for some reason `coverityToolDownloadSystem` fails for some unexpected reason?  They won't know about that until they go to look at the scan results yes?

JohnMalmberg: The only failure that is allowed is coverityToolDownloadSystem not being found.
If a site provides one that is broken, the failure will stop the job, so they will need to investigate.

brianjmurrell: I just created an MR in the system pipeline-lib for such a case as I was thinking about here.  I guess we just have to hope there are no others.

PR #95, opened 2019-11-18
CORCI-826 build: Clean out /tmp/daos.log

JohnMalmberg:
Make sure /tmp/daos.log and /tmp/server.log are deleted during
provisionNodes.

### DIFF
--- a/vars/provisionNodes.groovy
+++ b/vars/provisionNodes.groovy
@@ -156,19 +156,22 @@ def call(Map config = [:]) {
   }
 
   sshagent (credentials: ['daos-provisioner']) {
-
-    sh script:'clush -l root -w ' + nodeString +
-              ' --connect_timeout 30' +
-              ' "ls -lh /tmp/*.log; rm -f /tmp/daos.log /tmp/server.log"'
-        returnStatus: true
+    def ns_rc = 0
+    ns_rc = sh script:'clush -l root -w ' + nodeString +
+                      ' --connect_timeout 30 -S' +
+                      ' "ls -lh /tmp/*.log; rm -f /tmp/daos.log /tmp/server.log"'
+               returnStatus: true
+    println "Cleanup result = ${ns_rc}"
     if (config['power_only']) {
-      sh script: """./jenkins/node_powercycle.py \
-                     --node=${nodeString}""",
-         returnStatus: true
+      ns_rc = sh script: """./jenkins/node_powercycle.py \
+                            --node=${nodeString}""",
+                 returnStatus: true
+      println "Powercycle result = ${ns_rc}"
     } else {
-      sh script: """./jenkins/node_provision_start.py \
-                     --nodes=${nodeString} ${options}""",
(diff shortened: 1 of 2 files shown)


vars/provisionNodes.groovy
```diff
@@ -157,6 +157,10 @@ def call(Map config = [:]) {
 
   sshagent (credentials: ['daos-provisioner']) {
 
+    sh script:'clush -l root -w ' + nodeString +
+              ' --connect_timeout 30' +
+              ' "ls -lh /tmp/*.log; rm -f /tmp/daos.log /tmp/server.log"'
+        returnStatus: true
```
brianjmurrell: If you are going to `returnStatus: true` you probably want `clush`'s `-S` switch.

JohnMalmberg: Well, I was just preventing failures such as nodes unreachable from failing the script, so the status did not matter.

I hadded some println statements.

At the time that enter into routine, any damage has already been done, so all we can do is try to collect some information about the issue and then clean it up.

To fix the issue, we need to make sure that we have exit/abort handlers that can do diagnostics and cleanup in the test stages.

brianjmurrell: > Well, I was just preventing failures such as nodes unreachable from failing the script, so the status did not matter.

But you will want that `clush` command (and the provisioning on the whole) to fail if an `rm -f` fails to remove a log file won't you?

> At the time that enter into routine, any damage has already been done, so all we can do is try to collect some information about the issue and then clean it up.

To fix the issue, we need to make sure that we have exit/abort handlers that can do diagnostics and cleanup in the test stages.

￼Agreed.  But this will paper over any such issue.  How will we know there is even an issue to resolve if we just silently clean up here?  Perhaps finding a log file to remove here needs to notify somebody/thing in some way?

JohnMalmberg: When we detect the problem here, deleting the file is the first attempt to restore things.
If it we are unable to delete the files, then provisioning is going to fail so it really does not matter if we fail first here.

This is mainly stop-gap to give us some breathing room to get the real problems fixed.

JohnMalmberg: We can have a followup ticket to setup a notification when something is detected.

brianjmurrell: > provisioning is going to fail so it really does not matter if we fail first here.

But if we fail it here, we can fail it with a useful error message so that we don't add to the ongoing and steady stream of "why did this fail?" queries.  We need to aim to provide better failure messages for developers, so that they can determine why things fail on their own.

JohnMalmberg: This is all I am going to have time for today.  I am not sure when I will be able to do all that you are asking.  Let's get this in now, and follow up with a design for something better.

brianjmurrell: #96 based on this PR which can land into it to provide that better error reporting to developers.

PR #102, opened 2019-12-13
CORCI-752 build Coverity scan updates

JohnMalmberg:
Add coverityToolDownload as a separate step.
Add an optional Coverity buiding step to packageBuildingPipelineDAOS

### DIFF
--- a/vars/packageBuildingPipelineDAOS.groovy
+++ b/vars/packageBuildingPipelineDAOS.groovy
@@ -139,6 +139,56 @@ def call(Map pipeline_args) {
             } //stage('Lint')
             stage('Build') {
                 parallel {
+                    stage('Coverity') {
+                        when {
+                            beforeAgent true
+                            allOf {
+                                expression { pipeline_args.get('coverity','') != '' }
+                            }
+                        }
+                        agent {
+                            dockerfile {
+                                filename 'packaging/Dockerfile.coverity'
+                                label 'docker_runner'
+                                args '--privileged=true'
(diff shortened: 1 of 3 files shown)


JohnMalmberg:
Daos test job: https://build.hpdd.intel.com/job/daos-stack/job/daos/view/change-requests/job/PR-1574/

JohnMalmberg:
Raft test job https://build.hpdd.intel.com/job/daos-stack/job/raft/view/change-requests/job/PR-29/

vars/packageBuildingPipelineDAOS.groovy
```diff
@@ -139,6 +139,56 @@ def call(Map pipeline_args) {
             } //stage('Lint')
             stage('Build') {
                 parallel {
+                    stage('Coverity') {
+                        when {
+                            beforeAgent true
+                            allOf {
+                                expression { pipeline_args.get('coverity','') != '' }
```
brianjmurrell: This would match on `false` wouldn't it?  Maybe we should be looking for `true` explicitly so that one can toggle it on and off with `true` and `false` and not have to delete the option altogether?

JohnMalmberg: The parameter needs to be a Coverity project name, true and false are both invalid,  So either no parameter, or a parameter of '' means no Coverity build.

brianjmurrell: Ah, yes, I see that below.

PR #106, opened 2020-01-23
CORCI-768 build: Add CI profile support

JohnMalmberg:
Add CI profile support for el7 and sles15, opensuse15.
Remove SLES-12 support from provisionNodes.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,5 +1,5 @@
 #!/usr/bin/env groovy
-// Copyright (c) 2018 Intel Corporation
+// Copyright (c) 2018-2020 Intel Corporation
 //
 // Permission is hereby granted, free of charge, to any person obtaining a
 // copy of this software and associated documentation files (the "Software"),
@@ -26,6 +26,7 @@
 // is landed, both PR branches can be deleted.
 
 // @Library(value="pipeline-lib@my_pr_branch") _
+@Library(value="pipeline-lib@corci-768") _
 
 pipeline {
     agent { label 'lightweight' }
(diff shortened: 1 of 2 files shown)


Jenkinsfile
```diff
@@ -90,6 +91,7 @@ pipeline {
                 steps {
                     provisionNodes NODELIST: env.NODELIST,
                                    distro: 'el7',
+                                   profile: 'ci',
```
brianjmurrell: Should we be future-looking here and call this something like `ci-daos` in anticipation of having other profiles for CI for other projects that need a different image?

Jenkinsfile
```diff
@@ -131,7 +132,8 @@ pipeline {
                 }
                 steps {
                     provisionNodes NODELIST: env.NODELIST,
-                                   distro: 'sles12sp3',
+                                   distro: 'opensuse15',
+                                   profile: 'ci',
```
brianjmurrell: Ditto

vars/provisionNodes.groovy
```diff
@@ -328,7 +319,7 @@ EOF'''
                                                    libcmocka python-pathlib     \
                                                    python2-numpy git            \
                                                    python2-clustershell         \
-                                                   golang-bin'''
+                                                   golang-bin ipmctl'''
```
brianjmurrell: I guess all of the above are not in the profile image (yet)?  Will they be?

Jenkinsfile
JohnMalmberg: This is testing the new CI profile to make sure it works.
As we add more profiles to support we can convert it to a matrix and manage the names.
It will probably be a while before we add more profiles.

vars/provisionNodes.groovy
JohnMalmberg: They are in the new "ci" profiles for el7 and opensuse15, but not all are known to be in the existing "test" profile in production.
We have some challenges for el8 support that will need to be worked out before then.

Jenkinsfile
JohnMalmberg: Created daos_ci profile.

PR #110, opened 2020-02-21
CORCI-854 build: Support Snapshot restores.

JohnMalmberg:
Change to default to always doing snapshot restores.

### DIFF
--- a/vars/provisionNodes.groovy
+++ b/vars/provisionNodes.groovy
@@ -140,10 +140,6 @@ host wolf-*
     UserKnownHostsFile /dev/null
     LogLevel error
 EOF'''
-  if (distro_type == "suse") {
-    provision_script += '\nclush -B -S -l root -w ' + nodeString +
-                        ' "zypper --non-interactive install sudo"'
-  }
   provision_script += '\nclush -B -l root -w ' + nodeString + ' -c' +
                     ''' ci_key* --dest=/tmp/
                         clush -B -S -l root -w ''' + nodeString +
@@ -282,7 +278,8 @@ EOF'''
     }
     provision_script += '\nzypper --non-interactive' +
                         ' --gpg-auto-import-keys --no-gpg-checks ref'
-    provision_script += '\nzypper --non-interactive in ed nfs-client ipmctl ndctl'
+    provision_script += '\nzypper --non-interactive in' +
+                        ' sudo ed nfs-client ipmctl ndctl'
     if (inst_rpms) {
       provision_script += ' ' + inst_rpms
     }
--- a/vars/provisionNodes.groovy
+++ b/vars/provisionNodes.groovy
@@ -55,47 +55,10 @@ def call(Map config = [:]) {
       println "Jenkins not configured for CI provisioning."
       return node_cnt
   }
-
-  // corci-792 assist code.
-  // corci-792 will convert the VM in a physical CI allocation to be
-  // a physical node, so we want to skip the snapshot restore when
-  // the only node is not a VM.
-  if (config['snapshot'] && (config['NODELIST'].indexOf('vm') < 0)) {
-    println "Skipping re-provisioning snapshot for physical nodes."
-    return 0
-  }
-
-  checkoutScm url: 'git@gitlab.hpdd.intel.com:lab/ci-node-mgmt.git',
-                                  checkoutDir: 'jenkins',
-                                  credentialsId: 'daos-special-read'
-  def options = ''
-  def snapshot = ''
-  def distro = 'el7'
-  def wait_for_it = true
-  def inst_rpms = ''
-  def inst_repos = ''
-  if (config['snapshot']) {
-    options += ' --snapshot'
-    wait_for_it = false
-  }
-  if (config['arch']) {
-    options += " --arch=${config['arch']}"
-  }
-  if (config['distro']) {
-    options += " --distro=${config['distro']}"
-    distro = config['distro']
-  }
-  if (config['profile']) {
-    options += " --profile=${config['profile']}"
-  }
-
-  if (config['inst_rpms']) {
-      inst_rpms = config['inst_rpms']
-  }
-
-  if (config['inst_repos']) {
-      inst_repos = config['inst_repos']
-  }
+  Map new_config = config
+  def distro = config.get('distro', 'el7')
+  def inst_rpms = config.get('inst_rpms', '')
+  def inst_repos = config.get('inst_repos','')
 
   def repository_g = ''
   def repository_l = ''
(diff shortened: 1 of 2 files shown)


vars/provisionNodes.groovy
```diff
@@ -149,7 +112,7 @@ def call(Map config = [:]) {
   }
 
   def cleanup_logs = 'clush -l root -w ' + nodeString +
```
brianjmurrell: Let's add a `-B` to this `clush` also.

vars/provisionNodes.groovy
```diff
@@ -163,236 +126,177 @@ def call(Map config = [:]) {
                       if [ -d /tmp/Functional_*/ ]; then
                           rm -rf /tmp/Functional_*
                       fi'''
+  new_config['pre_reboot'] = cleanup_logs
+  new_config['post_reboot'] = cleanup_logs
```
brianjmurrell: Do we actually need these if we are provisioning nodes fresh every time?

brianjmurrell: But ultimately, any post-test state will be wiped by the re-provision so i'm not sure i see the value here.

vars/provisionNodes.groovy
```diff
@@ -163,239 +126,180 @@ def call(Map config = [:]) {
...
+  def provision_script = '''set -ex
+                            rm -f ci_key*
+                            ssh-keygen -N "" -f ci_key
+                            cat << "EOF" > ci_key_ssh_config
 host wolf-*
     CheckHostIp no
     StrictHostKeyChecking no
     UserKnownHostsFile /dev/null
     LogLevel error
 EOF'''
```
brianjmurrell: This **is** something that should go into a system-pipeline-lib, or some other _site-setting_.

Not suggesting it needs to be fixed here.  Just sayin'.

vars/provisionNodes.groovy
```diff
@@ -163,239 +126,180 @@ def call(Map config = [:]) {
...
-                                   repo=\\"\\\${repo%@*}\\"
-                                   if [[ \\\$branch = *:* ]]; then
-                                       build_number=\\"\\\${branch#*:}\\"
-                                       branch=\\"\\\${branch%:*}\\"
-                                   fi
-                               fi'''
-    if (distro.startsWith("el7")) {
-       // Since we don't have CORCI-711 yet, erase things we know could have
-       // been put on the node previously
+  provision_script += '\nclush -B -l root -w ' + nodeString + ' -c' +
```
brianjmurrell: Any reason to end the previous `provision_script` assignment on line 142 and then just append to it here?  Why not just keep it going in the same `'''`?

vars/provisionNodes.groovy
```diff
@@ -163,239 +126,180 @@ def call(Map config = [:]) {
...
+                           if [[ \\\$repo = *@* ]]; then
+                             branch=\\"\\\${repo#*@}\\"
+                             repo=\\"\\\${repo%@*}\\"
+                             if [[ \\\$branch = *:* ]]; then
+                               build_number=\\"\\\${branch#*:}\\"
+                               branch=\\"\\\${branch%:*}\\"
+                             fi
+                           fi'''
+  if (distro.startsWith("el7")) {
+    if (config['power_only']) {
```
brianjmurrell: Do we need to keep this?

vars/provisionNodes.groovy
```diff
@@ -163,239 +126,180 @@ def call(Map config = [:]) {
...
+                                                 python36-nose                \
+                                                 python36-pip valgrind        \
+                                                 python36-paramiko            \
+                                                 python2-avocado              \
+                                                 python2-avocado-plugins-output-html \
+                                                 python2-avocado-plugins-varianter-yaml-to-mux \
+                                                 libcmocka python-pathlib     \
+                                                 python2-numpy git            \
+                                                 python2-clustershell         \
+                                                 golang-bin ipmctl ndctl'''
```
brianjmurrell: I'd like to see another flavour of CI image one day that includes all of this static stuff that we take time to install on every test run.

vars/provisionNodes.groovy
```diff
@@ -163,239 +126,180 @@ def call(Map config = [:]) {
...
-    def rc = 0
-    rc = sh(script: provision_script,
-            label: "Post provision configuration",
-            returnStatus: true)
-    if (rc != 0) {
-      stepResult name: env.STAGE_NAME, context: "test", result: "FAILURE"
-      error "One or more nodes failed post-provision configuration!"
+    if (inst_repos) {
+      provision_script +=   '\n' + iterate_repos +
+                          '''\n    zypper --non-interactive ar --gpgcheck-allow-unsigned -f ''' + env.JENKINS_URL + '''job/daos-stack/job/\\\${repo}/job/\\\${branch//\\//%252}/\\\${build_number}/artifact/artifacts/leap15/ \\\$repo
```
brianjmurrell: Looks like the _F_ got dropped here.  This line should just not change in this patch.  This reverts this change:
```suggestion
                          '''\n    zypper --non-interactive ar --gpgcheck-allow-unsigned -f ''' + env.JENKINS_URL + '''job/daos-stack/job/\\\${repo}/job/\\\${branch//\\//%252F}/\\\${build_number}/artifact/artifacts/leap15/ \\\$repo
```

vars/provisionNodes.groovy
```diff
@@ -163,239 +126,180 @@ def call(Map config = [:]) {
...
+                               done'''
+    }
+    provision_script += '\nzypper --non-interactive' +
+                        ' --gpg-auto-import-keys --no-gpg-checks ref'
+    provision_script += '\nzypper --non-interactive in' +
+                        ' ed nfs-client ipmctl ndctl sudo '
+    if (inst_rpms) {
+      provision_script += ' ' + inst_rpms
```
brianjmurrell: There is a space trailing from line 268 already.
```suggestion
      provision_script += inst_rpms
```

vars/provisionNodes.groovy
brianjmurrell: What do you mean by _testing in flight_?

vars/provisionNodes.groovy
brianjmurrell: > All this stuff is being baked into the new CI packer image.

Excellent.  So can/should it be removed from here?

> Eventually we want to be using the information in the RPM or other package to make sure that the dependencies are present instead of having to code them in here

I'm not sure all of those dependencies will ever be in one or more RPMs.

JohnMalmberg: This needs to be present until daos pr-2008 is landed and all existing existing PRs updated to use that commit.

We should have some type of file in the daos repository and or some sort of tool that will report to us all the dependent packages needed for testing DAOS.

PR #118, opened 2020-03-04
Don't grep non-existent file patterns

brianjmurrell:
Otherwise it returns an error and confuses the result with grep not
finding any matches.


brianjmurrell:
Passing test result with this PR using itself is in https://build.hpdd.intel.com/blue/organizations/jenkins/daos-stack%2Fpipeline-lib/detail/PR-118/30/pipeline.

PR #120, opened 2020-03-20
CORCI-887 build: Matrix parameters are strings

JohnMalmberg:
Matrix parameters which need to be passed to node provisioning
are strings, so need to be used as int.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -157,7 +157,7 @@ pipeline {
                     provisionNodes NODELIST: env.NODELIST,
                                    distro: 'el7',
                                    profile: 'daos_ci',
-                                   node_count: 1,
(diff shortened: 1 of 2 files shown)


JohnMalmberg:
Test run 5 is the one that actually tests the fixed code.

Jenkinsfile
```diff
@@ -180,7 +180,7 @@ pipeline {
                             status: 'SUCCESS')
                     }
                 }
-            } //stage('provisionNodes with slurm EL7')
+            } //stage('provisionNodes withl release/0.9 Repo')
```
brianjmurrell:
```suggestion
            } //stage('provisionNodes with release/0.9 Repo')
```

PR #122, opened 2020-03-27
Don't install fuse in post-provisioning

brianjmurrell:
DAOS depends on it so no need to install it explicitly.

### DIFF
--- a/vars/provisionNodes.groovy
+++ b/vars/provisionNodes.groovy
@@ -236,16 +236,16 @@ EOF'''
                              rm -f /tmp/daos_control.log
                              yum -y erase metabench mdtest simul IOR compat-openmpi16
                              yum -y install epel-release
-                             if ! yum -y install CUnit fuse python36-PyYAML   \
-                                                 python36-nose                \
-                                                 python36-pip valgrind        \
-                                                 python36-paramiko            \
-                                                 python2-avocado              \
-                                                 python2-avocado-plugins-output-html \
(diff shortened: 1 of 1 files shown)


vars/provisionNodes.groovy
```diff
@@ -236,16 +236,16 @@ EOF'''
                              rm -f /tmp/daos_control.log
                              yum -y erase metabench mdtest simul IOR compat-openmpi16
                              yum -y install epel-release
-                             if ! yum -y install CUnit fuse python36-PyYAML   \
```
JohnMalmberg: What happens for the non RPM test cases that need fuse3 installed?

Would it be better to install fuse3 instead  fuse here until we can switch to all RPM testing?

brianjmurrell: > What happens for the non RPM test cases that need fuse3 installed?

LIke what?  I did [run this change through a DAOS run](https://build.hpdd.intel.com/job/daos-stack/job/daos/job/PR-2232/1/) and there were no issues related to a missing _fuse_ package.

> Would it be better to install fuse3 instead fuse here until we can switch to all RPM testing?

DAOS (and therefore all component) RPMs are installed on all of the nodes even without us doing full RPM testing, so that brings in `fuse` ([soon to be `fuse3`](daos-stack/daos#2224)) already.  In any case if we needed to, specifying installing `fuse3` belongs in the Jenkinsfile, not the (supposed to be generic) post-provisioning configuration here.

PR #126, opened 2020-04-13
Add a runTestFunctional step

brianjmurrell:

### DIFF
--- a/vars/runTest.groovy
+++ b/vars/runTest.groovy
@@ -24,8 +24,10 @@ def call(Map config = [:]) {
     dir('install') {
         deleteDir()
     }
-    config['stashes'].each {
-        unstash it
+    if (config['stashes']) {
+        config['stashes'].each {
+            unstash it
+        }
     }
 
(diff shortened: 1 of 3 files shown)


brianjmurrell:
@JohnMalmberg A step in the direction of factoring-down our Jenkinsfile.  It was used in https://build.hpdd.intel.com/blue/organizations/jenkins/daos-stack%2Fdaos/detail/PR-2407/112/pipeline/, FWIW.

brianjmurrell:
At some point, if we need to further factor-down Jenkinsfile, we could pull the node provisioning into this function.  But that seems like a future step, if/when needed.

brianjmurrell:
> Looks like a good first step.

Yeah, I think there is a bunch more of this we can attack.

> I still want to move the inline scripts into the DAOS repository.

Yeah, I actually did have a go at that with the inline script in this one.  The result was not terribly great because the script has some inline script within it inside the `clush` commands.  They end up confusing ShellCheck.

PR #133, opened 2020-05-01
Remove the install/ dir for RPM testing

brianjmurrell:
To make sure that we are testing from the RPM installed files.


brianjmurrell:
Much thanks @JohnMalmberg!

PR #134, opened 2020-05-01
DAOS-623 build: Add ability to bypass clean

jolivier23:
In order to test multiple build types, we don't want to
rebuild everything so enable ability to skip clean step

Signed-off-by: Jeff Olivier <jeffrey.v.olivier@intel.com>


jolivier23:
> I am trying to understand the use case.
> Are you planning on adding stages that do multiple sconsBuild steps in sequence?

I tested this here:

https://github.com/daos-stack/daos/pull/2601

I do clean on first build but then rebuild with different BUILD_TYPE option.  I don't want to rebuild same prerequisites in such case but rather reuse them

jolivier23:
Verified to be working now with https://github.com/daos-stack/daos/pull/2601

PR #138, opened 2020-05-14
CORCI-27 test: Allow the setting of a codespell file for checkpatch

ashleypittman:
Add an option to provide a path to a codespell file to catch frequent spelling mistakes.

### DIFF
--- a/vars/checkPatch.groovy
+++ b/vars/checkPatch.groovy
@@ -54,7 +54,7 @@ def call(Map config = [:]) {
 
     def checkpatch_args="--notree --show-types -"
     if (config['codespell_file']) {
-       checkpatch_args="--notree --show-types --codespell --codespell-file=" + config['codespell_file'] + " -"
+       checkpatch_args="--notree --show-types --codespell --codespellfile=" + config['codespell_file'] + " -"
     }
 
     int rc = 1
--- a/vars/checkPatch.groovy
+++ b/vars/checkPatch.groovy
@@ -58,8 +58,8 @@ def call(Map config = [:]) {
     }
 
     int rc = 1
-    def script = 'CHECKPATCH_IGNORED_FILES="' + ignored_files + '";' + \
-                 'CHECKPATCH_ARGS="' + checkpatch_args + '";' + \
+    def script = 'CHECKPATCH_IGNORED_FILES="' + ignored_files + '"' + \
+                 ' CHECKPATCH_ARGS="' + checkpatch_args + '"' + \
                  ' code_review/jenkins_github_checkwarn.sh'
 
     if (config['review_creds']) {
--- a/vars/checkPatch.groovy
+++ b/vars/checkPatch.groovy
@@ -14,6 +14,7 @@
  * config['ignored_files'] is colon delimited string of files not to check.
  * config['review_creds'] is the credentials needed for a code review.
  * config['branch'] is the code_review repo branch to use.
+ * config['codespell_file'] is the name of a codespell dictionary to use
  *
  */
 def call(Map config = [:]) {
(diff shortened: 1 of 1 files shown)


vars/checkPatch.groovy
```diff
@@ -51,8 +52,14 @@ def call(Map config = [:]) {
...
     }
 
+    def checkpatch_args="--notree --show-types -"
+    if (config['codespell_file']) {
+       checkpatch_args="--notree --show-types --codespell --codespellfile=" + config['codespell_file'] + " -"
+    }
+
     int rc = 1
     def script = 'CHECKPATCH_IGNORED_FILES="' + ignored_files + '"' + \
+                 ' CHECKPATCH_ARGS="' + checkpatch_args + '"' + \
```
ashleypittman, line 62: This does work, but the script this function calls is jenkins_github_checkwarn.sh which I seems to build a default CHECKPATCH_ARGS if one isn't set, and this is where the ignored_files option is used, so I think this PR probably disables that option.

An alternative would be to modify github_checkpatch.py to insert the codespell options there, based on availability of the package.

JohnMalmberg: I wanted the checkpatch stage and scripts to be configured to use optional files to control what is checked and what is ignored, sort of like the pylint rc files, rather than passing a list to it.

I have not looked at the current check patch code in a few years now so I need to look up the impact of this change.

ashleypittman: Yes, it's far from clean the way checkpatch currently works.  Trying a different approach elsewhere but I do appreciate advise on the best way forward here.

vars/checkPatch.groovy
```diff
@@ -51,8 +52,14 @@ def call(Map config = [:]) {
         ignored_files += ":" + config['ignored_files']
     }
 
+    def checkpatch_args="--notree --show-types -"
```
brianjmurrell, line 55:
```suggestion
    def checkpatch_args = "--notree --show-types -"
```

vars/checkPatch.groovy
```diff
@@ -51,8 +52,14 @@ def call(Map config = [:]) {
...
+    if (config['codespell_file']) {
+       checkpatch_args="--notree --show-types --codespell --codespellfile=" + config['codespell_file'] + " -"
```
brianjmurrell, line 57:
```suggestion
       checkpatch_args += "-codespell --codespellfile=" + config['codespell_file'] + " -"
```
Just to DRY up a little bit.

vars/checkPatch.groovy
brianjmurrell: `code_review/jenkins_github_checkwarn.sh` already has _codespell_ support in it?

vars/checkPatch.groovy
brianjmurrell, line 55: What are these extra (new in this patch) arguments going to do/change in the checkpatch behaviour?

vars/checkPatch.groovy
ashleypittman: checkpatch.pl already has support in it, it just needs the name of the file on the command line.

vars/checkPatch.groovy
ashleypittman: The way these args are set is a bit messed up, but the - means "stdin" so has to come last in the list.

vars/checkPatch.groovy
ashleypittman: They're the default options from code_review/jenkins_github_checkwarn.sh, so if this code sets anything, it needs to add those back in (and the ignore list which is why I don't like this approach).

ashleypittman:
Please review https://github.com/daos-stack/code_review/pull/21 for what I think is a better approach.

vars/checkPatch.groovy
brianjmurrell: Indeed.  My suggestion preserves that if I am not mistaken.

But in any case, you are proposing abandoning this for daos-stack/code_review#21 correct?

ashleypittman: I think that approach works better overall, yes.

PR #139, opened 2020-05-15
CORCI-918 build: Prepaare for daos refactor

JohnMalmberg:
* Jenkinsfile: Refactored a bit to match future Daos Jenkinsfile.

* README.md: Added some actual documentation.

* cancelPreviousBuilds: Try trusted system pipeline lib,
                        Whitelisted methods, or log message if can't do.

* checkPatch: Use scmNotify instead of githubNotify.

* commitPragma: Fix typo.

* coverityToolDownload: Fix documentation comment.

* emailext: Remomve uneeded log message.

* gitnubNotify: Log that it is redirected to scmNotify.

* provsionNodes: Fix documentation.
                 Remove dependency on UID environment variable.

* publishToRepository: Fix documentation typo.

* publishVagrind: Remove uneeded log message.

* runTest: Use scmNotify

* scmNotify: New file, calls scmNotifySystem if available.
             If scmNotifySystem is not avialable, then no scm is
             setup for notification.

* sconsBuild: Use scmNotify

* stepResult: Use scmNotify, allow common context variable to be used.


vars/scmNotify.groovy
```diff
@@ -0,0 +1,31 @@
...
+   */
+
+def call(Map config = [:]) {
+
+  if (env.DAOS_JENKINS_NOTIFY_STATUS == null) {
+    println "Jenkins not configured to notify SCM repository of builds."
+    return
+  }
+  try {
+    scmNotifySystem(config)
```
brianjmurrell, line 27: Why is this needed?  I'm not OK with removing even more functionality from the open code and burying it into closed modules, particularly when it results in the open code becoming less functional that it was before.

JohnMalmberg: This does three purposes per the README.md file.
1. It makes the Jenkinsfile just work on systems that use our repository but do not have permission to notify it with out needing a Jenkins environment variable to inhibit updates.
2. It isolates if building on systems that are not checked out from Github that can still get feedback, such as Gitlab if the site chooses to provide their own additional pipeline library.
3. It removes requiring the implementation that other sites use the same credential-id for the scm access and at the same time removes the need to expose the credential-id.

brianjmurrell: But it also disables notifying a GitHub fork that somebody might make and want to duplicate our CI workflow.  In that regard, this is a functionality regression.

But also, if we have code to update either a GitHub fork or a fork in a Gitlab, why wouldn't we want the open project to benefit from that code instead of hiding it away?

brianjmurrell: And as I have said before, _hiding_ credential-ids is security-through-obscurity (which means little to no security and should not be relied on as it only provides a false sense of security).  The security mechanism that should be applied to the use of credential-ids is pre-execution review gating of PRs, not obscurity.

JohnMalmberg: It only disables notifying a Github fork if the end user has not setup their own additional pipeline-lib.

Previous implementation restricted the SCM hosting to GitHub and also the required the end user to use the same Jenkins credential ID as we were using.  Or it required them setting an environment variable.

This patch removes those restrictions, making the project more open and portable.

JohnMalmberg: Not the forum for discussing security issues.

JohnMalmberg: Changed code to first try the scmNotifySystem library and if it is not present, use the environment variable that enables SCM reporting as the source of a credential-id that is used for a githubNoify call, to allow an alternate way for a site to be configured.

Jenkinsfile
```diff
@@ -26,16 +17,72 @@
...
 pipeline {
     agent { label 'lightweight' }
     environment {
-        BAHTTPS_PROXY = "${env.HTTP_PROXY ? '--build-arg HTTP_PROXY="' + env.HTTP_PROXY + '" --build-arg http_proxy="' + env.HTTP_PROXY + '"' : ''}"
-        BAHTTP_PROXY = "${env.HTTP_PROXY ? '--build-arg HTTPS_PROXY="' + env.HTTPS_PROXY + '" --build-arg https_proxy="' + env.HTTPS_PROXY + '"' : ''}"
-        UID = sh(script: "id -u", returnStdout: true)
         SSH_KEY_FILE='ci_key'
         SSH_KEY_ARGS="-i$SSH_KEY_FILE"
         CLUSH_ARGS="-o$SSH_KEY_ARGS"
+        BUILDARGS = docker_build_args()
```
brianjmurrell, line 81:
```suggestion
```

Jenkinsfile
```diff
@@ -112,7 +157,7 @@ pipeline {
                     dockerfile {
                         filename 'docker/Dockerfile.centos.7'
                         label 'docker_runner'
-                        additionalBuildArgs  '--build-arg UID=$(id -u)'
+                        additionalBuildArgs  '$BUILDARGS'
```
brianjmurrell, line 160: Can the function be used directly here?
```suggestion
                        additionalBuildArgs  docker_build_args()                       
```

If so, there is no need to set an env. var.

Jenkinsfile
```diff
@@ -198,22 +239,21 @@ pipeline {
                                    inst_repos: "daos@master"
                     runTest script: '''NODE=${NODELIST%%,*}
                                        ssh $SSH_KEY_ARGS jenkins@$NODE "set -ex
-                                       yum --disablerepo=\\* --enablerepo=build\\* makecache"''',
+                                       yum --disablerepo=\\* \
+                                         --enablerepo=build\\* makecache"''',
```
brianjmurrell, line 243:
```suggestion
                                            --enablerepo=build\\* makecache"''',
```
to line up with the arg. on the previous line.

vars/githubNotify.groovy
```diff
@@ -1,20 +1,20 @@
 // vars/githubNotifiy.groovy
 
 /**
- * run.groovy
+ * gethubNotify.groovy
```
brianjmurrell, line 4:
```suggestion
 * githubNotify.groovy
```

vars/scmNotify.groovy
```diff
@@ -0,0 +1,39 @@
...
+
+def call(Map config = [:]) {
+
+  def errtxt = 'Jenkins not configured to notify SCM repository of builds.'
+  if (env.DAOS_JENKINS_NOTIFY_STATUS == null) {
+    println errtxt
+    return
+  }
+  try {
+    scmNotifySystem(config)
```
brianjmurrell, line 28: Is there any reason not to just pull vars/scmNotifySystem.groovy￼ up to here so that it adds the flexibility for somebody else to use gitlab instead of github?

vars/stepResult.groovy
```diff
@@ -23,6 +23,7 @@ def call(Map config= [:]) {
    * config['junit_files'] The names of any available junit files.
    */
 
+    // node block runs this on different build agent
```
brianjmurrell, line 26: Did you try removing the `node { }` to see why this needs one?  Perhaps/probably, we can narrow down the scope of the need for it, or remove it altogether armed with our more advanced knowledge of Jenkins Pipelines now.

vars/scmNotify.groovy
JohnMalmberg: The gitlab notify in system-pipeline-lib will only work for our internal gitlab jobs currently.

I have not looked into what it would take to make it work with DAOS.

Jenkinsfile
JohnMalmberg: Will be in next push

vars/githubNotify.groovy
JohnMalmberg: Will be in next push.

vars/scmNotify.groovy
brianjmurrell: Really?  It looks like the very generic equivalent of it's `githubNotify` counterpart.  What is specific to our internal gitlab?

vars/stepResult.groovy
brianjmurrell: I think you mentioned in IM that the `node { }` isn't needed, or am I mis-recalling?

vars/scmNotify.groovy
JohnMalmberg: Now a wrapper to scmNotifyTrusted in trusted-pipeline-lib

PR #145, opened 2020-05-27
Make commitPragma return a String

brianjmurrell:


brianjmurrell:
Can't overload based on return types only.

PR #151, opened 2020-07-08
Enable valgrind option for unitTest and unitTestPost

marcelarosalesj:
For CORCI-951 we want to enable run_test.sh stage to run with valgrind
memcheck tool. As these 'run_test.sh' and 'run_test.sh with memcheck' stages
are very similar, this commit intends to refactor code from Jenkinsfile.

Signed-off-by: Marcela Rosales <marcela.a.rosales.jimenez@intel.com>

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -15,7 +15,6 @@
 // That PR should be landed with out deleting the PR branch.
 // Then a second PR submitted to comment out the @Library line, and when it
 // is landed, both PR branches can be deleted.
-
 //@Library(value="pipeline-lib@my_branch_name") _
 @Library(value="pipeline-lib@marosale/CORCI-951-refactor") _
 
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -16,7 +16,6 @@
 // Then a second PR submitted to comment out the @Library line, and when it
 // is landed, both PR branches can be deleted.
 //@Library(value="pipeline-lib@my_branch_name") _
-@Library(value="pipeline-lib@marosale/CORCI-951-refactor") _
 
 // The docker agent setup may need to know the
 // UID that the build agent is running under.
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -16,6 +16,7 @@
 // Then a second PR submitted to comment out the @Library line, and when it
 // is landed, both PR branches can be deleted.
 
+//@Library(value="pipeline-lib@my_branch_name") _
 @Library(value="pipeline-lib@marosale/CORCI-951-refactor") _
 
 // The docker agent setup may need to know the
--- a/vars/parseStageInfo.groovy
+++ b/vars/parseStageInfo.groovy
@@ -116,7 +116,7 @@ def call(Map config = [:]) {
     if (env.STAGE_NAME.contains('run_test.sh')) {
       result['junit_files'] = 'test_results/*.xml'
       result['artifacts'] = ['run_test.sh/*', 'vm_test/**']
-      result['with_valgrind'] = 'disabled'
+      result['with_valgrind'] = ''
       result['valgrind_pattern'] = 'dnt.*.memcheck.xml'
 
       if (config['junit_files']) {
--- a/vars/parseStageInfo.groovy
+++ b/vars/parseStageInfo.groovy
@@ -146,10 +146,9 @@ def call(Map config = [:]) {
       result['ftest_arg'] = config['ftest_arg']
     }
 
-    if (env.STAGE_NAME.contains('Unit Test')) {
-      if(env.STAGE_NAME.contains('memcheck')) {
+    if (env.STAGE_NAME.contains('Unit Test') &&
+        env.STAGE_NAME.contains('memcheck')) {
         result['with_valgrind'] = 'memcheck'
-      }
     }
 
     return result
--- a/vars/parseStageInfo.groovy
+++ b/vars/parseStageInfo.groovy
@@ -147,8 +147,6 @@ def call(Map config = [:]) {
     }
 
     if (env.STAGE_NAME.contains('Unit Test')) {
-      result['with_valgrind'] = ''
-
       if(env.STAGE_NAME.contains('memcheck')) {
         result['with_valgrind'] = 'memcheck'
       }
--- a/vars/parseStageInfo.groovy
+++ b/vars/parseStageInfo.groovy
@@ -134,7 +134,6 @@ def call(Map config = [:]) {
 
       if (result['with_valgrind'] == 'memcheck') {
         result['valgrind_pattern'] = 'run_test_memcheck.sh/*memcheck.xml'
-        result['junit_files'] = ''
         result['artifacts'] = ['run_test_memcheck.sh/*']
       }
     }
--- a/vars/parseStageInfo.groovy
+++ b/vars/parseStageInfo.groovy
@@ -113,28 +113,31 @@ def call(Map config = [:]) {
       result['ftest_arg'] = config['ftest_arg']
     }
 
-    result['junit_files'] = 'test_results/*.xml'
-    if (config['junit_files']) {
-      result['junit_files'] = config['junit_files']
+    if (env.STAGE_NAME.contains('run_test.sh')) {
+      result['junit_files'] = 'test_results/*.xml'
+      result['artifacts'] = ['run_test.sh/*', 'vm_test/**']
+      result['with_valgrind'] = 'disabled'
+      result['valgrind_pattern'] = 'dnt.*.memcheck.xml'
+
+      if (config['junit_files']) {
+        result['junit_files'] = config['junit_files']
+      }
+      if (config['artifacts']) {
+        result['artifacts'] = config['artifacts']
+      }
+
+      if (config['with_valgrind']) {
+        result['with_valgrind'] = config['with_valgrind']
+      } else if(env.STAGE_NAME.contains('memcheck')) {
+        result['with_valgrind'] = 'memcheck'
+      }
+
+      if (result['with_valgrind'] == 'memcheck') {
+        result['valgrind_pattern'] = 'run_test_memcheck.sh/*memcheck.xml'
+        result['junit_files'] = ''
+        result['artifacts'] = ['run_test_memcheck.sh/*']
+      }
     }
 
-    result['artifacts'] = ['run_test.sh/*', 'vm_test/**']
-    if (config['artifacts']) {
-      result['artifacts'] = config['artifacts']
-    }
-
-    result['valgrind'] = 'disabled'
-    if (config['valgrind']) {
-      result['valgrind'] = config['valgrind']
-    } else if(env.STAGE_NAME.contains('memcheck')) {
-      result['valgrind'] = 'memcheck'
-    }
-
-    result['valgrind_pattern'] = 'dnt.*.memcheck.xml'
-    if (result['valgrind'] == 'memcheck') {
-      result['valgrind_pattern'] = 'run_test_memcheck.sh/*memcheck.xml'
-      result['junit_files'] = ''
-      result['artifacts'] = ['run_test_memcheck.sh/*']
-    }
     return result
 }
--- a/vars/parseStageInfo.groovy
+++ b/vars/parseStageInfo.groovy
@@ -146,28 +146,11 @@ def call(Map config = [:]) {
       result['ftest_arg'] = config['ftest_arg']
     }
 
-    if (env.STAGE_NAME.contains('run_test.sh')) {
-      result['junit_files'] = 'test_results/*.xml'
-      result['artifacts'] = ['run_test.sh/*', 'vm_test/**']
+    if (env.STAGE_NAME.contains('Unit Test')) {
       result['with_valgrind'] = ''
-      result['valgrind_pattern'] = 'dnt.*.memcheck.xml'
 
-      if (config['junit_files']) {
-        result['junit_files'] = config['junit_files']
-      }
-      if (config['artifacts']) {
-        result['artifacts'] = config['artifacts']
-      }
-
-      if (config['with_valgrind']) {
-        result['with_valgrind'] = config['with_valgrind']
-      } else if(env.STAGE_NAME.contains('memcheck')) {
+      if(env.STAGE_NAME.contains('memcheck')) {
         result['with_valgrind'] = 'memcheck'
-      }
-
-      if (result['with_valgrind'] == 'memcheck') {
-        result['valgrind_pattern'] = 'run_test_memcheck.sh/*memcheck.xml'
-        result['artifacts'] = ['run_test_memcheck.sh/*']
       }
     }
 
--- a/vars/parseStageInfo.groovy
+++ b/vars/parseStageInfo.groovy
@@ -52,8 +52,20 @@ def call(Map config = [:]) {
       }
     }
 
+    if (result['target'].startsWith('centos')) {
+      result['java_pkg'] = 'java-1.8.0-openjdk'
+    } else if (result['target'].startsWith('ubuntu')) {
+      result['java_pkg'] = 'openjdk-8-jdk'
+    } else if (result['target'].startsWith('leap')) {
+      result['java_pkg'] = 'java-1_8_0-openjdk'
+    } else {
+      error 'Java package not known for ' + result['target']
+    }
+
     result['compiler'] = 'gcc'
-    if (env.COMPILER) {
+    if (config['COMPILER']) {
+      result['compiler'] = config['COMPILER']
+    } else if (env.COMPILER) {
       result['compiler'] = env.COMPILER
     } else if (env.STAGE_NAME.contains('Clang')) {
       result['compiler'] = 'clang'
@@ -61,6 +73,16 @@ def call(Map config = [:]) {
       result['compiler'] = 'icc'
     } else if (env.STAGE_NAME.contains('Bullseye')) {
       result['compiler'] = 'covc'
+    }
+
+    if (config['build_type']) {
+      result['build_type'] = config['build_type']
+    } else if (config['BUILD_TYPE']) {
+      result['build_type'] = config['BUILD_TYPE']
+    } else if (env.STAGE_NAME.contains('release')) {
+      result['build_type'] = 'release'
+    } else if (env.STAGE_NAME.contains('debug')) {
+      result['build_type'] = 'debug'
     }
 
     if (config['target_prefix']) {
@@ -71,6 +93,17 @@ def call(Map config = [:]) {
 
     if (env.STAGE_NAME.contains('Coverity')) {
       result['test'] = 'coverity'
+    }
+
+    if (config['log_to_file']) {
+      result['log_to_file'] = config['log_to_file']
+    } else {
+       result['log_to_file'] = result['target'] + '-' +
+                               result['compiler']
+      if (result['build_type']) {
+        result['log_to_file'] += '-' + result['build_type']
+      }
+      result['log_to_file'] += '-build.log'
     }
 
     // Unless otherwise specified, all tests will only use one node.
--- a/vars/parseStageInfo.groovy
+++ b/vars/parseStageInfo.groovy
@@ -113,6 +113,16 @@ def call(Map config = [:]) {
       result['ftest_arg'] = config['ftest_arg']
     }
 
+    result['junit_files'] = 'test_results/*.xml'
+    if (config['junit_files']) {
+      result['junit_files'] = config['junit_files']
+    }
+
+    result['artifacts'] = ['run_test.sh/*', 'vm_test/**']
+    if (config['artifacts']) {
+      result['artifacts'] = config['artifacts']
+    }
+
     result['valgrind'] = 'disabled'
     if (config['valgrind']) {
       result['valgrind'] = config['valgrind']
@@ -122,7 +132,9 @@ def call(Map config = [:]) {
 
     result['valgrind_pattern'] = 'dnt.*.memcheck.xml'
     if (result['valgrind'] == 'memcheck') {
-        result['valgrind_pattern'] = 'run_test_memcheck.sh/*memcheck.xml'
+      result['valgrind_pattern'] = 'run_test_memcheck.sh/*memcheck.xml'
+      result['junit_files'] = ''
+      result['artifacts'] = ['run_test_memcheck.sh/*']
     }
     return result
 }
--- a/vars/parseStageInfo.groovy
+++ b/vars/parseStageInfo.groovy
@@ -16,7 +16,7 @@
  *                         Default is not present unless the word
  *                         TARGET_PREFIX is present in env.STAGE_NAME,
  *                         and then it will default to 'install/opt'
- * 
+ *
  * This is to simplify the addition of stages to the Jenkinsfile so
  * that in many cases, only the Stage Name text needs to be updated.
  *
--- a/vars/parseStageInfo.groovy
+++ b/vars/parseStageInfo.groovy
@@ -116,7 +116,7 @@ def call(Map config = [:]) {
     if (env.STAGE_NAME.contains('run_test.sh')) {
       result['junit_files'] = 'test_results/*.xml'
       result['artifacts'] = ['run_test.sh/*', 'vm_test/**']
-      result['with_valgrind'] = ''
+      result['with_valgrind'] = 'disabled'
       result['valgrind_pattern'] = 'dnt.*.memcheck.xml'
 
       if (config['junit_files']) {
--- a/vars/runTest.groovy
+++ b/vars/runTest.groovy
@@ -101,9 +101,14 @@ def call(Map config = [:]) {
 
     def cb_result = currentBuild.result
     
+    echo config['with_valgrind']
+    echo env.WITH_VALGRIND
     environment {
         WITH_VALGRIND = config['with_valgrind']
     }
+    echo env.WITH_VALGRIND
+    env.WITH_VALGRIND = config['with_valgrind']
+    echo env.WITH_VALGRIND
 
     int rc = 0
     rc = sh(script: script, label: flow_name, returnStatus: true)
--- a/vars/runTest.groovy
+++ b/vars/runTest.groovy
@@ -99,6 +99,10 @@ def call(Map config = [:]) {
                           config['failure_artifacts'] + '"'
     }
 
+    println config['with_valgrind']
+    environment {
+        WITH_VALGRIND = config['with_valgrind']
+    }
     if(config['with_valgrind']) {
         env.WITH_VALGRIND = config['with_valgrind']
     }
--- a/vars/runTest.groovy
+++ b/vars/runTest.groovy
@@ -99,18 +99,24 @@ def call(Map config = [:]) {
                           config['failure_artifacts'] + '"'
     }
 
-    println config['with_valgrind']
+    def cb_result = currentBuild.result
+    
+    echo config['with_valgrind']
     environment {
         WITH_VALGRIND = config['with_valgrind']
     }
+    echo env.WITH_VALGRIND
     if(config['with_valgrind']) {
         env.WITH_VALGRIND = config['with_valgrind']
     }
+    echo env.WITH_VALGRIND
 
-    def cb_result = currentBuild.result
-
-    int rc = 0
-    rc = sh(script: script, label: flow_name, returnStatus: true)
+    withEnv(['WITH_VALGRIND=${config[with_valgrind]}']) {
+      echo env.WITH_VALGRIND
+      int rc = 0
+      rc = sh(script: script, label: flow_name, returnStatus: true)
+    }
+    echo env.WITH_VALGRIND
 
     if (cb_result != currentBuild.result) {
       println "The runTest script changed build result to " +
--- a/vars/runTest.groovy
+++ b/vars/runTest.groovy
@@ -101,22 +101,12 @@ def call(Map config = [:]) {
 
     def cb_result = currentBuild.result
     
-    echo config['with_valgrind']
     environment {
         WITH_VALGRIND = config['with_valgrind']
     }
-    echo env.WITH_VALGRIND
-    if(config['with_valgrind']) {
-        env.WITH_VALGRIND = config['with_valgrind']
-    }
-    echo env.WITH_VALGRIND
 
-    withEnv(['WITH_VALGRIND=${config[with_valgrind]}']) {
-      echo env.WITH_VALGRIND
-      int rc = 0
-      rc = sh(script: script, label: flow_name, returnStatus: true)
-    }
-    echo env.WITH_VALGRIND
+    int rc = 0
+    rc = sh(script: script, label: flow_name, returnStatus: true)
 
     if (cb_result != currentBuild.result) {
       println "The runTest script changed build result to " +
--- a/vars/runTest.groovy
+++ b/vars/runTest.groovy
@@ -99,11 +99,9 @@ def call(Map config = [:]) {
                           config['failure_artifacts'] + '"'
     }
 
-    echo "marj> in runTest env.WITH_VALGRIND = ${env.WITH_VALGRIND}"
     if(config['with_valgrind']) {
         env.WITH_VALGRIND = config['with_valgrind']
     }
-    echo "marj> in runTest env.WITH_VALGRIND = ${env.WITH_VALGRIND}"
 
     int rc = 0
     rc = sh(script: script, label: flow_name, returnStatus: true)
--- a/vars/runTest.groovy
+++ b/vars/runTest.groovy
@@ -101,8 +101,6 @@ def call(Map config = [:]) {
 
     def cb_result = currentBuild.result
     
-    env.WITH_VALGRIND = config['with_valgrind']
-
     int rc = 0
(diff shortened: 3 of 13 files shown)


vars/runTestSh.groovy
```diff
@@ -0,0 +1,55 @@
...
+    timeout(time:60, unit:'MINUTES') {
+      runTest stashes: [ 'centos7-gcc-tests',
+                       'centos7-gcc-install',
+                       'centos7-gcc-build-vars' ],
+            script: "SSH_KEY_ARGS=${env.SSH_KEY_ARGS} " +
+                    "NODELIST=${env.NODELIST} " +
+                    'ci/unit/test_main.sh',
+            junit_files: 'test_results/*.xml'
+    }
+  }
```
brianjmurrell, line 54: To DRY this out a bit, how about:

```suggestion
  String mode = ""
  String junit_files = 'test_results/*.xml'
  if (config['mode'] == 'memcheck') {
    mode 'memcheck'
    junit_files = ''
  }
  timeout(time:60, unit:'MINUTES') {
    runTest stashes: [ 'centos7-gcc-tests',
                     'centos7-gcc-install',
                     'centos7-gcc-build-vars' ],
          script: "SSH_KEY_ARGS=${env.SSH_KEY_ARGS} " +
                  "NODELIST=${env.NODELIST} " +
                  'ci/unit/test_main.sh ' + mode,
          junit_files: junit_files
  }
```

vars/runTestSh.groovy
```diff
@@ -0,0 +1,55 @@
+// vars/runTestSh.groovy
```
brianjmurrell, line 1: This name is not really very descriptive of what this function is.

How about `runTestUnitTest` instead?

JohnMalmberg:
This looks like it will need to be reworked after PR 146 lands.

vars/runTestSh.groovy
marcelarosalesj: Hi @brianjmurrell, thanks for your suggestion, I'll update this for my next patch

JohnMalmberg:
For a Matrix build, we are going to want to have 1 groovy method that handles all unit tests.

Each test stage in the Doas Jenkinsfile needs to have a unique stage name that indicates what TARGET, Compiler and any thing else that differentiates that stage from the other similar stages.

The ParseStageInfo.groovy is used to parse either that unique stage name or the environment variables that would be supplied by a Matrix to set the defaults for parameters that can also be passed to the 1 groovy method.

A Matrix should be the most efficient way of coding a Jenkinsfile for this.  Currently due to a bug in the Jenkins Matrix implementations, Matrixes only report correctly on the WEB-UI when all axes of the Matrix get run all the time.

The desired end goal is to end up with identical unit test stages, except for the unique stage names.

vars/unitTest.groovy
```diff
@@ -84,12 +87,14 @@ def call(Map config = [:]) {
     stashes.add("${target_compiler}-install")
     stashes.add("${target_compiler}-build-vars")
   }
+  
+  def valgrind = config.get('valgrind', '')
```
JohnMalmberg, line 91: We want the major defaults to come from the `stage_info` map, so that they can be set either by setting the contents fo the stage name or by environment variables in the case of a matrix job.

To be able to move to Jenkins matrix notation all steps inside the stage need to have identical parameters and have defaults set by either parsing the stage name environment variable or environment variables set for by a Jenkins matrix.

vars/unitTest.groovy
```diff
@@ -84,12 +87,14 @@ def call(Map config = [:]) {
...
 
   Map params = [:]
   params['stashes'] = stashes
   params['script'] = "SSH_KEY_ARGS=${env.SSH_KEY_ARGS} " +
                      "NODELIST=${nodelist} " +
-                     test_script
+                     test_script + " " + valgrind
```
JohnMalmberg, line 97: Pretty sure that is not going to work.  This is appending to the end of a passed script file, not a command.

There may or may not be a newline at the end of the test_script.  No way to predict what a future change might do even if it works now.

vars/unitTestPost.groovy
```diff
@@ -25,7 +31,8 @@ def call(Map config = [:]) {
 
   def always_script = config.get('always_script',
                                  'ci/unit/test_post_always.sh')
-  sh script: always_script,
+  def valgrind = config.get('valgrind', '')
```
JohnMalmberg, line 34: Same as above, the defaults need to come from the Map stage_info that is populated a few lines down.

vars/unitTestPost.groovy
```diff
@@ -45,45 +52,51 @@ def call(Map config = [:]) {
...
-
-               qualityGates: [
-                 [threshold: 1, type: 'TOTAL_HIGH', unstable: true],
-                 [threshold: 1, type: 'TOTAL_ERROR', unstable: true],
-                 [threshold: 1, type: 'NEW_NORMAL', unstable: true]],
-                name: "Node local testing",
-                tool: issues(pattern: 'vm_test/nlt-errors.json',
-                             name: 'NLT results',
-                             id: 'VM_test')
+  if (valgrind == '') {
```
JohnMalmberg, line 78: if (stage_info['valgrind']) {

This tests false when the 'valgrind' member does not exist, or is set to a value that evaluates as false such as an empty string.

vars/unitTestPost.groovy
```diff
@@ -45,45 +52,51 @@ def call(Map config = [:]) {
...
+    junit testResults: test_results
+  }
 
   def artifact_list = config.get('artifacts', ['run_test.sh/*', 'vm_test/**'])
   artifact_list.each {
     archiveArtifacts artifacts: it
   }
-
+  
+  def valgrind_pattern = config.get('valgrind_pattern', 'dnt.*.memcheck.xml')
```
JohnMalmberg, line 65: We are doing already doing some valgrind tests?

marcelarosalesj: yes, it's enabled for run_test.sh, I'm working on your comments, thanks

jolivier23:
Looks like this has a conflict now with master

jolivier23:
Looks like this has a conflict now with master

vars/unitTest.groovy
```diff
@@ -84,6 +87,10 @@ def call(Map config = [:]) {
     stashes.add("${target_compiler}-install")
     stashes.add("${target_compiler}-build-vars")
   }
+  
+  if (stage_info['valgrind']) {
+    test_script = test_script.trim() + " " + stage_info['valgrind']
```
brianjmurrell, line 92: You cannot/should not do this.  `test_script` holds a path to the test script.  It should not be depended on that these scripts take arguments.  Environment variables is the way we pass in values to these scripts.

vars/unitTestPost.groovy
```diff
@@ -25,10 +29,13 @@ def call(Map config = [:]) {
 
   def always_script = config.get('always_script',
                                  'ci/unit/test_post_always.sh')
+  Map stage_info = parseStageInfo(config)
+  if (stage_info['valgrind']) {
+    always_script = always_script.trim() + " " + stage_info['valgrind']
```
brianjmurrell, line 34: Ditto.

vars/unitTestPost.groovy
```diff
@@ -45,45 +52,50 @@ def call(Map config = [:]) {
...
+  if (test_results != "") {
+    junit testResults: test_results
+  }
 
   def artifact_list = config.get('artifacts', ['run_test.sh/*', 'vm_test/**'])
   artifact_list.each {
     archiveArtifacts artifacts: it
   }
-
+
```
brianjmurrell, line 64: Whitespace.
```suggestion

```

vars/unitTestPost.groovy
```diff
@@ -45,45 +52,50 @@ def call(Map config = [:]) {
     return
   }
 
-  junit testResults: config.get('testResults', 'test_results/*.xml')
+  def test_results = config.get('testResults', 'test_results/*.xml')
+  if (test_results != "") {
```
brianjmurrell, line 56: When would this be empty?  If it never should, we shouldn't paper over it being empty by not running the `junit` step and preventing it from register the missing results.

vars/unitTest.groovy
marcelarosalesj: okay, I'm working on this

vars/unitTestPost.groovy
marcelarosalesj: you're right, I'll fix it

vars/unitTest.groovy
marcelarosalesj: Hi @brianjmurrell , I've been experimenting with this and I realized that environment variables are shared among stages. For example, in here https://build.hpdd.intel.com/blue/organizations/jenkins/daos-stack%2Fdaos/detail/PR-3023/48/pipeline/442 the complete log of `run_test.sh with memcheck` and `run_test.sh` have the same value for `WITH_VALGRIND`, I think it's a race condition of the last stage that set the variable, and then both stages end up executing the same branch of the `run_test.sh` script

I'm going to try using this, though I don't know if that will work https://wiki.jenkins.io/pages/viewpage.action?pageId=70878639 let me know if you have any suggestions on how to solve this problem

vars/unitTest.groovy
```diff
@@ -87,9 +90,21 @@ def call(Map config = [:]) {
 
   Map params = [:]
   params['stashes'] = stashes
-  params['script'] = "SSH_KEY_ARGS=${env.SSH_KEY_ARGS} " +
-                     "NODELIST=${nodelist} " +
-                     test_script
+  
+
+  echo "marj> env['WITH_VALGRIND'] = ${env['WITH_VALGRIND']}"
+  lock(label: 'WITH_VALGRIND resource', variable: 'WITH_VALGRIND') {
```
brianjmurrell, line 96: This is quite probably not the right solution.  I'm actually not entirely sure what you are trying to do with environment variables here.  Can you try to explain what the goal of using this environment variable here in this groovy step is?

vars/unitTest.groovy
brianjmurrell: > I realized that environment variables are shared among stages

Well, to some degree, yes.  Depending on how you are setting/using them.  Is the problem the evaluation of env.WITH_VALGRIND here:https://github.com/daos-stack/pipeline-lib/blob/1ffb977116ef4fed07f3ce746e42c78d409b9f29/vars/unitTest.groovy#L95-L105?  If so, why do you want to use an environment variable there?  Why not just use the `config` `Map` value?

vars/unitTest.groovy
```diff
@@ -87,9 +90,21 @@ def call(Map config = [:]) {
...
+    env['WITH_VALGRIND'] = stage_info['valgrind']
+    echo "marj> env['WITH_VALGRIND'] = ${env['WITH_VALGRIND']}"
+    params['script'] = "SSH_KEY_ARGS=${env.SSH_KEY_ARGS} " +
+                       "NODELIST=${nodelist} " +
+                       test_script
+
+  }
+  echo "marj> env['WITH_VALGRIND'] = ${env['WITH_VALGRIND']}"
+  
+
```
brianjmurrell, line 107: I think perhaps what you are trying to achieve here is:

```suggestion
    params['script'] = "SSH_KEY_ARGS=${env.SSH_KEY_ARGS} " +
                       "NODELIST=${nodelist} " +
                       "WITH_VALGRIND=" + config.get('valgrind',
                                                     stage_info['valgrind']) + '  ' +
                       test_script
```

vars/unitTest.groovy
```diff
@@ -87,13 +87,10 @@ def call(Map config = [:]) {
...
-  if (stage_info['valgrind']) {
-    test_script = test_script.trim() + " " + stage_info['valgrind']
-  }
 
   Map params = [:]
   params['stashes'] = stashes
+  params['with_valgrind'] = stage_info['valgrind']
   params['script'] = "SSH_KEY_ARGS=${env.SSH_KEY_ARGS} " +
                      "NODELIST=${nodelist} " +
                      test_script
```
jolivier23, line 96: what is test_script here?   Could you just do "WITH_VALGRIND=" + stage_info['valgrind'] " + test_script ?

vars/unitTestPost.groovy
marcelarosalesj: right, no, this should execute when valgrind is 'disabled', I'll fix it next patch

vars/unitTest.groovy
marcelarosalesj: right, I didn't need to set up that variable here. I moved it to runTest and it is not running the same stage twice anymore. I quite don't understand why the problem is solved, but the WITH_VALGRIND variable is not shared in that part of the groovy code

vars/unitTest.groovy
marcelarosalesj: We cannot make sure how test_script will change in the future, that's why @brianjmurrell suggested the environment variable. I set up the variable in `runTest.groovy` instead of `unitTest.groovy` and it works now.

vars/parseStageInfo.groovy
```diff
@@ -112,5 +112,29 @@ def call(Map config = [:]) {
...
+
+    result['junit_files'] = 'test_results/*.xml'
+    if (config['junit_files']) {
+      result['junit_files'] = config['junit_files']
+    }
+
+    result['artifacts'] = ['run_test.sh/*', 'vm_test/**']
+    if (config['artifacts']) {
+      result['artifacts'] = config['artifacts']
+    }
```
JohnMalmberg, line 124: Settings here are global to all stages, not just unit test stages.

This should be checking to see if this is a unit test stage from the stage name before setting a default that is specific for Unit tests.

vars/parseStageInfo.groovy
```diff
@@ -112,5 +112,29 @@ def call(Map config = [:]) {
...
+
+    result['valgrind'] = 'disabled'
```
JohnMalmberg, line 126: Would not be disabled just be the absence of a 'valgrind' key?

vars/runTest.groovy
```diff
@@ -99,6 +99,10 @@ def call(Map config = [:]) {
                           config['failure_artifacts'] + '"'
     }
 
+    if(config['with_valgrind']) {
```
JohnMalmberg, line 102: Is the key 'valgrind' or 'with_valgrind'?

vars/unitTest.groovy
JohnMalmberg: The stage name in Jenkins in Jenkins for non-matrix jobs should be unique and have the keywords in it to set the defaults used in the block of the stage.

This allows cut/paste and no edits except for the stage name to setup the actions for the stage.

When matrix jobs start working, the matrix will set environment variables directly instead of parsing the stage name, which will be more generic.

vars/parseStageInfo.groovy
marcelarosalesj: right, I'll add this on my next patch

vars/parseStageInfo.groovy
marcelarosalesj: I'm using the 'disabled' keyword on run_test.sh here https://github.com/daos-stack/daos/pull/3023

vars/runTest.groovy
marcelarosalesj: I'm using 'valgrind' for parseStageInfo and as the parameter for unitTest.groovy and 'with_valgrind' for setting up the environment variable, but I could stick to 'with_valgrind' for consistency

marcelarosalesj:
@JohnMalmberg I sent some changes with your feedback.
The stage running the unit testing with valgrind is called "run_test.sh with memcheck" and uses the same set up as "run_test.sh". In fact, with changes here in #151 I made the unitTest and unitTestPost calls to be the same for both stages (https://github.com/daos-stack/daos/pull/3023/). Will that fit okay for the Jenkins non-matrix?
And about the Jenkins matrix, I don't understand very well what is this about. Do I have to set up some other variables in this patch? Or is this some other activity?

JohnMalmberg:
See [Daos PR-3104](https://github.com/daos-stack/daos/pull/3104)
The "run_test.sh" name is eventually going away.  It can not leave immediately because github PRs are checking for that name and they also need to be changed at that time.

The parallel stages inside the "Unit Test" global stage should have "Unit Test" in their names.  I need to fix that in PR-3104 and then get a full test run for it.

You do not need to anything special to your code for matrix builds.

For matrix builds, what happens is that instead of parallel block with several internal stage blocks inside of it, there is a matrix block with only one stage in side of it.   There is still no action on the bug in Jenkins that blocks using sparse matrix.   A sparse matrix is one that not all the combinations are always run.   There are also some issues with the display of the matrix stages on the Web UI that it would be nice if we could get Jenkins to fix.

If we do things "right", ideally then only parseStageInfo needs to be changed to understand to use the environment variables that are set by the matrix for the conversion instead of parsing the stage name.

So ideally. for now we will have a set of parallel stages that are identical except for the stage names.

The first prototype of testing the matrix was that for all combinations of "TARGET" which could be centos7, centos8, leap15, ubuntu20, and "COMPILER" which could be gcc, clang, icc, covc.   That is why that parseStageInfo is doing those checks in that order.

We may be adding code that has Jenkins change the default of TARGET of a stage from "centos7" to some thing else like "leap15" based on either a branch name of a commit pragma.   You should not be concerned about that for this PR.

I think that after [Daos PR-3104](https://github.com/daos-stack/daos/pull/3104) lands, there will need to be code to make a code coverage test run and any type of valgrind run to be mutually exclusive.

vars/parseStageInfo.groovy
```diff
@@ -112,5 +112,32 @@ def call(Map config = [:]) {
     if (config['ftest_arg']) {
       result['ftest_arg'] = config['ftest_arg']
     }
+
+    if (env.STAGE_NAME.contains('run_test.sh')) {
+      result['junit_files'] = 'test_results/*.xml'
+      result['artifacts'] = ['run_test.sh/*', 'vm_test/**']
+      result['with_valgrind'] = 'disabled'
```
brianjmurrell, line 119: It seems that we can just leave this key unset if valgrind is not desired and then not enable valgrind if the key is unset.

vars/parseStageInfo.groovy
```diff
@@ -112,5 +112,32 @@ def call(Map config = [:]) {
...
+      if (config['with_valgrind']) {
+        result['with_valgrind'] = config['with_valgrind']
+      } else if(env.STAGE_NAME.contains('memcheck')) {
+        result['with_valgrind'] = 'memcheck'
+      }
+
+      if (result['with_valgrind'] == 'memcheck') {
+        result['valgrind_pattern'] = 'run_test_memcheck.sh/*memcheck.xml'
+        result['junit_files'] = ''
+        result['artifacts'] = ['run_test_memcheck.sh/*']
```
brianjmurrell, line 138: Why is the memcheck version of the UT stage including this file as an artifact but the non-memcheck one is not?

vars/parseStageInfo.groovy
```diff
@@ -112,5 +112,32 @@ def call(Map config = [:]) {
...
+
+      if (config['with_valgrind']) {
+        result['with_valgrind'] = config['with_valgrind']
+      } else if(env.STAGE_NAME.contains('memcheck')) {
+        result['with_valgrind'] = 'memcheck'
+      }
+
+      if (result['with_valgrind'] == 'memcheck') {
+        result['valgrind_pattern'] = 'run_test_memcheck.sh/*memcheck.xml'
+        result['junit_files'] = ''
```
brianjmurrell, line 137: Why are there no junit files for the memcheck version of the UT stage?

vars/unitTestPost.groovy
```diff
@@ -44,46 +50,53 @@ def call(Map config = [:]) {
                      allowEmptyArchive: true
     return
   }
+  
+  if (stage_info['with_valgrind'] == 'disabled') {
+    def test_results = config.get('testResults', 'test_results/*.xml')
+    junit testResults: test_results
+  }
```
brianjmurrell, line 57: Yeah, so again, why only JUnit results for non-memcheck?  That seems to be adding some unnecessary gating here.

vars/unitTestPost.groovy
```diff
@@ -44,46 +50,53 @@ def call(Map config = [:]) {
...
+                    failThresholdTotal: '0',
+                    pattern: stage_info['valgrind_pattern'],
+                    publishResultsForAbortedBuilds: false,
+                    publishResultsForFailedBuilds: true,
+                    sourceSubstitutionPaths: '',
+                    unstableThresholdDefinitelyLost: '0',
+                    unstableThresholdInvalidReadWrite: '0',
+                    unstableThresholdTotal: '0'
+  }
+  if (stage_info['with_valgrind'] == 'disabled') {
```
brianjmurrell, line 78: This is an `else` isn't it?
```suggestion
  } else {
```

vars/parseStageInfo.groovy
marcelarosalesj: Okay, I can do that

vars/parseStageInfo.groovy
marcelarosalesj: UT includes `run_test.sh` as an artifact, while UT with memcheck includes `run_test_memcheck.sh`. I decided a different name to identify the different logs.

vars/parseStageInfo.groovy
marcelarosalesj: I'm considering only valgrind results for UT with memcheck, and those supposedly should be publish by valgrind plugin. UT test results are reported in UT stage, through junit. Though, I don't know if using junit for memcheck xml could be a workaround for publishValgrind bug.

vars/unitTestPost.groovy
marcelarosalesj: I could use an else if. There's a task about enabling pmemcheck some time in the future, so I don't want to assume that anything else should do this code.

vars/parseStageInfo.groovy
marcelarosalesj: This variable ends up as an environment variable in `daos/ci/unit/test_post_always.sh`, and if this is not set to something the CI gives an unbound variable error. I like that test_post_always.sh doesn't change much in this pr https://github.com/daos-stack/daos/pull/3023 so probably I'll put back the `disabled` keyword.
I could remove the `set -u` in test_post_always.sh, but I don't think that's correct. I don't know about other options, let me know your comments.

vars/parseStageInfo.groovy
```diff
@@ -112,5 +112,31 @@ def call(Map config = [:]) {
     if (config['ftest_arg']) {
       result['ftest_arg'] = config['ftest_arg']
     }
+
+    if (env.STAGE_NAME.contains('run_test.sh')) {
+      result['junit_files'] = 'test_results/*.xml'
+      result['artifacts'] = ['run_test.sh/*', 'vm_test/**']
+      result['with_valgrind'] = ''
```
JohnMalmberg, line 119: Key should just not be present.

vars/parseStageInfo.groovy
```diff
@@ -112,5 +112,31 @@ def call(Map config = [:]) {
     if (config['ftest_arg']) {
       result['ftest_arg'] = config['ftest_arg']
     }
+
+    if (env.STAGE_NAME.contains('run_test.sh')) {
```
JohnMalmberg, line 116: run_test.sh name is going to go away, possibly soon.

Since there is no information to parse from the 'run_test.sh" as to what the stage is, the defaults for it should be the defaults for the "unitTest.groovy" and "unitTestPost.groovy" and not specified here.

vars/unitTestPost.groovy
```diff
@@ -44,46 +50,50 @@ def call(Map config = [:]) {
...
     return
   }
+  
+  def test_results = stage_info['junit_files']
+  junit testResults: test_results
 
-  junit testResults: config.get('testResults', 'test_results/*.xml')
-
-  def artifact_list = config.get('artifacts', ['run_test.sh/*', 'vm_test/**'])
+  def artifact_list = stage_info['artifacts']
```
JohnMalmberg, line 57: This needs to be reverted per the above comments in parseStageInfo.

parseStageInfo only provides defaults based on parsing the stage name.

For our current stage names that do not have information to be expected to be those settings become the default.

marcelarosalesj: I changed this, now I'm sending a value from daos Jenkinsfile and using config.get

vars/parseStageInfo.groovy
marcelarosalesj: I fixed this, now I'm modifing the appropriate defaults from daos Jenkinsfile. In this parseStageInfo section I still have the set up of with_valgrind variable, though.

vars/parseStageInfo.groovy
marcelarosalesj: how should this be managed? I thought we wanted to know this information (if run with valgrind or not) from the stage name

vars/runTest.groovy
```diff
@@ -99,6 +99,10 @@ def call(Map config = [:]) {
...
+        env.WITH_VALGRIND = config['with_valgrind']
```
marcelarosalesj, line 103: @JohnMalmberg @brianjmurrell on my latest review and jenkins job I'm getting the same old error of having both `run_test.sh` and `Unit Test with memcheck` stages run with memcheck 

This is run_test.sh complete log https://build.hpdd.intel.com/blue/rest/organizations/jenkins/pipelines/daos-stack/pipelines/daos/branches/PR-3023/runs/67/nodes/407/steps/487/log/?start=0
and this is UT with memcheck https://build.hpdd.intel.com/blue/rest/organizations/jenkins/pipelines/daos-stack/pipelines/daos/branches/PR-3023/runs/67/nodes/408/steps/486/log/?start=0

This was solved before by setting up the environment variable here in `runTest.groovy`, but it seems to not working anymore, do you know how to fix this?

vars/parseStageInfo.groovy
JohnMalmberg: If the 'with_valgrind' key is not present then it means that valgrind tests are not wanted, so the default is no key.

vars/runTest.groovy
JohnMalmberg: I don't think you can persistently set an environment variable this way, and if you can, it will probably introduce a bug.

It is a matter of variable scope, which is one of our common problems in trying to debug flaky behavior in groovy scripts and Jenkinsfiles.

In a groovy library if you do not declare a variable as "def" or a type, that variable becomes global to the entire Jenkins job.  Which means that if you have two unit test stages running in parallel and are missing that declaration somewhere, sometimes it will work, and sometimes it will not.

I am not sure what scope the "env" variable is in.

Usually I debug groovy scripts by adding println statements in them to dump variables and let me know why a conditional path is taken.

vars/runTest.groovy
```diff
@@ -100,6 +100,8 @@ def call(Map config = [:]) {
     }
 
     def cb_result = currentBuild.result
+    
+    env.WITH_VALGRIND = config['with_valgrind']
```
JohnMalmberg, line 104: I don't think we should be setting environment variables this way.

Why is this being changed global to the stage?

See below where the NODELIST environment variable is being set to the script to be run.

To pass this to a script it should be passed as a parameter, possibly something like:

"WITH_VALGRIND=${stage_info.get('with_valgrind',\"\")}" +

marcelarosalesj: thanks, I sent this fix and it is working

Jenkinsfile
```diff
@@ -16,6 +16,7 @@
 // Then a second PR submitted to comment out the @Library line, and when it
 // is landed, both PR branches can be deleted.
 //@Library(value="pipeline-lib@my_branch_name") _
+@Library(value="pipeline-lib@marosale/CORCI-951-refactor") _
```
JohnMalmberg, line 19: I don't think you changed anything that would requires this line.

marcelarosalesj: Thanks for your feedback. I removed this line

jolivier23 (approved):
@JohnMalmberg @brianjmurrell where are we on this?

marcelarosalesj:
I think this is complete now. I was seeing some errors related to 'dirty' workspaces, basically when executing unitTest.groovy  (and test_main.sh) there where Unit Test files and directories from previous builds, The same happened for unitTestPost.groovy, before unstashing there were already some files from before. I thought it was due to a dirty stash, but it was the workspace that was not clean. I added a cleanup removes in daos/ci/unit/test_main.sh and here in unitTestPost.groovy because it seems that every call to groovy its own clean up.

jolivier23 (approved):
@JohnMalmberg @brianjmurrell can you guys review this again, please?

JohnMalmberg (dismissed):

vars/unitTestPost.groovy
```diff
@@ -50,6 +50,16 @@ def call(Map config = [:]) {
...
   }
 
+  if(stage_info['with_valgrind']) {
+    String target_dir = "unit_test_memcheck_logs"
+    String src_files = "unit-test-*.memcheck.xml"
+    fileOperations([fileCopyOperation(excludes: '',
+                                      flattenFiles: false,
+                                      includes: src_files,
+                                      targetLocation: target_dir)])
+    sh """tar -czf ${target_dir}.tar.gz ${target_dir}"""
```
brianjmurrell, line 60: No need for a triple-quote here.

```suggestion
    sh "tar -czf ${target_dir}.tar.gz ${target_dir}"
```

vars/parseStageInfo.groovy
```diff
@@ -145,5 +145,12 @@ def call(Map config = [:]) {
     if (config['ftest_arg']) {
       result['ftest_arg'] = config['ftest_arg']
     }
+
+    if (env.STAGE_NAME.contains('Unit Test')) {
+      if(env.STAGE_NAME.contains('memcheck')) {
+        result['with_valgrind'] = 'memcheck'
+      }
+    }
```
brianjmurrell, line 153: Can be a single if statement with multiple conditions:

```suggestion
    if (env.STAGE_NAME.contains('Unit Test') &&
        env.STAGE_NAME.contains('memcheck')) {
        result['with_valgrind'] = 'memcheck'
    }
```

vars/valgrindReportPublish.groovy
```diff
@@ -30,21 +30,25 @@ def call(Map config = [:]) {
     // Need to have only one Valgrind publish stage
     println "No valgrind_stashes passed!   Running older code!"
   }
+  
+  sh label: 'Remove *.memcheck.xml files in this workspace',
+     script: '''rm -rf *.memcheck.xml'''
```
brianjmurrell, line 35: You don't need triple quotes here.

```suggestion
     script: 'rm -rf *.memcheck.xml'
```
But why `-r` for what looks like it should be a file pattern?  And why not use `fileOperations->fileDeleteOperation` instead of a `sh` step here?

brianjmurrell (changes requested):

vars/valgrindReportPublish.groovy
marcelarosalesj: I use the -r because of the dirty workspaces problem it may be the case these files do not exist (it they're not here from previous builds). In fact in my experiments sometimes I saw memcheck result files with very old names instead of the *memcheck.xml pattern here. I don't know how fileOperations will behave in the case of a workspace with no files that match this pattern

marcelarosalesj: I'm going to send a patch deleting a file that do not exist to see how this behaves, if it doesn't throw an error, I think we can go safe with fileOperations

jolivier23 (approved):

vars/valgrindReportPublish.groovy
jolivier23: -f handles the non-existent case.  -r is for recursion which we probably don't want.

marcelarosalesj: Oh, right, my mistake. 

I've addressed Brian's comments now. I removed the triple quote and use fileDeleteOperation instead of rm.

brianjmurrell: @marcelarosalesj You should leave conversations to be resolved by the person who started them, so that they can see what their previous issues were and whether they were resolved or not.

vars/valgrindReportPublish.groovy
```diff
@@ -30,17 +30,17 @@ def call(Map config = [:]) {
     // Need to have only one Valgrind publish stage
     println "No valgrind_stashes passed!   Running older code!"
   }
-  
-  sh label: 'Remove *.memcheck.xml files in this workspace',
-     script: '''rm -rf *.memcheck.xml'''
+
+  fileOperations([fileDeleteOperation(includes: '*.memcheck.xml',
+                                      excludes: '')])
```
brianjmurrell, line 35: Is the empty excludes necessary?  Seems like that would be the default value.

Your experiments with trying to delete a pattern revealed that it didn't cause any errors or exceptions?

marcelarosalesj: You're right, I think it's not necessary. I sent a patch removing it. 

And yes, the delete with fileOperations did not cause any problems.

jolivier23 (approved):

vars/valgrindReportPublish.groovy
brianjmurrell: @marcelarosalesj Please don't resolve conversations you didn't start.  It's up to the person who made the initial comment to resolve the conversation when they are satisfied that their issue was resolved.

brianjmurrell (approved):

PR #154, opened 2020-07-15
Only supply repos and rpms arg if present

brianjmurrell:


brianjmurrell:
Landed in another PR.

PR #155, opened 2020-07-16
CORCI-918 build: Adding coverage builds

JohnMalmberg:
Add defaults that are automatically set for debug, release,
and coverage builds.

### DIFF
--- a/vars/cloverReportPublish.groovy
+++ b/vars/cloverReportPublish.groovy
@@ -24,15 +24,6 @@
    *
    * config['ignore_failure']      Ignore test failures.  Default false.
    *
-   * config['inst_rpms']           Additional rpms to install.  Optional
-   *
-   * config['NODELIST']            NODELIST of nodes to run tests on.
-   *                               Default env.NODELIST
-   *
-   * config['node_count']          Count of nodes that will actually be used
-   *                               the test.  Default will be based on the
-   *                               enviroment variables for the stage.
-   *
    * config['stash']               Stash name for the ".build-vars.*" files.
    */
 
@@ -45,28 +36,12 @@ def call(Map config = [:]) {
   }
   Map stage_info = parseStageInfo(config)
 
-  String inst_rpms = config.get('inst_rpms','')
-
-  if (stage_info['java_pkg']) {
-    inst_rpms += " ${stage_info['java_pkg']}"
-  }
-
-  provisionNodes NODELIST: nodelist,
-                 node_count: stage_info['node_count'],
-                 distro: stage_info['target'],
-                 inst_repos: config.get('inst_repos',''),
-                 inst_rpms: inst_rpms
-
   String coverage_website = config.get('coverage_website',
                                        'coverage_website.zip')
 
   String url_base = env.JENKINS_URL +
                     'job/daos-stack/job/tools/job/master' +
                     '/lastSuccessfulBuild/artifact/'
-
-  httpRequest url:  url_base + 'bullseyecoverage-linux.tar',
-              httpMode: 'GET',
-              outputFile: 'bullseye.tar'
 
   httpRequest url: url_base + 'bullshtml.jar',
               httpMode: 'GET',
@@ -91,11 +66,10 @@ def call(Map config = [:]) {
     unstash it
     stash_cnt++
     String new_name = "test.cov_${stash_cnt}"
-    // Need plugin Upgrade
-    //fileOperations([fileRenameOperation(source: 'test.cov',
-    //                                    destination: new_name)])
-    sh label: 'Rename file',
-       script: "mv test.cov ${new_name}"
+    fileOperations([fileRenameOperation(source: 'test.cov',
+                                        destination: new_name)])
+    //sh label: 'Rename file',
+    //   script: "mv test.cov ${new_name}"
   }
 
   sh label: 'Create Coverage Report',
--- a/vars/cloverReportPublish.groovy
+++ b/vars/cloverReportPublish.groovy
@@ -66,10 +66,11 @@ def call(Map config = [:]) {
     unstash it
     stash_cnt++
     String new_name = "test.cov_${stash_cnt}"
-    fileOperations([fileRenameOperation(source: 'test.cov',
-                                        destination: new_name)])
-    //sh label: 'Rename file',
-    //   script: "mv test.cov ${new_name}"
+    // Need a new plugin installed
+    //fileOperations([fileRenameOperation(source: 'test.cov',
+    //                                    destination: new_name)])
+    sh label: 'Rename file',
+       script: "mv test.cov ${new_name}"
   }
 
   sh label: 'Create Coverage Report',
--- a/vars/cloverReportPublish.groovy
+++ b/vars/cloverReportPublish.groovy
@@ -38,6 +38,11 @@
 
 def call(Map config = [:]) {
 
+  // If we don't have a BULLSEYE environment variable set
+  // there are no Bullsye reports to process.
+  if (!env.BULLSEYE) {
+    return
+  }
   Map stage_info = parseStageInfo(config)
 
   String inst_rpms = config.get('inst_rpms','')
--- a/vars/cloverReportPublish.groovy
+++ b/vars/cloverReportPublish.groovy
@@ -1,0 +1,128 @@
+// vars/cloverReportPublish.groovy
+
+  /**
+   * clover Report Publish step method
+   *
+   * Consolidiate clover data for publising.
+   *
+   * @param config Map of parameters passed
+   *
+   *
+   * config['coverage_healthy']    Map of a coverage healthy target.
+   *                               Default is:
+   *                                 [method:70, conditional:80, statement:80]
+   *
+   * config['coverage_script']     Script to run to get coverage results
+   *                               Default is "ci/bullseye_generate_report.sh"
+   *
+   * config['coverage_stashes']    list of stashes for coverage reports.
+   *                               Required.  Each stash must contain one
+   *                               file named test.cov.
+   *
+   * Config['coverage_website']    Zip file to contain the resulting
+   *                               coverage website, if it is a coverage
+   *                               build.  Defaults to 'coverage_website.zip'
+   *
+   * config['ignore_failure']      Ignore test failures.  Default false.
+   *
+   *
+   * config['inst_rpms']           Additional rpms to install.  Optional
+   *
+   * config['NODELIST']            NODELIST of nodes to run tests on.
+   *                               Default env.NODELIST
+   *
+   * config['node_count']          Count of nodes that will actually be used
+   *                               the test.  Default will be based on the
+   *                               enviroment variables for the stage.
+   *
+   * config['stash']               Stash name for the ".build-vars.*" files.
+   */
+
+def call(Map config = [:]) {
+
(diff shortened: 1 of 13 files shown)


vars/cloverReportPublish.groovy
```diff
@@ -0,0 +1,128 @@
...
+   *                               Required.  Each stash must contain one
+   *                               file named test.cov.
+   *
+   * Config['coverage_website']    Zip file to contain the resulting
+   *                               coverage website, if it is a coverage
+   *                               build.  Defaults to 'coverage_website.zip'
+   *
+   * config['ignore_failure']      Ignore test failures.  Default false.
+   *
+   *
```
brianjmurrell, line 28:
```suggestion
```

vars/cloverReportPublish.groovy
```diff
@@ -0,0 +1,128 @@
...
+  Map stage_info = parseStageInfo(config)
+
+  String inst_rpms = config.get('inst_rpms','')
+
+  if (stage_info['target'].startsWith("centos")) {
+    inst_rpms += ' java-1.8.0-openjdk'
+  } else if (stage_info['target'].startsWith('ubuntu')) {
+    inst_rpms += ' openjdk-8-jdk'
+  } else {
+    // Assume a Suse distro
```
brianjmurrell, line 52: What if we introduce a new distro that doesn't match this name?  I think a branch for each different java package name is needed with a fall-through of an an `error("Don't know which java package ...")` for the new, unknown distro.

vars/cloverReportPublish.groovy
```diff
@@ -0,0 +1,128 @@
...
+                 inst_rpms: inst_rpms
+
+  String coverage_website = config.get('coverage_website',
+                                       'coverage_website.zip')
+
+  String url_base = env.JENKINS_URL +
+                    'job/daos-stack/job/tools/job/master' +
+                    '/lastSuccessfulBuild/artifact/'
+
+  httpRequest url:  url_base + 'bullseyecoverage-linux.tar',
```
brianjmurrell, line 69: Is this tarball available outside of our jenkins or do we build it in some way?  If we are just caching it, we should start to add branching here for getting things from their canonical source if the site doesn't have a local cache.

vars/cloverReportPublish.groovy
```diff
@@ -0,0 +1,128 @@
...
+  sh label: 'Create Coverage Report',
+     script: config.get('coverage_script', 'ci/bullseye_generate_report.sh')
+
+  def cb_result = currentBuild.result
+  step([$class: 'CloverPublisher',
+        cloverReportDir: 'test_coverage',
+        cloverReportFileName: 'clover.xml',
+        healthyTarget: config.get('coverage_healthy',
+                                   [methodCoverage: 70,
+                                    conditionalCoverage: 80,
```
brianjmurrell, line 112: We're nowhere near these thresholds are we?  What happens to runs that are below the thresholds?

vars/sconsBuild.groovy
```diff
@@ -156,19 +191,26 @@ def call(Map config = [:]) {
...
         }
     }
 
+    // Builds of Raft may fail if this directory is present in the workspace
+    // from a previous run.
+    // Builds may fail if bandit.xml files present in the workspace.
+    sh label: 'Remove some old build files if present.',
+       script:'rm -rf src/rdb/raft/CLinkedListQueue bandit.xml test.cov',
+       returnStatus: true
+
```
brianjmurrell, line 213: Aren't these just additional `clean_files`?

vars/sconsBuild.groovy
```diff
@@ -212,8 +254,10 @@ def call(Map config = [:]) {
         script += "PATH+=:${WORKSPACE}/cov_analysis/bin\n"
         scons_exe = "cov-build --dir cov-int " + scons_exe
     }
+
     script += '''# the config cache is unreliable so always force a reconfig
                  # with "--config=force"
+                 pwd
```
brianjmurrell, line 260: Debug needing removed?

But more importantly, shouldn't this all be in a `ci/*` script now?

vars/sconsBuild.groovy
```diff
@@ -251,5 +295,26 @@ def call(Map config = [:]) {
...
+        if (stage_info['build_type']) {
+            target_stash += '-' + stage_info['build_type']
+        }
+        String install_includes = 'install/**'
+        if (stage_info['compiler'] == 'covc') {
+            println "sconsBuild stashing test.cov"
+            install_includes += ', test.cov'
+        }
+        println "sconsBuild install_includes=${install_includes}"
+        sh "ls -l test.cov || true"
```
brianjmurrell, line 310: More debug?

vars/sconsBuild.groovy
```diff
@@ -251,5 +295,26 @@ def call(Map config = [:]) {
     if (!config['returnStatus'] && (rc != 0)) {
       error "sconsBuild failed for ${full_script}"
     }
+    if (config['test_files']) {
```
brianjmurrell, line 298:
```suggestion
    if (config['stash_files']) {
```

Maybe a bit more self-explanatory when reading the calling side of this function?

vars/unitTest.groovy
```diff
@@ -69,20 +74,50 @@ def call(Map config = [:]) {
...
+  if (stage_info['compiler'] == 'covc') {
+    if (!inst_rpms) {
+      inst_rpms = ''
+    }
+    if (stage_info['target'].startsWith("centos")) {
+      inst_rpms += ' java-1.8.0-openjdk'
+    } else if (stage_info['target'].startsWith('ubuntu')) {
+      inst_rpms += ' openjdk-8-jdk'
+    } else {
+      // Assume a Suse distro
```
brianjmurrell, line 88: Ditto for the assumption here, but perhaps this should be DRY'd out into a single function so that adding new distros only requires updating in one place.

vars/cloverReportPublish.groovy
JohnMalmberg: Those are just the default thresholds for the CloverPublisher plugin.
Currently we are overriding them in the Jenkinsfile.
Eventually who ever is maintaining the coverage tests will need to set some reasonable defaults here, so that the Jenkinsfile overrides are not needed.

I am not sure what happens if we go below the thresh hold.  I am just setting up the framework that can be used in future PRs.

vars/sconsBuild.groovy
JohnMalmberg: They could be considered that.  But that other clean files step can be skipped and if these are not cleaned up builds can fail.   There is a ticket to fix the makefile for the raft file.

vars/sconsBuild.groovy
JohnMalmberg: Not sure why we put a comment in an imbedded script.  That does seem odd.
Moving the comment in the next push.

And yes, we want to get rid of building inline scripts.

I am mainly concentrating on adding coverage builds and cleaning up the Jenkinsfile for this PR, so that may need to be a separate PR.

vars/sconsBuild.groovy
JohnMalmberg: Yes more debug code that I missed.

vars/sconsBuild.groovy
JohnMalmberg: Yes.  I will have to change the DAOS PR-3104 to match.

vars/unitTest.groovy
JohnMalmberg: Yes, need to DRY it out.
Will fix in next push.

JohnMalmberg:
DAOS pr used for testing is https://github.com/daos-stack/daos/pull/3104

vars/parseStageInfo.groovy
```diff
@@ -52,6 +52,14 @@ def call(Map config = [:]) {
       }
     }
 
+    if (result['target'].startsWith('centos')) {
+      result['java_pkg'] = 'java-1.8.0-openjdk'
+    } else if (result['target'].startsWith('ubuntu')) {
+      result['target'] = 'openjdk-8-jdk'
+    } else if (result['target'].startsWith('leap')) {
+      result['target'] = 'java-1_8_0-openjdk'
+    }
```
brianjmurrell, line 61: It would be nice to be obvious here if a match couldn't be found rather than somebody having to back track it.
```suggestion
    } else {
        error("Don't know which java package to install for " + result['target'])
    }
```

vars/sconsBuild.groovy
brianjmurrell: My concern wasn't the comments in the inline script but rather the addition of the `pwd` to the script.  Is that debug that should be removed now?

vars/parseStageInfo.groovy
JohnMalmberg: If we add another distro to our support, we are likely going to have to make many changes in this file and others.

By leaving the key unspecified with out a match, it is up to the routine that consumes that information to generate the diagnostic if it is missing.  We are also now pre-installing the Java package on the CI nodes.

vars/sconsBuild.groovy
JohnMalmberg: I did not realize I was the one that added the pwd.  It will be removed in the next push.

vars/sconsBuild.groovy
```diff
@@ -87,21 +111,32 @@ def call(Map config = [:]) {
...
+                  outputFile: 'bullseye.tar'
+    }
+
     def set_cwd = ''
     if (config['directory']) {
         set_cwd = "cd ${config['directory']}\n"
-    } else if (config['target']) {
-        set_cwd = "cd ${config['target']}\n"
+    } else if (config_target) {
+        println "sconsLocal config_target = ${config_target}"
```
brianjmurrell, line 136: Debug that needs removing?

vars/sconsBuild.groovy
```diff
@@ -87,21 +111,32 @@ def call(Map config = [:]) {
...
+        set_cwd = "cd ${config_target}\n"
     }
+    println "sconsLocal set_cwd=${set_cwd}"
```
brianjmurrell, line 139: Ditto.

vars/parseStageInfo.groovy
brianjmurrell: But then we are going to be repeating a lot of code in all of the callers to validate the info was parsed correctly.  It seems more efficient and the right place to flag missing information where it's being parsed/detected.

JohnMalmberg: I have put that in for the next push.

Someone adding a new target distribution for a stage that did not need Java might be surprised when this error shows up.

brianjmurrell: I can see that perspective, indeed.  But I think porting a new distro should be complete anyway, so that future expansion doesn't need to deal with a previously incomplete distro port.

But saying that makes me think somehow getting all of these various distro variations across our many files and functions into a single location and then referenced from there would be a good enhancement one day.  A single place to go put all of the bits needed to teach our CI about a new distro.

JohnMalmberg:
Hold on any merging.   I need to refactor cloverReportPublish.groovy

JohnMalmberg:
DAOS PR-3104 job was used to test this PR.

jolivier23:
is it ok to land now?

JohnMalmberg:
Just needs the recent changes reviewed.

PR #162, opened 2020-08-12
DAOS-5371 tests: Junit results from Avocado not reported in Jenkins

phender:
As part of https://github.com/daos-stack/daos/pull/3190 the cmocka JUnit
files are now found in the avocado/job-results/job-\*/test-results/\*/data.
This change includes adding that path when searching for JUnit files.


phender:
This change has been tested with https://build.hpdd.intel.com/job/daos-stack/job/daos/job/PR-3190/16/ and the cmocka JUnit files that are now created in the avocado/job-results/job-*/test-results/*/data directory are included in the Test Results:
image

vars/functionalTestPost.groovy
```diff
@@ -30,6 +30,7 @@ def call(Map config = [:]) {
 
   def junit_results = config.get('testResults',
                                  'Functional/*/results.xml, ' +
+                                 'Functional/*/test-results/*/data/*_results.xml, ' +
                                  'install/lib/daos/TESTING/ftest/*_results.xml')
```
brianjmurrell, line 34: So is this path even being used at all any more, on master, say?  Should we remove it at this point?

vars/runTestFunctional.groovy
```diff
@@ -53,7 +53,9 @@ void call(Map config = [:]) {
                        "OPERATIONS_EMAIL=" + env.OPERATIONS_EMAIL + ' ' +
                        "ci/functional/test_main.sh"
                                      
-    config['junit_files'] = "install/lib/daos/TESTING/ftest/avocado/job-results/job-*/*.xml install/lib/daos/TESTING/ftest/*_results.xml"
+    config['junit_files'] = "install/lib/daos/TESTING/ftest/avocado/job-results/job-*/*.xml " +
+                            "install/lib/daos/TESTING/ftest/*_results.xml " +
```
brianjmurrell, line 57: Ditto.

vars/functionalTestPost.groovy
JohnMalmberg: Will that affect the runs on the release branches?

brianjmurrell: I don't believe so, since the original path for that file being wrong/unwritable is preventing results from being written there currently.  Is that your thought too @phender?

I did look at a master run and noticed that it's not matching any files with that pattern anyway.  I.e.:

```
[2020-08-13T04:24:50.289Z] + grep '<error ' install/lib/daos/TESTING/ftest/avocado/job-results/job-2020-08-13T04.00-8fee5ac-io-fio_small/results.xml install/lib/daos/TESTING/ftest/avocado/job-results/job-2020-08-13T04.23-3d18072-checksum-csum_error_logging/results.xml
```

phender: > So is this path even being used at all any more, on master, say? Should we remove it at this point?

I was wondering that myself.  I felt it was in benign to leave "install/lib/daos/TESTING/ftest/*_results.xml" in the list.  Is this only included to pickup the daos_test cmoka JUnit files or is there something else we expect to be generating files in this path?

brianjmurrell: The only two JUnit results we were ever trying to get were the cmocka ones and the Avocado ones.

phender: > Will that affect the runs on the release branches?

I'm not 100% sure what you're asking John.  My understanding is that this will just look in an additional location for JUnit files.  Standalone it should not have any adverse effects as the pattern is a standard avocado test results directory.  When combined with the changes in https://github.com/daos-stack/daos/pull/3190 it allows for the collection of daos_test cmocka JUnit files that are now written to this location.

phender: Are we in agreement that "install/lib/daos/TESTING/ftest/*_results.xml" can be removed from the list of directories to search?  I'm all for cleaner code if we are sure it won't cause a problem.

brianjmurrell: @JohnMalmberg is just trying to make sure that if we remove that now, that existing PRs (before daos-stack/daos#3190) lands won't break.

My opinion is that given that this is the path to what are currently the non-existent cmock junit results, it's not working as expected currently anyway.  Second looks at my opinion greatly appreciated.

phender: The only situation I can think of where the removal of "install/lib/daos/TESTING/ftest/*_results.xml" would be a problem is if a PR was testing the build files (as opposed to the RPM files) before #3190 lands.

brianjmurrell: > The only situation I can think of where the removal of "install/lib/daos/TESTING/ftest/*_results.xml" would be a problem is if a PR was testing the build files (as opposed to the RPM files) before #3190 lands. Since cmoka will be able to write JUnit files in the "install/lib/daos/TESTING/ftest/" directory in that case.

Which I quite doubt.  In any case, such a result wouldn't be any worse than the default on master right now.  The only really bad fallout would be the `junit` step outright failing because the glob doesn't expand and it tries to find a file with the glob in it's name.  But I think we are pretty confident that is not happening or it would have broken long before now.  Agree?

phender: Yes. I've removed the out of date path for the cmocka JUnit files.

vars/runTestFunctional.groovy
phender: Done

PR #163, opened 2020-09-01
Corci 887b bug

JohnMalmberg:


JohnMalmberg:
Needed only for bug demostration.

PR #167, opened 2020-11-03
CORCI-1010 test: Add support for dedicated NLT stage.

ashleypittman:
Add support for NLT jobs, and make the calling of recordIssues optional in unitTestPost.

The logic is slightly complicated by the requirement to be backwards compatible.

### DIFF
--- a/vars/parseStageInfo.groovy
+++ b/vars/parseStageInfo.groovy
@@ -149,9 +149,9 @@ def call(Map config = [:]) {
     }
 
     if (env.STAGE_NAME.contains('NLT')) {
-      result['NLT'] = True
+      result['NLT'] = true
     } else {
-      result['NLT'] = False
+      result['NLT'] = false
     }
 
     if (env.STAGE_NAME.contains('Unit Test') &&
--- a/vars/parseStageInfo.groovy
+++ b/vars/parseStageInfo.groovy
@@ -149,7 +149,7 @@ def call(Map config = [:]) {
     }
 
     if (env.STAGE_NAME.contains('NLT')) {
-      result['NLT'] = True     
+      result['NLT'] = True
     } else {
       result['NLT'] = False
     }
--- a/vars/parseStageInfo.groovy
+++ b/vars/parseStageInfo.groovy
@@ -149,7 +149,9 @@ def call(Map config = [:]) {
     }
 
     if (env.STAGE_NAME.contains('NLT')) {
-      result['NLT'] = 1
+      result['NLT'] = True     
+    } else {
+      result['NLT'] = False
     }
 
     if (env.STAGE_NAME.contains('Unit Test') &&
--- a/vars/parseStageInfo.groovy
+++ b/vars/parseStageInfo.groovy
@@ -148,6 +148,10 @@ def call(Map config = [:]) {
       result['ftest_arg'] = config['ftest_arg']
     }
 
+    if (env.STAGE_NAME.contains('NLT)) {
+      result['NLT'] = 1
+    }
+
     if (env.STAGE_NAME.contains('Unit Test') &&
         env.STAGE_NAME.contains('memcheck')) {
         result['with_valgrind'] = 'memcheck'
--- a/vars/parseStageInfo.groovy
+++ b/vars/parseStageInfo.groovy
@@ -148,7 +148,7 @@ def call(Map config = [:]) {
       result['ftest_arg'] = config['ftest_arg']
     }
 
-    if (env.STAGE_NAME.contains('NLT)) {
+    if (env.STAGE_NAME.contains('NLT')) {
       result['NLT'] = 1
     }
 
--- a/vars/unitTestPost.groovy
+++ b/vars/unitTestPost.groovy
@@ -37,17 +37,19 @@ def call(Map config = [:]) {
 
   Map stage_info = parseStageInfo(config)
 
-  double health_scale = 1.0
-  if (config['ignore_failure']) {
-    health_scale = 0.0
-  }
+  if (config['testResults'] != 'None' ) {
+    double health_scale = 1.0
+    if (config['ignore_failure']) {
+      health_scale = 0.0
+    }
 
-  def cb_result = currentBuild.result
-  junit testResults: config.get('testResults', 'test_results/*.xml'),
-        healthScaleFactor: health_scale
+    def cb_result = currentBuild.result
+    junit testResults: config.get('testResults', 'test_results/*.xml'),
+          healthScaleFactor: health_scale
 
-  if (cb_result != currentBuild.result) {
-    println "The junit plugin changed result to ${currentBuild.result}."
+    if (cb_result != currentBuild.result) {
+      println "The junit plugin changed result to ${currentBuild.result}."
+    }
   }
 
   if(stage_info['with_valgrind']) {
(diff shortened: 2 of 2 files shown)


vars/parseStageInfo.groovy
```diff
@@ -148,6 +148,10 @@ def call(Map config = [:]) {
       result['ftest_arg'] = config['ftest_arg']
     }
 
+    if (env.STAGE_NAME.contains('NLT')) {
+      result['NLT'] = 1
```
brianjmurrell, line 152: This seems like a boolean, but moreover, shouldn't rely on unset values.
```suggestion
      result['NLT'] = True
    } else {
        result['NLT'] = False
```

vars/unitTestPost.groovy
```diff
@@ -80,15 +86,9 @@ def call(Map config = [:]) {
...
-
-    // Need to leave this logic in here for backwards compatibility.
-    // Valgrind results need to stashed and reported in a common stage
-    // After all Valgrind tests are run.
-    valgrindReportPublish ignore_failure: ignore_failure,
-                          valgrind_stashes: []
   }
-  if (!stage_info['with_valgrind']) {
+  def no_record_issues = config.get('no_record_issues', true)
+  if ((!stage_info['with_valgrind'] && no_record_issues) || (stage_info['NLT'] == 1)) {
```
brianjmurrell, line 91: Since it's boolean:
```suggestion
  if ((!stage_info['with_valgrind'] && no_record_issues) || (stage_info['NLT'] == True)) {
```

Or perhaps even:

```suggestion
  if ((!stage_info['with_valgrind'] && no_record_issues) || (stage_info['NLT'])) {
```

Not sure if the evaluation there will evaluate the result or that the variable just has a value.

vars/unitTestPost.groovy
```diff
@@ -80,15 +86,9 @@ def call(Map config = [:]) {
...
-  } else {
-
-    // Need to leave this logic in here for backwards compatibility.
-    // Valgrind results need to stashed and reported in a common stage
-    // After all Valgrind tests are run.
-    valgrindReportPublish ignore_failure: ignore_failure,
-                          valgrind_stashes: []
   }
-  if (!stage_info['with_valgrind']) {
+  def no_record_issues = config.get('no_record_issues', true)
```
brianjmurrell, line 90: The name of this variable seems to be inverted.  On line 93, `recordIssues` is being invoked when the conditional on line 91 `no_record_issues` is `true`, yes?  Shouldn't this variable (and the equivalent config option) be `record_issues`?

brianjmurrell (changes requested):

vars/parseStageInfo.groovy
ashleypittman: Agreed.  I was basing it on the 'with_valgrind' setting which relies on unset values.  Not being familiar with Groovy I wasn't confident to try booling vars, will make changes as suggested.

vars/unitTestPost.groovy
ashleypittman: I'll try this second one.

vars/unitTestPost.groovy
ashleypittman: Yes.  This is a compatibility issue, the code needs to work (for a time) with master daos but with this PR, if I change the case then legacy builds of master won't record the issues.

brianjmurrell: But shouldn't the variable (and config argument) name be `record_issues` though?

vars/parseStageInfo.groovy
brianjmurrell: > Not being familiar with Groovy I wasn't confident to try booling vars

I'm only just getting familiar with it so some of my suggestions might not actually work.  But just thinking we should try to embrace the language as much as we can and use it's types when we can.

vars/unitTestPost.groovy
ashleypittman: If it was record_issues then master daos with this PR would not have entered the conditional.

brianjmurrell: I'm not talking about changing the logic in any way.  Just the variable/config item name.

The issue is that the name reads such that when `no_record_issue` == `true`, one believes that to mean that `recordIssues` will not be run.  However the logic on line 91 says that when `no_record_issue` == `true` `recordIssues` will be run.

Or am I totally missing something here?

ashleypittman: I did get quite confused when I was writing it.  I'll take another look at it tomorrow, it might just be that I can invert everything including the default and it would work.

vars/parseStageInfo.groovy
ashleypittman: Agreed.  I was just trying the simplest thing to get it working initially.

vars/unitTestPost.groovy
```diff
@@ -37,17 +41,19 @@ def call(Map config = [:]) {
...
+  if (config['testResults'] != 'None' ) {
+    double health_scale = 1.0
+    if (config['ignore_failure']) {
+      health_scale = 0.0
+    }
 
-  def cb_result = currentBuild.result
-  junit testResults: config.get('testResults', 'test_results/*.xml'),
-        healthScaleFactor: health_scale
+    def cb_result = currentBuild.result
```
JohnMalmberg, line 50: This makes cb_results local to the if block that it is in.

vars/unitTestPost.groovy
```diff
@@ -37,17 +41,19 @@ def call(Map config = [:]) {
...
-  if (config['ignore_failure']) {
-    health_scale = 0.0
-  }
+  if (config['testResults'] != 'None' ) {
+    double health_scale = 1.0
+    if (config['ignore_failure']) {
+      health_scale = 0.0
+    }
 
-  def cb_result = currentBuild.result
```
JohnMalmberg, line 45: This cb_result was local to the procedure.

vars/unitTestPost.groovy
```diff
@@ -80,15 +86,9 @@ def call(Map config = [:]) {
...
-    // Need to leave this logic in here for backwards compatibility.
-    // Valgrind results need to stashed and reported in a common stage
-    // After all Valgrind tests are run.
-    valgrindReportPublish ignore_failure: ignore_failure,
-                          valgrind_stashes: []
   }
-  if (!stage_info['with_valgrind']) {
+  def record_issues = config.get('record_issues', true)
+  if ((!stage_info['with_valgrind'] && record_issues) || stage_info['NLT']) {
     cb_result = currentBuild.result
```
JohnMalmberg, line 92: This cb_result is global to the entire Jenkins job in progress.

JohnMalmberg (changes requested):
Need to fix the scope issues for cb_result

vars/unitTestPost.groovy
ashleypittman: I've pushed a new version to make this local.

JohnMalmberg (approved):

vars/unitTestPost.groovy
```diff
@@ -87,9 +87,9 @@ def call(Map config = [:]) {
     def valgrind_pattern = config.get('valgrind_pattern', '*.memcheck.xml')
     stash name: config['valgrind_stash'], includes: valgrind_pattern
   }
-  def no_record_issues = config.get('no_record_issues', true)
-  if ((!stage_info['with_valgrind'] && no_record_issues) || (stage_info['NLT'] == 1)) {
-    cb_result = currentBuild.result
+  def record_issues = config.get('record_issues', true)
+  if ((!stage_info['with_valgrind'] && record_issues) || stage_info['NLT']) {
+    def cb_result = currentBuild.result
```
brianjmurrell, line 92: I think @JohnMalmberg mentioned that you have scoped `cb_result here` to just this `if` block and it will not be available where you use it below.

brianjmurrell (changes requested):

vars/unitTestPost.groovy
ashleypittman: I thought I'd fixed that, it's used twice and in scope for just where required both times.  Then again I've not done much groovy.

brianjmurrell: Hrm.  Maybe you did and it's just all of the leading whitespace in the side-by-side diff that is making me think they are not in the same scope.

brianjmurrell (approved):

PR #174, opened 2020-12-09
Only run the 'pre' tag on PRs, not 'pr'.

ashleypittman:
This probably wants new tags, for example 'pr' for prs and 'landings'
for master but using the pr tag for master here means fewer changes
and potential conflicts when testing this change.

### DIFF
--- a/vars/parseStageInfo.groovy
+++ b/vars/parseStageInfo.groovy
@@ -115,7 +115,7 @@ def call(Map config = [:]) {
 
     if (env.STAGE_NAME.contains('Functional')) {
       if (env.BRANCH_NAME == 'master') {
-          core_tag = 'pr'
+          core_tag = 'pr,pre'
       } else {
           core_tag = 'pre'
       }
--- a/vars/parseStageInfo.groovy
+++ b/vars/parseStageInfo.groovy
@@ -114,22 +114,27 @@ def call(Map config = [:]) {
     result['node_count'] = 1
 
     if (env.STAGE_NAME.contains('Functional')) {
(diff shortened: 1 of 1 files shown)


brianjmurrell:
What's the goal here?

ashleypittman:
> What's the goal here?

https://github.com/daos-stack/daos/pull/4059

Have a seperate test list for PRs as opposed to landings/master.

vars/parseStageInfo.groovy
```diff
@@ -114,22 +114,27 @@ def call(Map config = [:]) {
     result['node_count'] = 1
 
     if (env.STAGE_NAME.contains('Functional')) {
+      if (env.BRANCH_NAME == 'master') {
+          core_tag = 'pr'
```
brianjmurrell, line 118: Why is the `pr` tag being used for master and...

vars/parseStageInfo.groovy
```diff
@@ -114,22 +114,27 @@ def call(Map config = [:]) {
...
+      } else {
+          core_tag = 'pre'
```
brianjmurrell, line 120: The tag for non-master (PRs, but actually any other [release] branch) now `pre`?  What does `pre` mean?

ashleypittman: I put this in one of the comments, the tags should be "pr", and "landing" but in order to test that I'd need to do a search-and-replace across all tests, so for now I'm using "pre" for tests that run on PRs, and "pr" for tests that run on master, just to prove the concept with making minimal changes to the src.

PR #180, opened 2020-12-21
Add stage name to e-mail subject

brianjmurrell:
Also extend dockerBuildArgs() to do some steps optionally, defaulting to
existing behaviour.

### DIFF
--- a/vars/stepResult.groovy
+++ b/vars/stepResult.groovy
@@ -114,7 +114,7 @@ def call(Map config= [:]) {
                  ".  " + comment_url
 
       if (commitPragma(pragma: "Skip-PR-comments") == "true") {
-          emailext subject: "${config.name} status: ${config.result}",
+          emailext subject: "${JOB_NAME} ${config.name} stage ${config.result}",
                    recipientProviders: [[$class: 'RequesterRecipientProvider'],
                                         [$class: 'DevelopersRecipientProvider']],
                    body: msg +
--- a/vars/stepResult.groovy
+++ b/vars/stepResult.groovy
@@ -114,7 +114,7 @@ def call(Map config= [:]) {
                  ".  " + comment_url
 
       if (commitPragma(pragma: "Skip-PR-comments") == "true") {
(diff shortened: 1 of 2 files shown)


vars/stepResult.groovy
```diff
@@ -114,7 +114,7 @@ def call(Map config= [:]) {
                  ".  " + comment_url
 
       if (commitPragma(pragma: "Skip-PR-comments") == "true") {
-          emailext subject: "${JOB_NAME} ${config.name} stage ${config.result}",
+          emailext subject: "${config.name} status: ${config.result}",
```
phender, line 117: This doesn't seem to align with the PR description.  Isn't this removing the job name from the email subject?

ashleypittman:
On reflection I'm not sure what emails this relates to.  All mine seem to come from Github, aside from the occasional master failed to build one.

vars/stepResult.groovy
brianjmurrell: Ick.  Yeah.  Let me take another look at this.  It must just be the one other part that needs to be in this PR.

brianjmurrell:
> On reflection I'm not sure what emails this relates to.

This refers to the e-mails that Jenkins will send directly, when it doesn't add comments (which are noise in my opinion) to PRs when you add the `Skip-PR-comments: true` commit pragma to your commit.

https://github.com/daos-stack/pipeline-lib/blob/af4fdedcf49288a09022ec83c7cf04ff4113e912/vars/stepResult.groovy#L116-L135

> All mine seem to come from Github, aside from the occasional master failed to build one.

Yes, without `Skip-PR-comments: true` that is the current behaviour.  This is a new (optional, default disabled) behaviour I am testing out to see how much less noisy PRs can be when we don't spam it with comments about every failed build/test.  Ultimately, I find those useful at the time of the failure, but being left behind in the history of the PR is making our PRs messy and noisy, and sometimes unrenderable (a pink unicorn) in GitHub.

jolivier23 (approved):

jolivier23:
Looks like it needs to merge latest master

PR #181, opened 2020-12-22
CORCI-1032 nlt: Do not ignore LOW severity regressions.

ashleypittman:
This was allowed due to race conditions in the code but they're fixed
so remove the provision for this.

Tidy up some legacy code that's not required any more.

### DIFF
--- a/vars/unitTestPost.groovy
+++ b/vars/unitTestPost.groovy
@@ -27,9 +27,6 @@
    *                               Required if more than one stage is
    *                               creating valgrind reports.
    *
-   * config['record_issues']       Call recordIssues for Unit test,
-   *                               legacy option for backwards compatability.
-   *
    */
 
 def call(Map config = [:]) {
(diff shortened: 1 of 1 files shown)


ashleypittman:
https://github.com/daos-stack/daos/pull/4160 needs to land after this.

brianjmurrell (dismissed):

brianjmurrell (commented):
What is going to happen to DAOS PRs in flight that are expecting the old behaviour (i.e. that are still requesting `unitTestPost` with `record_issues: false`)?

ashleypittman:
> What is going to happen to DAOS PRs in flight that are expecting the old behaviour (i.e. that are still requesting `unitTestPost` with `record_issues: false`)?

I think it's OK, as they're providing the default value.  The only concern I had was if they passed a now known parameter, but I don't know how that's checked in Groovy.  The whole record_issues thing was only for backwards compatibility when I split NLT out from unit test so it's all legacy now, it's just a question of how we unwind it.

brianjmurrell:
> I think it's OK, as they're providing the default value. The only concern I had was if they passed a now known parameter, but I don't know how that's checked in Groovy. The whole record_issues thing was only for backwards compatibility when I split NLT out from unit test so it's all legacy now, it's just a question of how we unwind it.

Oh yes.  I was losing sight of the removal of `!stage_info['with_valgrind'] && record_issues) ||` from the `if`.

Unused (i.e. unknown) values in `config` are fine.  It's just a `Map` (hash or associative array, in other languages) which values are pulled from in the Groovy function.  We could just very well pass positional parameters instead of a Map.  A Map is just really easy to work with, adding and removing function _parameters_ without having to worry about compatibility of function signatures, etc.  Frankly though, we have probably over-used it, and could use some real positional parameters to functions in some places.

brianjmurrell (approved):

phender (approved):

PR #184, opened 2021-01-08
DAOS-6274 test: Restructure DAOS logs

sylviachanoiyee:
Restructure DAOS logs


brianjmurrell (commented):
This change breaks the API.  No daos commit prior to the one that corresponds to these changes will be able to run now because there will be a mismatch between the directory the daos commit puts it's artifacts in an where pipeline-lib expects to find them.

If this change was really necessary, a new version of this function should have been created and the daos commit should call that new version, not this one (which should have remained unchanged).

PR #185, opened 2021-01-09
Only run runnable tests in Features: pragma

brianjmurrell:
Limit the tests added by Features: tags to tests that we normally run
in one of our existing test buckets.

Skip-PR-comments: true

### DIFF
--- a/vars/parseStageInfo.groovy
+++ b/vars/parseStageInfo.groovy
@@ -141,7 +141,11 @@ def call(Map config = [:]) {
       result['test_tag'] = branch_tag + ',' + cluster_size
 
       for (feature in commitPragma(pragma: "Features").split(' ')) {
-        result['test_tag'] += ' ' + feature + ',' + cluster_size
+        // Add tests from all testing buckets
+        // Duplication with branch_tag is not a problem
+        result['test_tag'] += ' pr,' + feature + ',' + cluster_size
+        result['test_tag'] += ' daily_regression,' + feature + ',' + cluster_size
+        result['test_tag'] += ' full_regression,' + feature + ',' + cluster_size
       }
     }
     if (config['test']) {


vars/parseStageInfo.groovy
```diff
@@ -141,7 +141,11 @@ def call(Map config = [:]) {
       result['test_tag'] = branch_tag + ',' + cluster_size
 
       for (feature in commitPragma(pragma: "Features").split(' ')) {
-        result['test_tag'] += ' ' + feature + ',' + cluster_size
+        // Add tests from all testing buckets
+        // Duplication with branch_tag is not a problem
+        result['test_tag'] += ' pr,' + feature + ',' + cluster_size
```
mjean308, line 146: I just have a question on how this works
say the Feature pragma is "OSA"
branch_tag is "pr"
cluster_size is "hw,small"

The final result['test_tag"] would look something like

pr,hw,small pr,OSA,hw,small daily_regression,OSA,hw,small full_regression.OSA,hw,small

Why add in the daily and the full if the branch_tag is "pr"

brianjmurrell: Because `Features: osa` is asking to run all (currently, but perhaps some curated subset of all at some time in the future) tests tagged with `osa`.  @skirvan44 has tickets open and assigned to some test developers to work with the other development leads to curate lists of tests that represent _features_ but those are at a lower priority at the moment.

Recall that the entire reason for the `Features:` pragma was because the set of tests that have `pr` on them has been [(significantly?) reduced](https://github.com/daos-stack/daos/commit/e72960b88).  Continuing with your `osa` example:

https://github.com/daos-stack/daos/commit/e72960b88#diff-fe6b5d2447f23442c82b57ba5b1799cd63a3a96a7573f69dff3f3b517666bc7b
https://github.com/daos-stack/daos/commit/e72960b88#diff-296edc79e14385d7b5972857691fd193173d9f48329085f93117afc6c7cc369e
https://github.com/daos-stack/daos/commit/e72960b88#diff-296edc79e14385d7b5972857691fd193173d9f48329085f93117afc6c7cc369e
(and so on) demonstrate where `pr,osa` no longer runs many tests as they have been moved to other test _buckets_ (i.e. `daily_regression` and `weekly_regression`).

So in order to get as much `osa` test coverage, we need to get `osa` tagged tests from the other buckets.

We could suppose that we could simply not add any bucket names and just run all `osa` tests.  That was the approach being taken prior to this PR but it was very quickly demonstrated that there are tests that don't have any of the `pr`, `daily_regression` or `weekly_regression` tags assigned and therefore are not run ever or are not intended to be run in CI and are broken (some of those are only broken when run in CI, due to a lack of a cluster big enough for example).  So we need to limit the tests run to tests that are run with at least some frequency (PRs, daily or weekly).

JohnMalmberg (approved):

vars/parseStageInfo.groovy
mjean308: Thanks Brian

mjean308 (approved):
LGTM

vars/parseStageInfo.groovy
brianjmurrell: NPAA @mjean308.  Always happy to explain any time there is a question about why a change is being made.  This is of course exactly what code review is all about.  :smile:

Thanks for your review on this one!

PR #193, opened 2021-02-08
DAOS-6742: build: Update dockerBuildArgs to handle false correctly.

ashleypittman:
The logic for cachebust and add_repos was wrong in that they defaulted
to true, and could only be set to true.

Update they logic so they can also be set to false.

### DIFF
--- a/vars/dockerBuildArgs.groovy
+++ b/vars/dockerBuildArgs.groovy
@@ -36,7 +36,7 @@ String call(Map config = [:]) {
       deps_build = config['deps_build']
     }
 
-    ret_str = " --build-arg NOBUILD=1 --build-arg DAOS_BUILD=no" +
+    ret_str = " --build-arg NOBUILD=1 --build-arg " +
               " --build-arg UID=" + getuid() +
               " --build-arg JENKINS_URL=$env.JENKINS_URL"
     if (cachebust) {
--- a/vars/dockerBuildArgs.groovy
+++ b/vars/dockerBuildArgs.groovy
@@ -77,7 +77,9 @@ String call(Map config = [:]) {
       ret_str += ' --build-arg QUICKBUILD=true'
     }
     if (deps_build) {
-      ret_str += ' --build-arg DAOS_DEPS_BUILD=yes'
+      ret_str += ' --build-arg DAOS_DEPS_BUILD=yes --build-arg DAOS_BUILD=no'
+    } else {
+      ret_str += ' --build-arg DAOS_DEPS_BUILD=no'
     }
     ret_str += ' '
     return ret_str
--- a/vars/dockerBuildArgs.groovy
+++ b/vars/dockerBuildArgs.groovy
@@ -7,6 +7,7 @@
    *
    * config['qb']      Whether to generate Quick-build args
    *
+   * config['add_repos'] Whether to add yum repos to image.
    */
 
 // The docker agent setup and the provisionNodes step need to know the
@@ -20,10 +21,10 @@ int getuid() {
 String call(Map config = [:]) {
     Boolean cachebust = true
     Boolean add_repos = true
-    if (config['cachebust']) {
+    if (config.containsKey('cachebust')) {
       cachebust = config['cachebust']
     }
-    if (config['add_repos']) {
+    if (config.containsKey('add_repos')) {
       add_repos = config['add_repos']
     }
     ret_str = " --build-arg NOBUILD=1 --build-arg UID=" + getuid() +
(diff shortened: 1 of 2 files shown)


jolivier23 (approved):

JohnMalmberg (approved):

brianjmurrell (commented):
Comment was for wrong PR.  Will update here with my comments for this PR.

vars/dockerBuildArgs.groovy
```diff
@@ -20,13 +25,19 @@ int getuid() {
 String call(Map config = [:]) {
     Boolean cachebust = true
     Boolean add_repos = true
-    if (config['cachebust']) {
+    Boolean deps_build = false
+    if (config.containsKey('cachebust')) {
```
brianjmurrell, line 29: Just wondering what the difference in syntax is here.  My understanding is that an unset key is false in groovy.

vars/sconsBuild.groovy
```diff
@@ -189,12 +188,12 @@ def call(Map config = [:]) {
     if (config['skip_clean']) {
         clean_cmd += "echo 'skipping scons -c'\n"
     } else {
-        def clean_files = "_build.external"
+        def clean_files = "install"
```
brianjmurrell, line 191: This wouldn't be backward compatible with building an older commit of daos would it?

vars/sconsBuild.groovy
```diff
@@ -189,12 +188,12 @@ def call(Map config = [:]) {
...
         if (config['clean']) {
             clean_files = config['clean']
         }
-        clean_files += ' install build {daos_m,daos,iof,cart-Linux}.conf'
-        clean_files += ' .sconsign{,-Linux}.dblite .sconf-temp{,-Linux}'
+        clean_files += ' build daos.conf'
+        clean_files += ' .sconsign.dblite .sconf_temp'
```
brianjmurrell, line 196: Ditto.

brianjmurrell (commented):
Remember that all builds of daos, from any commit in the past or future will build with the "current" master of pipeline-lib.  So without versioning functions (to change their behaviour, which we have done, so that we have old and new behaviours) they must remain compatible with all daos code in the past.

vars/dockerBuildArgs.groovy
ashleypittman: It is.  The default is true so if you set it to false by a parameter then it doesn't update the default so it remained false.

vars/sconsBuild.groovy
ashleypittman: We dropped _build.external about a year ago.

vars/sconsBuild.groovy
ashleypittman: And the -Linux suffixes never existed in DAOS, only cart and IOF.

vars/sconsBuild.groovy
brianjmurrell: > We dropped _build.external about a year ago.

It doesn't matter how long ago it was.  If somebody checked out a PR from over a year ago, the clean would fail, where it would have worked when it was originally built.

vars/sconsBuild.groovy
brianjmurrell: And ditto.

vars/sconsBuild.groovy
ashleypittman: Do we support builds of DAOS from that long ago?  I can easily back out this part but I'd be amazed if that worked.

brianjmurrell: Just to answer the question...

> Do we support builds of DAOS from that long ago?

I haven't been told otherwise.  Sounds like a policy decision issue to me.  But note that this would break being able to build even our current (1.0.x) release.

> I can easily back out this part but I'd be amazed if that worked.

Maybe, maybe not, because we certainly don't actively test changes for API breakage.  Probably we could/should.  As it is, we just try to do our best with our eyeballs.  It's entirely possible that if somebody tried to build our 1.0.x release in CI it would break and we'd have to patch up the pipeline-lib API.  But we should avoid breaking it when our eyeballs show us it is getting broken.

I'm in fact un-breaking pipeline-lib API at this very moment, that landed yesterday or the day before.

vars/sconsBuild.groovy
brianjmurrell: > And the -Linux suffixes never existed in DAOS, only cart and IOF.

Indeed.  But they did in cart which is still separate in our current (1.0.x) release, so if we want to have a goal of at lest being able to build that, we need to maintain this compatibility.  Which I see you have done.  But just justifying why.

johannlombardi (approved):

PR #199, opened 2021-02-25
DAOS-6742 build: Add CB0 to CACHEBUST option to allow multi-stage.

ashleypittman:
This allows periodic rebuild as well as per-build refreshes, the
benefit of which is to keep the refresh time from increasing over
time as the base images age.

### DIFF
--- a/vars/dockerBuildArgs.groovy
+++ b/vars/dockerBuildArgs.groovy
@@ -9,8 +9,9 @@
    *
    * config['add_repos'] Whether to add yum repos to image.
    *
-   * config['cachebust'] Whether to set CB0 and CACHEBUST args.
-   *                     CB0 will be set weekly and should force entire rebuild
+   * config['cachebust'] Whether to set CB_WEEKLY and CACHEBUST args.
+   *                     CB_WEEKLY will be set weekly and should force entire
+   *                     rebuild
    *                     CACHEBUST should be unique and should force updates.
    *                     Defaults to true.
    *
@@ -46,7 +47,7 @@ String call(Map config = [:]) {
     if (cachebust) {
       Calendar current_time = Calendar.getInstance()
       ret_str += " --build-arg CACHEBUST=${currentBuild.startTimeInMillis}"
-      ret_str += " --build-arg CB0=" + current_time.get(Calendar.WEEK_OF_YEAR)
+      ret_str += " --build-arg CB_WEEKLY=" + current_time.get(Calendar.WEEK_OF_YEAR)
     }
 
     if (add_repos) {
--- a/vars/dockerBuildArgs.groovy
+++ b/vars/dockerBuildArgs.groovy
@@ -9,9 +9,13 @@
    *
    * config['add_repos'] Whether to add yum repos to image.
    *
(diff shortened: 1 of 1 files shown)


vars/dockerBuildArgs.groovy
```diff
@@ -40,7 +44,9 @@ String call(Map config = [:]) {
               " --build-arg UID=" + getuid() +
               " --build-arg JENKINS_URL=$env.JENKINS_URL"
     if (cachebust) {
+      Calendar current_time = Calendar.getInstance()
       ret_str += " --build-arg CACHEBUST=${currentBuild.startTimeInMillis}"
+      ret_str += " --build-arg CB0=" + current_time.get(Calendar.WEEK_OF_YEAR)
```
brianjmurrell, line 49: Why not be more transparent about it?
```suggestion
      ret_str += " --build-arg CB_WEEKLY=" + current_time.get(Calendar.WEEK_OF_YEAR)
```

Then, when reading the `Dockerfile`, it will be clear when to expect that to trigger.

brianjmurrell (commented):
I suppose it's useful to keep cachebusting optional, but given our security requirements, I wonder when we'd ever want it `false`.

vars/dockerBuildArgs.groovy
ashleypittman: Because we might change our policy on how often we do this, I've been deliberately not littering references to it throughout the code.   The checkpatch dockerfile is already using CB0 so that would have to be changed but we could do that if you felt it was worth it.  One of the advantages if having this in pipeline-lib actually is it would be easier to do this.

ashleypittman: Done, and I've pushed a new version to daos as well.

brianjmurrell: > Because we might change our policy on how often we do this,

Fair point.  Or other users of the Dockerfile can apply their own value to `CB0` to implement their own policy of how frequently to update at the `CB0` interval.

brianjmurrell: > Done, and I've pushed a new version to daos as well.

Apologies for only getting to this PR now.  Just had to many plates spinning the last day or two.  I actually don't disagree with your reasoning for keeping it more generic, such as `CB0` is.  So if you wanted to unwind the most recent change, I wouldn't have a problem with that.  I wonder if we can come up with something more descriptive as to what `CB0` is actually meant to be though, so that it's more obvious from the outside, rather than having to look through Dockerfiles to figure it out.  Maybe there is no better name.

brianjmurrell (approved):
Ultimately fine as is, or if you do prefer CB_WEEKLY being more generic, I would be fine with the previous iteration.  Up to you.

vars/dockerBuildArgs.groovy
ashleypittman: On reflection I'll change this back to CB0, as you say we intend for other users to be able to use these and a documented CB0 meets that need better.  I'll add a one-liner comment to the Dockerfiles next to it.

# Option to periodically force refresh of entire image, to avoid later updates growing unbounded.

brianjmurrell (approved):

PR #207, opened 2021-03-18
CORCI-1063 build: Support docker images repos

JohnMalmberg:
Support additional repositories for Dockerfiles.

DAOS_STACK_distro_ver_type_REPO where:
  distro is oneof CENTOS LEAP or UBUNTU
  ver is the major version of the distro
  type is one of "LOCAL", "STABLE", "RELEASE", "DEV"

  LOCAL is only the locally built artifacts
  STABLE is local artifacts + Proxied distro repositories
  RELEASE is release artifacts + Proxied distro repositories
  DEV is for testing new repositories.

Only pass the repository definitions that the dockerfile can
use.

### DIFF
--- a/vars/dockerBuildArgs.groovy
+++ b/vars/dockerBuildArgs.groovy
@@ -100,28 +100,23 @@ String call(Map config = [:]) {
       ret_str += " --build-arg CACHEBUST=${currentBuild.startTimeInMillis}"
       ret_str += " --build-arg CB0=" + current_time.get(Calendar.WEEK_OF_YEAR)
     }
-    if (add_repos) {
+
+    String repo_base = env.REPOSITORY_URL
+    // No repo_base, no repos to add.
+    if (add_repos && repo_base) {
       String repo_name = null
       if (stage_info['target'] == 'centos7') {
         repo_alias = 'EL_7'
-        repo_name = env["DAOS_STACK_${repo_alias}${repo_mod}_${repo_type}_REPO"]
-        if (repo_name) {
-          ret_str += ' --build-arg REPO_EL7=' + repo_name
-        }
+        repo_arg = 'EL7'
       }
       if (stage_info['target'] == 'centos8') {
         repo_alias = 'EL_8'
-        repo_name = env["DAOS_STACK_${repo_alias}${repo_mod}_${repo_type}_REPO"]
-        if (repo_name) {
-          ret_str += ' --build-arg REPO_EL8=' + repo_name
-        }
+        repo_arg = 'EL8'
       }
       if (stage_info['target'] == 'leap15') {
         repo_alias = 'LEAP_15'
-        repo_name = env["DAOS_STACK_${repo_alias}${repo_mod}_${repo_type}_REPO"]
-        if (repo_name) {
-          ret_str += ' --build-arg REPO_EL7=' + repo_name
-        }
+        repo_arg = 'LEAP'
+        // Backwards compatibilty for LOCAL
         if (repo_type == 'LOCAL') {
          if (env.DAOS_STACK_LEAP_15_GROUP_REPO) {
             ret_str += ' --build-arg REPO_GROUP_LEAP15=' +
@@ -130,15 +125,17 @@ String call(Map config = [:]) {
         }
       }
       if (stage_info['target'] == 'ubuntu20.04') {
-        repo_name = env.DAOS_STACK_UBUNTU_20_04_LOCAL_REPO
-        if (repo_type == 'LOCAL') {
-          repo_name = env.DAOS_STACK_UBUNTU_20_04_LOCAL_REPO
-          if (repo_name) {
-            ret_str += ' --build-arg REPO_UBUNTU_20_04=' + repo_name
-          }
-        }
+        // Ubuntu repos usage not yet implemented
+        // When it is implemented it will probably be similar to above.
+        // And the URLS for will be for installing a list of repos.
+        // Details still to be worked out.
+        repo_alias = 'UBUNTU_20_04'
+        repo_arg = 'REPO_UBUNTU'
       }
-      if (repo_name && env.REPOSITORY_URL) {
+      repo_name = env["DAOS_STACK_${repo_alias}${repo_mod}_${repo_type}_REPO"]
+      // Only add the build args if a repo was found.
+      if (repo_name) {
+        ret_str += " --build-arg REPO_${REPO_ARG}=" + repo_name
         ret_str += ' --build-arg REPO_URL=' + env.REPOSITORY_URL
       }
     }
--- a/vars/dockerBuildArgs.groovy
+++ b/vars/dockerBuildArgs.groovy
@@ -26,9 +26,9 @@
    * base URL that is used for all accesses to the repository.
    *
    * The rest of the URL for each specific repo is in additional environment
-   * varables with a name format of "DAOS_STACK_${mod}$(distro}_${type}_REPO".
+   * varables with a name format of "DAOS_STACK_$(distro}${mod}_${type}_REPO".
    *
-   * The ${mod} currently is "DOCKER_" for repositories that are for use
+   * The ${mod} currently is "_DOCKER" for repositories that are for use
    * with dockerfiles, and "" for generic repositories.
    * DOCKER repositories are for both replacing the distro built in
    * repositories and adding locally built packages.
@@ -90,7 +90,6 @@ String call(Map config = [:]) {
         repo_mod = '_DOCKER'
       }
     }
-
     Map stage_info = parseStageInfo(config)
 
     ret_str = " --build-arg NOBUILD=1 " +
@@ -102,27 +101,24 @@ String call(Map config = [:]) {
       ret_str += " --build-arg CB0=" + current_time.get(Calendar.WEEK_OF_YEAR)
     }
     if (add_repos) {
-      if (env.REPOSITORY_URL) {
-        ret_str += ' --build-arg REPO_URL=' + env.REPOSITORY_URL
-      }
       String repo_name = null
       if (stage_info['target'] == 'centos7') {
         repo_alias = 'EL_7'
-        repo_name = env["DAOS_STACK_${repo_mod}${repo_alias}_${repo_type}_REPO"]
+        repo_name = env["DAOS_STACK_${repo_alias}${repo_mod}_${repo_type}_REPO"]
         if (repo_name) {
           ret_str += ' --build-arg REPO_EL7=' + repo_name
         }
       }
       if (stage_info['target'] == 'centos8') {
         repo_alias = 'EL_8'
-        repo_name = env["DAOS_STACK_${repo_mod}${repo_alias}_${repo_type}_REPO"]
+        repo_name = env["DAOS_STACK_${repo_alias}${repo_mod}_${repo_type}_REPO"]
         if (repo_name) {
           ret_str += ' --build-arg REPO_EL8=' + repo_name
         }
       }
       if (stage_info['target'] == 'leap15') {
         repo_alias = 'LEAP_15'
-        repo_name = env["DAOS_STACK_${repo_mod}${repo_alias}_${repo_type}_REPO"]
+        repo_name = env["DAOS_STACK_${repo_alias}${repo_mod}_${repo_type}_REPO"]
         if (repo_name) {
           ret_str += ' --build-arg REPO_EL7=' + repo_name
         }
@@ -134,11 +130,16 @@ String call(Map config = [:]) {
         }
       }
       if (stage_info['target'] == 'ubuntu20.04') {
-        if ((repo_type == 'local') &&
-            (env.DAOS_STACK_UBUNTU_20_04_LOCAL_REPO)) {
-              ret_str += ' --build-arg REPO_UBUNTU_20_04=' +
-                         env.DAOS_STACK_UBUNTU_20_04_LOCAL_REPO
+        repo_name = env.DAOS_STACK_UBUNTU_20_04_LOCAL_REPO
+        if (repo_type == 'LOCAL') {
+          repo_name = env.DAOS_STACK_UBUNTU_20_04_LOCAL_REPO
+          if (repo_name) {
+            ret_str += ' --build-arg REPO_UBUNTU_20_04=' + repo_name
+          }
         }
+      }
+      if (repo_name && env.REPOSITORY_URL) {
+        ret_str += ' --build-arg REPO_URL=' + env.REPOSITORY_URL
       }
     }
     if (env.HTTP_PROXY) {
--- a/vars/dockerBuildArgs.groovy
+++ b/vars/dockerBuildArgs.groovy
@@ -126,12 +126,12 @@ String call(Map config = [:]) {
         }
       }
       if (stage_info['target'] == 'ubuntu20.04') {
-        // Ubuntu repos usage not yet implemented
-        // When it is implemented it will probably be similar to above.
+        // Ubuntu repos curently only used for package building.
+        // When fully implemented it will probably be similar to above.
         // And the URLS for will be for installing a list of repos.
         // Details still to be worked out.
         repo_alias = 'UBUNTU_20_04'
-        repo_arg = 'REPO_UBUNTU'
+        repo_arg = repo_alias
       }
       repo_name = env["DAOS_STACK_${repo_alias}${repo_mod}_${repo_type}_REPO"]
       // Only add the build args if a repo was found.
--- a/vars/dockerBuildArgs.groovy
+++ b/vars/dockerBuildArgs.groovy
@@ -30,6 +30,8 @@ String call(Map config = [:]) {
     Boolean cachebust = true
     Boolean add_repos = true
     Boolean deps_build = false
+    String repo_type = 'local'
+
     if (config.containsKey('cachebust')) {
       cachebust = config['cachebust']
     }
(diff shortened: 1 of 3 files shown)


vars/dockerBuildArgs.groovy
```diff
@@ -48,27 +55,75 @@ String call(Map config = [:]) {
...
     if (add_repos) {
       if (env.REPOSITORY_URL) {
         ret_str += ' --build-arg REPO_URL=' + env.REPOSITORY_URL
       }
-      if (env.DAOS_STACK_EL_7_LOCAL_REPO) {
-        ret_str += ' --build-arg REPO_EL7=' + env.DAOS_STACK_EL_7_LOCAL_REPO
-      }
-      if (env.DAOS_STACK_EL_8_LOCAL_REPO) {
-        ret_str += ' --build-arg REPO_EL8=' + env.DAOS_STACK_EL_8_LOCAL_REPO
+      string repo_name = null
```
brianjmurrell, line 62: I have not quite dug into `string` vs. `String`, but we have been using the latter.

vars/dockerBuildArgs.groovy
```diff
@@ -48,27 +55,75 @@ String call(Map config = [:]) {
...
       }
-      if (env.DAOS_STACK_LEAP_15_GROUP_REPO) {
-        ret_str += ' --build-arg REPO_GROUP_LEAP15=' +
-                   env.DAOS_STACK_LEAP_15_GROUP_REPO
+      if (stage_info['target'] == 'ubuntu20.04') {
+        if ((repo_type == 'local') &&
+            (env.DAOS_STACK_UBUNTU_20_04_LOCAL_REPO)) {
+              ret_str += ' --build-arg REPO_UBUNTU_20_04=' +
+                         env.DAOS_STACK_UBUNTU_20_04_LOCAL_REPO
+        }
```
brianjmurrell, line 126: It seems like this could be a lot smaller and maintainable by first setting a distro alias such as:

```groovy
String dist = ''
if (stage_info['target'] == 'centos7') {
    dist='EL_7'
} else if (stage_info['target'] == 'centos8') {
    dist = 'EL_8'
}
...
```
and then use `${dist}` in a single block of `repo_type` `if`s.  Not sure if variables will be resolved in the env. var name, but worth trying perhaps to get this huge block under a bit more control.

There are some alternative mechanisms to [get environment variables](https://stackoverflow.com/questions/40215394/how-to-get-environment-variable-in-jenkins-groovy-script-console) that might be a bit programmatic.

brianjmurrell (commented):
Can you expand on, or better yet, document somewhere the list/types of repos that will exist that map to what you have listed in the PR here and examples of what would be in them so that we don't have the same kind of recent confusion as we did?

vars/dockerBuildArgs.groovy
```diff
@@ -16,6 +16,47 @@
...
+   * The first envronment variable is "REPOSITORY_URL" which is a common
+   * base URL that is used for all accesses to the repository.
+   *
+   * The rest of the URL for each specific repo is in additional environment
+   * varables with a name format of "DAOS_STACK_$(distro}${mod}_${type}_REPO".
+   *
+   * The ${mod} currently is "_DOCKER" for repositories that are for use
+   * with dockerfiles, and "" for generic repositories.
+   * DOCKER repositories are for both replacing the distro built in
+   * repositories and adding locally built packages.
```
brianjmurrell, line 34: What's the motivation for having Docker specific repos?  How/why are they different from generic repos?

vars/dockerBuildArgs.groovy
```diff
@@ -56,75 +101,46 @@ String call(Map config = [:]) {
...
-          repo_name = env.DAOS_STACK_EL_7_DOCKER_STABLE_REPO
-        } else if ((repo_tpe == 'release') &&
-                   (env.DAOS_STACK_EL_7_DOCKER_RELEASE_REPO)) {
-          repo_name = env.DAOS_STACK_EL_7_DOCKER_RELEASE_REPO
-        } else if ((repo_type == 'dev') &&
-                   (env.DAOS_STACK_EL_7_DOCKER_DEV_REPO)) {
-          repo_name = env.DAOS_STACK_EL_7_DOCKER_DEV_REPO
-        }
+        repo_alias = 'EL_7'
+        repo_name = env["DAOS_STACK_${repo_alias}${repo_mod}_${repo_type}_REPO"]
```
brianjmurrell, line 107:
```suggestion
        repo_alias = 'EL_7'
        repo_arg = 'EL7'
```

and then assign:

```
        repo_name = env["DAOS_STACK_${repo_alias}${repo_mod}_${repo_type}_REPO"]
        ret_str += ' --build-arg REPO_' + repo_arg + '=' + repo_name
```

once at the end of the big loop, DRYing out the repeated duplicate assignments.  I didn't do all of the same edits for all of the branches below as some other cut-n-pasto(s) are making it confusing.

vars/dockerBuildArgs.groovy
```diff
@@ -56,75 +101,46 @@ String call(Map config = [:]) {
...
         if (repo_name) {
           ret_str += ' --build-arg REPO_EL7=' + repo_name
         }
       }
```
brianjmurrell:
```suggestion
```

vars/dockerBuildArgs.groovy
```diff
@@ -56,75 +101,46 @@ String call(Map config = [:]) {
...
-          repo_name = env.DAOS_STACK_EL_8_DOCKER_STABLE_REPO
-        } else if ((repo_tpe == 'release') &&
-                   (env.DAOS_STACK_EL_8_DOCKER_RELEASE_REPO)) {
-          repo_name = env.DAOS_STACK_EL_8_DOCKER_RELEASE_REPO
-        } else if ((repo_type == 'dev') &&
-                   (env.DAOS_STACK_EL_8_DOCKER_DEV_REPO)) {
-          repo_name = env.DAOS_STACK_EL_8_DOCKER_DEV_REPO
-        }
+        repo_alias = 'EL_8'
+        repo_name = env["DAOS_STACK_${repo_alias}${repo_mod}_${repo_type}_REPO"]
```
brianjmurrell, line 114:
```suggestion
        repo_arg = 'EL8'
```

vars/dockerBuildArgs.groovy
```diff
@@ -56,75 +101,46 @@ String call(Map config = [:]) {
...
         if (repo_name) {
           ret_str += ' --build-arg REPO_EL8=' + repo_name
         }
       }
```
brianjmurrell:
```suggestion
```

vars/dockerBuildArgs.groovy
```diff
@@ -56,75 +101,46 @@ String call(Map config = [:]) {
...
-          if (env.DAOS_STACK_LEAP_15_LOCAL_REPO) {
-            ret_str += ' --build-arg REPO_LOCAL_LEAP15=' +
-                       env.DAOS_STACK_LEAP_15_LOCAL_REPO
-          }
-          if (env.DAOS_STACK_LEAP_15_GROUP_REPO) {
+        repo_alias = 'LEAP_15'
+        repo_name = env["DAOS_STACK_${repo_alias}${repo_mod}_${repo_type}_REPO"]
+        if (repo_name) {
+          ret_str += ' --build-arg REPO_EL7=' + repo_name
+        }
```
brianjmurrell, line 124: Basically ditto for the above, but I doubt the --build-arg is supposed to be `REPO_EL7` here.

vars/dockerBuildArgs.groovy
```diff
@@ -48,27 +100,46 @@ String call(Map config = [:]) {
...
-        ret_str += ' --build-arg REPO_LOCAL_LEAP15=' +
-                   env.DAOS_STACK_LEAP_15_LOCAL_REPO
+      if (stage_info['target'] == 'leap15') {
+        repo_alias = 'LEAP_15'
+        repo_name = env["DAOS_STACK_${repo_alias}${repo_mod}_${repo_type}_REPO"]
+        if (repo_name) {
+          ret_str += ' --build-arg REPO_EL7=' + repo_name
+        }
+        if (repo_type == 'LOCAL') {
+         if (env.DAOS_STACK_LEAP_15_GROUP_REPO) {
```
brianjmurrell, line 126: Why is there (only) group repo support for Leap 15?

vars/dockerBuildArgs.groovy
```diff
@@ -48,27 +100,46 @@ String call(Map config = [:]) {
...
-      if (env.DAOS_STACK_UBUNTU_20_04_LOCAL_REPO) {
-        ret_str += ' --build-arg REPO_UBUNTU_20_04=' + env.DAOS_STACK_UBUNTU_20_04_LOCAL_REPO
+      if (stage_info['target'] == 'ubuntu20.04') {
+        repo_name = env.DAOS_STACK_UBUNTU_20_04_LOCAL_REPO
+        if (repo_type == 'LOCAL') {
+          repo_name = env.DAOS_STACK_UBUNTU_20_04_LOCAL_REPO
+          if (repo_name) {
+            ret_str += ' --build-arg REPO_UBUNTU_20_04=' + repo_name
+          }
+        }
```
brianjmurrell, line 139: This will get folded into the above more concise processing at some point?

vars/dockerBuildArgs.groovy
JohnMalmberg: Bug, will be fixed in next push.

vars/dockerBuildArgs.groovy
JohnMalmberg: We are going from having one local repo for locally built artifacts and a group repo for the OS artifacts to one group repo that has both.

By just using a group repo, it gets us around the issue that Nexus will not let us rename a local repo if we want save it and start using a different repo.

vars/dockerBuildArgs.groovy
JohnMalmberg: Will be done in the next push.   We are not using the UBUNTU local repo now, and we have not setup the infrastructure for proxying the other Ubuntu repositories yet.

vars/dockerBuildArgs.groovy
```diff
@@ -49,26 +101,43 @@ String call(Map config = [:]) {
       ret_str += " --build-arg CB0=" + current_time.get(Calendar.WEEK_OF_YEAR)
     }
 
-    if (add_repos) {
-      if (env.REPOSITORY_URL) {
-        ret_str += ' --build-arg REPO_URL=' + env.REPOSITORY_URL
-      }
-      if (env.DAOS_STACK_EL_7_LOCAL_REPO) {
-        ret_str += ' --build-arg REPO_EL7=' + env.DAOS_STACK_EL_7_LOCAL_REPO
+    String repo_base = env.REPOSITORY_URL
```
brianjmurrell, line 104: Is there a reason to access environment variables indirectly through local variable assignments?  I don't think there is any extra cost associated with accessing them directly and accessing them directly instead of through a local variable is more clear I think rather than having to remember (i.e. while reading) yet another variable's value.

vars/dockerBuildArgs.groovy
```diff
@@ -49,26 +101,43 @@ String call(Map config = [:]) {
...
+      if (stage_info['target'] == 'centos8') {
+        repo_alias = 'EL_8'
+        repo_arg = 'EL8'
       }
-      if (env.DAOS_STACK_LEAP_15_LOCAL_REPO) {
-        ret_str += ' --build-arg REPO_LOCAL_LEAP15=' +
-                   env.DAOS_STACK_LEAP_15_LOCAL_REPO
+      if (stage_info['target'] == 'leap15') {
+        repo_alias = 'LEAP_15'
+        repo_arg = 'LEAP'
```
brianjmurrell, line 119: Should we have the version on this one, like we do for EL?

vars/dockerBuildArgs.groovy
```diff
@@ -49,26 +101,43 @@ String call(Map config = [:]) {
...
+    if (add_repos && repo_base) {
+      String repo_name = null
+      String repo_arg = ''
+      if (stage_info['target'] == 'centos7') {
+        repo_alias = 'EL_7'
+        repo_arg = 'EL7'
       }
-      if (env.DAOS_STACK_EL_8_LOCAL_REPO) {
-        ret_str += ' --build-arg REPO_EL8=' + env.DAOS_STACK_EL_8_LOCAL_REPO
+      if (stage_info['target'] == 'centos8') {
```
brianjmurrell, line 113: Should use:
```
} else if (stage_info['target'] == 'centos8') {
```
here, and on all of the others below to prevent evaluating every branch when the matching one has already been found.

vars/dockerBuildArgs.groovy
```diff
@@ -49,26 +101,43 @@ String call(Map config = [:]) {
...
+        repo_alias = 'EL_8'
+        repo_arg = 'EL8'
       }
-      if (env.DAOS_STACK_LEAP_15_LOCAL_REPO) {
-        ret_str += ' --build-arg REPO_LOCAL_LEAP15=' +
-                   env.DAOS_STACK_LEAP_15_LOCAL_REPO
+      if (stage_info['target'] == 'leap15') {
```
brianjmurrell, line 117: Ditto.

vars/dockerBuildArgs.groovy
```diff
@@ -49,26 +101,43 @@ String call(Map config = [:]) {
...
+        if (repo_type == 'LOCAL') {
+         if (env.DAOS_STACK_LEAP_15_GROUP_REPO) {
+            ret_str += ' --build-arg REPO_GROUP_LEAP15=' +
+                       env.DAOS_STACK_LEAP_15_GROUP_REPO
+          }
+        }
       }
-      if (env.DAOS_STACK_UBUNTU_20_04_LOCAL_REPO) {
-        ret_str += ' --build-arg REPO_UBUNTU_20_04=' + env.DAOS_STACK_UBUNTU_20_04_LOCAL_REPO
+      if (stage_info['target'] == 'ubuntu20.04') {
```
brianjmurrell, line 128: Ditto.

vars/dockerBuildArgs.groovy
```diff
@@ -49,26 +101,43 @@ String call(Map config = [:]) {
...
+        repo_arg = repo_alias
       }
-      if (env.DAOS_STACK_LEAP_15_GROUP_REPO) {
-        ret_str += ' --build-arg REPO_GROUP_LEAP15=' +
-                   env.DAOS_STACK_LEAP_15_GROUP_REPO
+      repo_name = env["DAOS_STACK_${repo_alias}${repo_mod}_${repo_type}_REPO"]
+      // Only add the build args if a repo was found.
+      if (repo_name) {
+        ret_str += " --build-arg REPO_${repo_arg}=" + repo_name
+        ret_str += ' --build-arg REPO_URL=' + env.REPOSITORY_URL
```
brianjmurrell, line 140: Similar to the comment above about assigning `env.*` to local vars, you already have `repo_base` containing this value, so some inconsistency at least.  But as I said before, I find the direct use of `env.REPOSITORY_URL` here more transparent and less burdensome than having to keep track of/remember the local variable (while reading the code), for what I believe is about the same cost.

brianjmurrell (commented):
Looking much more maintainable.

Just a few comments/questions.  Most pressing is probably the issue of not using a version with Leap.  Wonder if that will bite us at some point as we transition from 15 to 16.

vars/dockerBuildArgs.groovy
JohnMalmberg: Missed that on the refactor.

vars/dockerBuildArgs.groovy
JohnMalmberg: Actually the tag just needs to match what is in the dockerfile.
For the leap dockerfile, I am refactoring down to one passed REPO from two, so can use a simpler repo_arg.

I need to double check that I did not break backwards compatibility in the last refactor though.

vars/dockerBuildArgs.groovy
JohnMalmberg: I was thinking that I needed to assign the env.REPOSITORY_URL at the top to a variable to make it simpler to test if it was not null.   That is obviously not needed.

Will make another push.

vars/dockerBuildArgs.groovy
brianjmurrell: Does the `repo_arg` need to even exist, or can we just use the same generic name in all of the Dockerfiles, since we are only passing the repos relevant to the distro?

vars/dockerBuildArgs.groovy
JohnMalmberg: Dockerfiles need to have the option of replacing the distro provided repos with local repos and they also need locally built artifact repos added to them.

VMs/Physical nodes at our site are built with the replacement for distro provided repos already so only the locally built artifact repos need to be added.

By using a artifact server with and a group repository, we go from needing to supply two repos to just supplying one, and we also gain a limited ability to change the repository on a PR basis.

vars/dockerBuildArgs.groovy
JohnMalmberg: The current dockerfiles all use different names for the repo_arg parameter.  We would have to pass both for this change to be compatible with Daos, which has not yet been patched to request using one of the new repos.

The older code passed all the repos to all the Dockerfiles, and resulted in log noise for the build-args that were not used for the dockerfile.

We may want to phase in switching to a single --build-args parameter in a future PR.

vars/dockerBuildArgs.groovy
brianjmurrell: > VMs/Physical nodes at our site are built with the replacement for distro provided repos already so only the locally built artifact repos need to be added.

Could Dockerfiles not mirror that?  Just seems like more work to keep separate CI image and Docker image repo definitions.

brianjmurrell (approved):

brianjmurrell:
I assume you have done a current master test of this PR to ensure it doesn't break the status quo.

JohnMalmberg:
Running a current master test now to get all the changes.  Should have results in the morning.

JohnMalmberg:
Got the same failures on the test pr build with master as that master build had.

PR #208, opened 2021-03-21
Move functions from daos Jenkinsfile

brianjmurrell:


vars/functionalPackages.groovy
```diff
@@ -0,0 +1,41 @@
...
+                  "romio-tests-daos-${client_ver} " +
+                  "testmpio " +
+                  "mpi4py-tests " +
+                  "hdf5-mpich-tests " +
+                  "hdf5-openmpi3-tests " +
+                  "hdf5-vol-daos-mpich-tests " +
+                  "hdf5-vol-daos-openmpi3-tests " +
+                  "MACSio-mpich " +
+                  "MACSio-openmpi3 " +
+                  "mpifileutils-mpich-daos-${client_ver} "
```
JohnMalmberg, line 30: Wouldn't we want this list to be stored in the DAOS repository?

Or use a tool to query the spec file?

vars/unitPackages.groovy
```diff
@@ -0,0 +1,41 @@
...
+                           'fuse3-libs fuse3 ' +
+                           'boost-devel ' +
+                           'libisa-l-devel libpmem ' +
+                           'libpmemobj protobuf-c ' +
+                           'spdk-devel libfabric-devel '+
+                           'pmix numactl-devel ' +
+                           'libipmctl-devel python36-pyxattr ' +
+                           'python36-tabulate numactl ' +
+                           'libyaml-devel ' +
+                           'valgrind-devel patchelf'
```
JohnMalmberg, line 27: Again, should this be provided by a file in the DAOS repo?

vars/unitPackages.groovy
```diff
@@ -0,0 +1,41 @@
...
+        if (env.STAGE_NAME.contains('Bullseye') ||
+            quickBuild()) {
+            unstash stage_info['target'] + '-required-mercury-rpm-version'
+            packages += " spdk-tools mercury-" +
+                        readFile(stage_info['target'] +
+                                 '-required-mercury-rpm-version').trim() +
+                        " boost-devel libisa-l_crypto libfabric-debuginfo" +
+                        " argobots-debuginfo protobuf-c-debuginfo"
```
JohnMalmberg, line 35: Ditto.

vars/functionalPackages.groovy
brianjmurrell: > Wouldn't we want this list to be stored in the DAOS repository?

Perhaps.  Let me see how it can be done usefully.
 
> Or use a tool to query the spec file?

Well, there is nothing to query,  Ultimately, this whole list should be dependencies of the _daos-tests_ RPM so that they are automatically installed when _daos-tests_ is installed.  There is a back-burnered PR open to do this.

brianjmurrell: Fixed in next push.

vars/unitPackages.groovy
brianjmurrell: Fixed in next push.

vars/unitPackages.groovy
brianjmurrell: Fixed in next push.

vars/daosPackagesVersion.groovy
```diff
@@ -0,0 +1,96 @@
...
+ */
+
+import groovy.transform.Field
+
+@Field static rpm_version_cache = ""
+
+String daos_latest_version(String next_version) {
+    String v = sh(label: "Get RPM packages version",
+                  script: 'repoquery --repofrompath=daos,' + env.REPOSITORY_URL +
+                           env.DAOS_STACK_EL_7_LOCAL_REPO + '''/ \
```
JohnMalmberg, line 20: See https://github.com/daos-stack/pipeline-lib/pull/207

For CENTOS, the new repos are:
DAOS-STACK-EL-7-STABLE-REPO  - Contains signed releases, will contain tagged builds in the future, and the DAOS_STACK_EL_7_LOCAL-REPO contents.

or

DAOS_STACK_EL_7_RELEASE_REPO - This only contains signed released RPMS

or

DAOS_STACK_EL_7_DEV-REPO - This is what ever we put in to if for special PR testing.

JohnMalmberg (changes requested):
Not sure if this is going to impact https://github.com/daos-stack/daos/pull/5116

vars/daosPackagesVersion.groovy
brianjmurrell: Is this still true after our discussion about having group repos with signed and unsigned packages in it?

JohnMalmberg: Yes for this case.

For yum repos, we only want to be reading from group repos in the artifacts server, not the local repos.

The DAOS-STACK-EL-7-STABLE-REPO and DAOS-STACK-EL-7-DEV-REPO will need all gpg checks disabled as they are all locally built content.

The DAOS-STACK-EL-7-RELEASE-REPO  can have package gpg-checks enabled, but not repo gpg-checks enabled since our artifact server does not support repo gpg-checks for yum repos.

brianjmurrell: So then, given that I have no clue what repos are what any more, what should I be using here?

vars/daosPackagesVersion.groovy
```diff
@@ -0,0 +1,96 @@
...
+ */
+
+import groovy.transform.Field
+
+@Field static rpm_version_cache = ""
+
+String daos_latest_version(String next_version) {
+    String v = sh(label: "Get RPM packages version",
+                  script: 'repoquery --repofrompath=daos,' + env.REPOSITORY_URL +
+                           env.DAOS-STACK-EL-7-STABLE-REPO + '''/ \
```
JohnMalmberg, line 20: `
String daos_latest_version(String next_version, String repo_type='stable`) {
    String repo_type_u = repo_type.toString().toUpperCase()
    String repo_url = env.REPOSITORY_URL + env["DAOS_STACK_${repo_type_u}_GROUP_REPO"]
    String v = sh(label: "Get RPM packages version",
                  script: 'repoquery --repofrompath=daos,' + repo_url +'''/ \

vars/daosPackagesVersion.groovy
```diff
@@ -0,0 +1,96 @@
...
+                                     --repoid daos --qf %{version}-%{release} --whatprovides 'daos(x86-64) < ''' +
+                                     next_version + '''' |
+                                 rpmdev-sort | tail -1''',
+                  returnStdout: true).trim()
```
JohnMalmberg, line 24: Does this fix the issue that I am having with https://github.com/daos-stack/daos/pull/5116 that forced me to use yum by removing the trailing newline?

JohnMalmberg (changes requested):

vars/daosPackagesVersion.groovy
brianjmurrell: I don't know.  I have not seen what the issue looks like or how it manifests itself.  I really need to see a run with it exhibiting the bad behaviour to know.

vars/daosPackagesVersion.groovy
```diff
@@ -0,0 +1,96 @@
...
+ */
+
+import groovy.transform.Field
+
+@Field static rpm_version_cache = ""
+
+String daos_latest_version(String next_version, String repo_type='stable') {
+    String v = sh(label: "Get RPM packages version",
+                  script: 'repoquery --repofrompath=daos,' + env.REPOSITORY_URL +
+                          env["DAOS-STACK-EL-7-${repo_type}-REPO"].toUpperCase() +
```
JohnMalmberg, line 20: You are uppercasing the returned URL, it is the environment variable name that needs to be in upper case.

JohnMalmberg (changes requested):

vars/daosPackagesVersion.groovy
brianjmurrell: I don't think so.  The `toUpperCase()` should only being applied to `env["DAOS-STACK-EL-7-${repo_type}-REPO"]`:

```groovy
String foo="foo"
String bar="bar"
println(foo + bar.toUpperCase())
```
results in:

```
fooBAR
```

As you can see, the `toUpperCase()` method is applied to the most direct string, not the composed string.  IOW, the `toUpperCase()` method has higher precedence than the `+`.

JohnMalmberg (approved):

PR #216, opened 2021-04-11
Add "Functional Test With Valgrind" stage

mallove79:
This PR is part of CART-952 (_Re-enable CART tests to run under valgrind_).

Signed-off-by: Ethan Mallove <ethanx.a.mallove@intel.com>


vars/skipStage.groovy
```diff
@@ -22,6 +22,15 @@ boolean skip_ftest(String distro) {
            skip_stage_pragma('func-test-' + distro)
 }
 
+boolean skip_ftest_valgrind(String distro) {
+    return distro == 'ubuntu20' ||
+           skip_stage_pragma('func-test') ||
+           skip_stage_pragma('func-test-vm') ||
+           skip_stage_pragma('func-test-vm-valgrind') ||
+           ! testsInStage() ||
+           skip_stage_pragma('func-test-' + distro)
```
brianjmurrell, line 31: I know you have not requested reviews yet, so you still have the right to force-push to this PR if you want.  Just thought I'd make a suggestion though.  You can call the function you copied instead of copying all of it's conditions (and creating a future situation where one gets updated with a new condition and the other doesn't):

```suggestion
    return skip_ftest(distro) ||
           skip_stage_pragma('func-test-vm-valgrind')
```

brianjmurrell (changes requested):
Please see my comment about not duplicating the conditions of `skip_ftest()` in another stage.

Also, I don't think this PR is complete with regard to your goals of only running cart tests in the new proposed daos testing stage.

I'd like to see the daos PR that needs these changes complete and ready to land, referencing this PR as it's pipeline-lib before we call this complete.

mallove79:
> Please see my comment about not duplicating the conditions of `skip_ftest()` in another stage.

Thank you for the suggestion!  I applied [your patch](https://github.com/daos-stack/pipeline-lib/pull/216/commits/429ceffb5bddd99c67143e897bc02032545431b3). 

> Also, I don't think this PR is complete with regard to your goals of only running cart tests in the new proposed daos testing stage.

True, `func-test-vm-valgrind` currently just sets `CART_TEST_MODE=memcheck`.  I've changed it so [`func-test-vm-valgrind` also sets `config['test_tag']="cart"`](https://github.com/daos-stack/daos/pull/4712/files#diff-e6ffa5dc854b843b3ee3c3c28f8eae2f436c2df2b1ca299cca1fa5982e390cf8R897).

> I'd like to see the daos PR that needs these changes complete and ready to land, referencing this PR as it's pipeline-lib before we call this complete. Can you put a link to that PR in this one so that we can track the changes together?

https://github.com/daos-stack/daos/pull/4712

brianjmurrell:
> True, `func-test-vm-valgrind` currently just sets `CART_TEST_MODE=memcheck`. I've changed it so [`func-test-vm-valgrind` also sets `config['test_tag']="cart"`](https://github.com/daos-stack/daos/pull/4712/files#diff-e6ffa5dc854b843b3ee3c3c28f8eae2f436c2df2b1ca299cca1fa5982e390cf8R897).

Hrm.  I'm on the fence if this is the right place or if it more correctly belongs in [https://github.com/daos-stack/pipeline-lib/blob/429ceffb5bddd99c67143e897bc02032545431b3/vars/parseStageInfo.groovy](parseStageInfo.groovy).

Is only running tests tagged with `cart` the end goal here or will this be expanded to running all tests in this mode?

vars/skipStage.groovy
brianjmurrell: Please leave comments for the comment creator to resolve, which they should/could do once they have verified that the resolution applied satisfies their concern.

mallove79:
> > True, `func-test-vm-valgrind` currently just sets `CART_TEST_MODE=memcheck`. I've changed it so [`func-test-vm-valgrind` also sets `config['test_tag']="cart"`](https://github.com/daos-stack/daos/pull/4712/files#diff-e6ffa5dc854b843b3ee3c3c28f8eae2f436c2df2b1ca299cca1fa5982e390cf8R897).
> 
> Hrm. I'm on the fence if this is the right place or if it more correctly belongs in [parseStageInfo.groovy](https://github.com/daos-stack/pipeline-lib/blob/429ceffb5bddd99c67143e897bc02032545431b3/vars/parseStageInfo.groovy).
> 
> Is only running tests tagged with `cart` the end goal here or will this be expanded to running all tests in this mode?

There's no immediate plan to run tests like `valgrind daos_{engine,agent,server}` (I believe we already valgrind unit tests in NLT)

mallove79:
> > There's no immediate plan to run tests like `valgrind daos_{engine,agent,server}` (I believe we already have valgrind unit tests [in NLT](https://github.com/daos-stack/daos/blob/master/utils/node_local_test.py).)
> 
> That makes me lean towards putting the tag setting into [parseStageInfo.groovy](https://github.com/daos-stack/pipeline-lib/blob/429ceffb5bddd99c67143e897bc02032545431b3/vars/parseStageInfo.groovy) then, as it's the default setting for that stage.

Sounds good: https://github.com/daos-stack/pipeline-lib/pull/216/commits/482fb6654937c47e19752040f74f977ef155bf4d.

vars/parseStageInfo.groovy
```diff
@@ -144,6 +144,10 @@ def call(Map config = [:]) {
         result['pragma_suffix'] = '-hw-medium'
       }
     }
+    if (stage_name.contains('with Valgrind')) {
+      result['ftest_arg'] = ''
```
brianjmurrell, line 148: Why are you overriding `result['ftest_arg']`?  In what situation is it's value wrong in the context of adding valgrind to (some) functional testing?

vars/parseStageInfo.groovy
```diff
@@ -144,6 +144,10 @@ def call(Map config = [:]) {
         result['pragma_suffix'] = '-hw-medium'
       }
     }
+    if (stage_name.contains('with Valgrind')) {
```
brianjmurrell, line 147: Putting this here feels like it would have (too) broad coverage.  If somebody were to add some other stage that simply had "with Valgrind" in it, these overrides would affect that stage also.  I think you want to target this check to a more specific stage name, possibly using the block above https://github.com/daos-stack/pipeline-lib/blob/f5cb9cc8a6561b2f871ecf7de209611f176c8b7c/vars/parseStageInfo.groovy#L127-L146

vars/parseStageInfo.groovy
mallove79: `ftest_arg` shouldn't be overwritten.  d704aad6672018107ebf92ba38a1bebff477dbda

vars/parseStageInfo.groovy
mallove79: I moved it into the above block, so we set `WITH_VALGRIND=memcheck` for a stage name match on `Functional` and `with Valgrind`.

vars/runTestFunctionalV2.groovy
```diff
@@ -53,6 +55,7 @@ void call(Map config = [:]) {
                        'PRAGMA_SUFFIX="' + config['pragma_suffix'] + '" ' +
                        'NODE_COUNT="' + config['node_count'] + '" ' +
                        'OPERATIONS_EMAIL="' + env.OPERATIONS_EMAIL + '" ' +
+                       "WITH_VALGRIND=${stage_info.get('with_valgrind', '')} " +
```
brianjmurrell, line 58: I don't see where this is consumed in https://github.com/daos-stack/daos/blob/0e62a8ade4bc4e7af080f867b6191c8951ff428c/src/tests/ftest/scripts/main.sh.  Rather, I see it's value being overridden:
```
if [[ "${TEST_TAG_ARG}" =~ memcheck ]]; then
  export WITH_VALGRIND="memcheck"
fi
```
by using the test tag value as a proxy for just using the passed in value of this variable.  It seems that using the value of this variable, as passed in is more accurate than overriding it based on test tag values.

brianjmurrell (approved):
Based on fixing the requested change in the corresponding daos PR, this looks good.

vars/runTestFunctionalV2.groovy
mallove79: That `WITH_VALGRIND` setting is from an earlier experiment, so I removed it. Good catch. 👍

mallove79:
Are these changes breaking [Build #60](https://build.hpdd.intel.com/blue/organizations/jenkins/daos-stack%2Fdaos/detail/PR-4712/62/pipeline)?  I can't figure out why it only executes [two `avocado run`'s](https://build.hpdd.intel.com/job/daos-stack/job/daos/view/change-requests/job/PR-4712/62/testReport/).

phender:
> Are these changes breaking [Build #60](https://build.hpdd.intel.com/blue/organizations/jenkins/daos-stack%2Fdaos/detail/PR-4712/62/pipeline)? I can't figure out why it only executes [two `avocado run`'s](https://build.hpdd.intel.com/job/daos-stack/job/daos/view/change-requests/job/PR-4712/62/testReport/).

Basically, due to the issue noted in https://github.com/daos-stack/daos/pull/4712#discussion_r647740548, the execution of launch.py is prematurely terminated due to this error:
```
[2021-06-07T16:24:54.093Z] mkdir: cannot create directory ‘/mnt/share/tmp.6yYaFyA3ky/latest/valgrind_logs’: No such file or directory
```
The fact that launch.py does not handle this type of error gracefully is already ticketed, but using `avocado_logs_dir` instead of `valgrind_logs_dir` in the `archive_files()` call should resolve the problem.

sylviachanoiyee (approved):

brianjmurrell (approved):

PR #223, opened 2021-05-07
DAOS-623 build: Change priority of landings on master back to default.

ashleypittman:
This will prevent landings on master consuming test cycles
ahead of development.wq

### DIFF
--- a/vars/getPriority.groovy
+++ b/vars/getPriority.groovy
@@ -8,7 +8,7 @@
    */
 
 def call() {
-      if (env.BRANCH_NAME == 'master' ||
+      if (startedByTimer() ||
         env.BRANCH_NAME.startsWith("release/") ||
         env.BRANCH_NAME == 'weekly-testing') {
         string p = '2'


brianjmurrell:
> Is there any way we can test this hypothesis?

None that I can think of, that doesn't include replicating the entire environment to run a test on.  But I'm not sure we need to test it.  It's very well known behaviour.  It's documented and we have see it ourselves.

ashleypittman:
> > Is there any way we can test this hypothesis?
> 
> None that I can think of, that doesn't include replicating the entire environment to run a test on. But I'm not sure we need to test it. It's very well known behaviour. It's documented and we have see it ourselves.

What I saw was it using the code from the previous run to set the priority for the next run, so if you changed the value from 2 to 4 in a PR that PR would still run at 2, but with this the code would be the same, so even if it ran the code from the previous run it would still do the wrong thing.  It could go either way but I've not seen evidence that it wouldn't work, only that we need to verify it.

PR #227, opened 2021-05-20
CORCI-1081 ci: Changes to increase CI throughput

brianjmurrell:
Change PRs to only run CentOS7 on VMs.

Landings and timed builds are moved down to the same priority as PRs.

Run build and test stages on an empty commit message with PR-repos* set.
Because there are no code changes, these commits will pass the doc-only
test and not run build and test, but the PR-repos* pragmas tell us that
this is a commit being used to test updates to a dependency build.

Skip UT stage on Quick-build (which is implied with Quick-Functional).

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -15,7 +15,7 @@
 // That PR should be landed with out deleting the PR branch.
 // Then a second PR submitted to comment out the @Library line, and when it
 // is landed, both PR branches can be deleted.
-//@Library(value="pipeline-lib@my_branch_name") _
+@Library(value="pipeline-lib@bmurrell/shuffle-builds") _
 
 pipeline {
     agent { label 'lightweight' }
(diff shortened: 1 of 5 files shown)


brianjmurrell:
https://build.hpdd.intel.com/blue/organizations/jenkins/daos-stack%2Fpipeline-lib/detail/PR-227/53/pipeline is a run using this PR as the pipeline-lib branch, demonstrating the new tests pass.

ashleypittman:
> https://build.hpdd.intel.com/blue/organizations/jenkins/daos-stack%2Fdaos/detail/PR-5758/16/pipeline/ shows a test run with `Skip-func-test-leap15: false` forcing the normally skipped (with this PR) Leap15 stage being run.
> 
> https://build.hpdd.intel.com/blue/organizations/jenkins/daos-stack%2Fdaos/detail/PR-5758/17/pipeline shows a test run that represents the new default behaviour with this PR.

Are those links correct?

ashleypittman:
> > Are those links correct?
> 
> Yes.

The second link shows a build that only ran Functional on CentOS 7, and nothing else.  Even if you mean it's showing what that stage will do rather than the whole pipeline then I'd expect to see Coverity and the two scan/test rpm job steps running.

brianjmurrell:
> The second link shows a build that only ran Functional on CentOS 7, and nothing else. Even if you mean it's showing what that stage will do rather than the whole pipeline then I'd expect to see Coverity and the two scan/test rpm job steps running.

It's running _Quick-Functional: true_ which skips those stages.

ashleypittman (approved):

brianjmurrell:
Here's a test run without _Quick-Functional: true_: https://build.hpdd.intel.com/blue/organizations/jenkins/daos-stack%2Fdaos/detail/PR-5758/18/pipeline.  It looks like the _{Scan,Test} CentOS 7 RPMs_ stages don't know how to deal with daos RPMs that are not in the PR but rather in a `PR-repos` repo.  We can see however that their stages as well as the _Coverity_ stage ran and the EL7 VM stage is (at the time of writing queued to run,) just waiting on a cluster.

ashleypittman:
Let's land this then and see where we are at that point.  Looking at the build queues today we're back to having long waits for vm9 and this should resolve that.

PR #232, opened 2021-06-14
DAOS-7817 test: Expand release branch PR testlist

sylviachanoiyee:
Expand release branch PR testing to include daily_regression

### DIFF
--- a/vars/parseStageInfo.groovy
+++ b/vars/parseStageInfo.groovy
@@ -195,6 +195,12 @@ def call(Map config = [:]) {
       }
     }
 
+    String target_branch = env.CHANGE_TARGET ? env.CHANGE_TARGET : env.BRANCH_NAME
+    if (target_branch == "release/1.2") {
+      echo "Updating tag to inclue daily_regression for release/1.2"
+      echo "Tag is ${tag}"
+      tag += ' daily_regression'
+    }
     result['test_tag'] = ""
     for (atag in tag.split(' ')) {
       result['test_tag'] += atag + ',' + cluster_size + ' '
--- a/vars/parseStageInfo.groovy
+++ b/vars/parseStageInfo.groovy
@@ -177,6 +177,12 @@ def call(Map config = [:]) {
               } else {
                 // Must be a PR run
                 tag = "pr"
+
+                String target_branch = env.CHANGE_TARGET ? env.CHANGE_TARGET : env.BRANCH_NAME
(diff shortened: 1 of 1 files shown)


vars/parseStageInfo.groovy
```diff
@@ -195,6 +195,12 @@ def call(Map config = [:]) {
       }
     }
 
+    String target_branch = env.CHANGE_TARGET ? env.CHANGE_TARGET : env.BRANCH_NAME
+    if (target_branch == "release/1.2") {
+      echo "Updating tag to inclue daily_regression for release/1.2"
+      echo "Tag is ${tag}"
+      tag += ' daily_regression'
+    }
```
brianjmurrell, line 203: Doing this here makes the addition absolute, even in cases where the user sets a Test-tag*.  That is not likely what we want.

It seems the goal here is to add `daily_regression` to PRs when normally it would have only been `pr`.  To achieve that, you want this addition to be done at https://github.com/daos-stack/pipeline-lib/blob/3f66aec1d9b61d96874ef6fcb6c70eceb18d096f/vars/parseStageInfo.groovy#L179.

brianjmurrell (changes requested):

vars/parseStageInfo.groovy
```diff
@@ -177,6 +177,12 @@ def call(Map config = [:]) {
               } else {
                 // Must be a PR run
                 tag = "pr"
+
+                String target_branch = env.CHANGE_TARGET ? env.CHANGE_TARGET : env.BRANCH_NAME
+                if (target_branch == "release/1.2") {
+                  echo "Updating tag to inclue daily_regression for release/1.2"
+                  tag += " daily_regression"
+                }
```
brianjmurrell, line 185: I wonder if this should be more generalized and driven as a _property_ of the branch's `Jenkinsfile`.  You could imagine that at any point in time that we could have a number of branches in various levels of activity and this change is really a reflection of the `release/1.2` branch's (low) activity level at the moment.  But at some point in the future, say when a 1.2.1 is being prepared the activity level could go back up and switching it back to pr with nightly runs might again become appropriate.  Other branches could be changing their activity levels at different times also.

We really shouldn't be flapping this repo to toggle such properties of a branch.

Maybe we need to define some activity levels and make generic changes here to reflect what we want happening at the different activity levels and then PRs to the branches can change that branch's activity level as needed.

PR #233, opened 2021-06-14
CORCI-1074 build: Allow special CI targets

JohnMalmberg:
Separate out a 'ci_target' from a build 'target' to allow more easily
running tests on newer pre-built CI images.

dockerBuildArgs.groovy:
   Add CentOS 8 special app stream repo
   Fix missing variable declaration

parseStageInfo.groovy
   Set the 'ci_target' member for node provisioning

rpmTestVersion.goovy
   Allow a build parameter for the version

skipStage.groovy
   Allow build paramters to select tests to skip.

functionalTest.groovy:
scanRpms.groovy:
testRpm.groovy:
unitTest.groovy
   Now use 'ci_target' for node provisioning

provisionNodes.groovy:
   Fix some declarations to be more specific types
   Remove some dead code

storagePrepTest.groovy:
   Stub for doing a Daos Storage Prep

### DIFF
--- a/vars/daosPackagesVersion.groovy
+++ b/vars/daosPackagesVersion.groovy
@@ -29,7 +29,7 @@ String daos_latest_version(String next_version, String repo_type='stable') {
 String rpm_dist(String distro) {
     if (distro.startsWith('el7') || distro.startsWith('centos7')) {
         return ".el7"
-    if (distro.startsWith('el8') || distro.startsWith('centos8')) {
+    } else if (distro.startsWith('el8') || distro.startsWith('centos8')) {
         return ".el8"
     } else if (distro.startsWith("leap15")) {
         return ".suse.lp152"
--- a/vars/daosPackagesVersion.groovy
+++ b/vars/daosPackagesVersion.groovy
@@ -29,6 +29,8 @@ String daos_latest_version(String next_version, String repo_type='stable') {
 String rpm_dist(String distro) {
     if (distro.startsWith('el7') || distro.startsWith('centos7')) {
         return ".el7"
+    if (distro.startsWith('el8') || distro.startsWith('centos8')) {
+        return ".el8"
     } else if (distro.startsWith("leap15")) {
         return ".suse.lp152"
     } else {
--- a/vars/daosPackagesVersion.groovy
+++ b/vars/daosPackagesVersion.groovy
@@ -18,7 +18,7 @@ String daos_latest_version(String next_version, String repo_type='stable') {
     String v = sh(label: "Get RPM packages version",
                   script: 'repoquery --repofrompath=daos,' + env.REPOSITORY_URL +
                           env["DAOS_STACK_EL_7_" + repo_type.toUpperCase() + "_REPO"] +
-                        '''/ --repoid daos --qf %{version}-%{release} --whatprovides 'daos(x86-64) < ''' +
+                        '''/ --repoid daos --qf %{version}-%{release} --whatprovides 'daos-tests(x86-64) < ''' +
                                      next_version + '''' |
                                  rpmdev-sort | tail -1''',
                   returnStdout: true).trim()
--- a/vars/getDAOSPackages.groovy
+++ b/vars/getDAOSPackages.groovy
@@ -24,5 +24,6 @@ String call(String distro, String next_version) {
     if (distro.startsWith('ubuntu20')) {
         return pkgs + "=" + daosPackagesVersion(distro, next_version)
     }
+    String ret_str = pkgs + "-" + daosPackagesVersion(distro, next_version)
     return pkgs + "-" + daosPackagesVersion(distro, next_version)
 }
--- a/vars/getDAOSPackages.groovy
+++ b/vars/getDAOSPackages.groovy
@@ -24,6 +24,5 @@ String call(String distro, String next_version) {
     if (distro.startsWith('ubuntu20')) {
         return pkgs + "=" + daosPackagesVersion(distro, next_version)
     }
-    String ret_str = pkgs + "-" + daosPackagesVersion(distro, next_version)
     return pkgs + "-" + daosPackagesVersion(distro, next_version)
 }
--- a/vars/parseStageInfo.groovy
+++ b/vars/parseStageInfo.groovy
@@ -64,16 +64,24 @@ def call(Map config = [:]) {
       echo "Could not determine target in ${env.STAGE_NAME}, defaulting to EL7"
     }
   }
+  String new_ci_target = params['CI_' +
+                                result['target'].toString().toUpperCase() +
+                                '_TARGET']
+  if (new_ci_target) {
+    result['ci_target'] = new_ci_target
+  } else {
+    result['ci_target'] = result['target']
+  }
 
-  if (result['target'].startsWith('el') ||
-      result['target'].startsWith('centos')) {
+  if (result['ci_target'].startsWith('el') ||
+      result['ci_target'].startsWith('centos')) {
     result['java_pkg'] = 'java-1.8.0-openjdk'
-  } else if (result['target'].startsWith('ubuntu')) {
+  } else if (result['ci_target'].startsWith('ubuntu')) {
     result['java_pkg'] = 'openjdk-8-jdk'
-  } else if (result['target'].startsWith('leap')) {
+  } else if (result['ci_target'].startsWith('leap')) {
     result['java_pkg'] = 'java-1_8_0-openjdk'
   } else {
-    error 'Java package not known for ' + result['target']
+    error 'Java package not known for ' + result['ci_target']
   }
 
   result['compiler'] = 'gcc'
@@ -227,7 +235,13 @@ def call(Map config = [:]) {
       }
     }
 
-  } // if (stage_name.contains('Functional'))
+    // if (stage_name.contains('Functional'))
+  } else if (stage_name.contains('Storage')) {
+    if (env.NODELIST) {
+      List node_list = env.NODELIST.split(',')
+      result['node_count'] = node_list.size()
+    }
+  } // else if (stage_name.contains('Storage'))
   if (config['test']) {
     result['test'] = config['test']
   }
--- a/vars/rpmTestVersion.groovy
+++ b/vars/rpmTestVersion.groovy
@@ -11,7 +11,7 @@
  */
 
 boolean call() {
-    if (params.CI_NOBUILD) {
+    if (params.CI_RPM_TEST_VERSION) {
         return params.CI_RPM_TEST_VERSION
     }
     return cachedCommitPragma('RPM-test-version')
--- a/vars/rpmTestVersion.groovy
+++ b/vars/rpmTestVersion.groovy
@@ -11,5 +11,8 @@
  */
 
 boolean call() {
+    if (params.CI_NOBUILD) {
+        return params.CI_RPM_TEST_VERSION
+    }
     return cachedCommitPragma('RPM-test-version')
 }
--- a/vars/skipStage.groovy
+++ b/vars/skipStage.groovy
@@ -22,15 +22,27 @@ String run_default_skipped_stage(String stage) {
 }
 
 boolean is_pr() {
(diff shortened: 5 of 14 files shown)


vars/rpmTestVersion.groovy
```diff
@@ -11,5 +11,8 @@
  */
 
 boolean call() {
+    if (params.CI_NOBUILD) {
+        return params.CI_RPM_TEST_VERSION
+    }
```
brianjmurrell, line 16: With commit pragmas, `RPM-test-version:` implies skipping building.  Parameters should probably have the same implicit skip without people having to specify it.

vars/skipStage.groovy
```diff
@@ -52,7 +70,13 @@ boolean skip_ftest_hw(String size, String target_branch) {
 }
 
 boolean skip_if_unstable() {
-    if (cachedCommitPragma('Allow-unstable-test').toLowerCase() == 'true' ||
+    if (params.CI_ALLOW_UNSTABLE_TEST) {
+        println('skip if unstable - params.CI_ALLOW_UNSTABLE_TEST is true')
+    } else {
+        println('skip if unstable - params.CI_ALLOW_UNSTABLE_TEST is true')
+    }
```
brianjmurrell, line 77: Debug that needs removing?

vars/skipStage.groovy
```diff
@@ -96,77 +121,102 @@ boolean call(Map config = [:]) {
...
+        case "Build RPM on CentOS 7":
+            return params.CI_RPM_centos7_NOBUILD ||
+                   (docOnlyChange(target_branch) &&
+                    prRepos('centos7') == '') ||
+                   skip_stage_pragma('build-centos7-rpm')
+        case "Build RPM on CentOS 8":
+            return params.CI_RPM_centos8_NOBUILD ||
+                   (docOnlyChange(target_branch) &&
+                    prRepos('centos8') == '') ||
+                   skip_stage_pragma('build-centos7-rpm')
```
brianjmurrell, line 137:
```suggestion
                   skip_stage_pragma('build-centos8-rpm')
```

vars/skipStage.groovy
```diff
@@ -178,9 +228,20 @@ boolean call(Map config = [:]) {
...
-            return skip_stage_pragma('unit-test') ||
+                   if (params.CI_UNIT_TEST) {
+                       println("CI_UNIT_TEST found!")
+                   }
+                   if (skip_stage_pragma('unit-test')) {
+                       println("Unit test skip requested")
+                   }
+                   if (skip_stage_pragma('run-test')) {
+                       println("Run test skip requested")
+                   }
```
brianjmurrell, line 242: Debug to be removed?

brianjmurrell (changes requested):

vars/getDAOSPackages.groovy
```diff
@@ -24,5 +24,6 @@ String call(String distro, String next_version) {
     if (distro.startsWith('ubuntu20')) {
         return pkgs + "=" + daosPackagesVersion(distro, next_version)
     }
+    String ret_str = pkgs + "-" + daosPackagesVersion(distro, next_version)
```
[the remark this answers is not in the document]
JohnMalmberg: Will be fixed in the next push.

vars/rpmTestVersion.groovy
JohnMalmberg: I want the RPM test version parameter to be defaulted to the close to the current release or test build version.

This makes it easy to have a separate job for running special builds or doing storage preparation on new or rebuilt CI nodes.

And having that value means that I can not just test for params.CI_RPM_TEST to have a text value.

vars/skipStage.groovy
JohnMalmberg: Yes.

vars/skipStage.groovy
JohnMalmberg: Will be fixed in the next push.

vars/skipStage.groovy
JohnMalmberg: Yes.  On my previous test run, this stage was skipped with out anything in the log to indicate why.

vars/rpmTestVersion.groovy
brianjmurrell: I think it's going to be confusing (and therefore bad) to have two different interfaces for doing the same thing that don't operate in the same way.  We already get enough of a beating for things not being intuitive enough in CI, that having two interfaces for doing the same thing operating on different assumptions will look bad on us.

JohnMalmberg:
https://build.hpdd.intel.com/job/daos-stack/job/daos/view/change-requests/job/PR-5622/ run with out the parameters in the Jenkins file and using commit pragmas.   This test did what it was supposed to.

vars/rpmTestVersion.groovy
JohnMalmberg: Fixed in current push.

JohnMalmberg:
https://build.hpdd.intel.com/job/daos-stack/job/daos/view/change-requests/job/PR-5622/158/ is a test run with DAOS that is not setting the parameter values and using commit pragmas.

https://build.hpdd.intel.com/job/daos-stack/job/daos/view/change-requests/job/PR-5622/157/ is a test run with DAOS that is setting the parameter values to specify a prebuilt RPM.  This shows a failure because no valgrind data was generated since the Unit tests were skipped.

vars/parseStageInfo.groovy
```diff
@@ -64,16 +64,24 @@ def call(Map config = [:]) {
       echo "Could not determine target in ${env.STAGE_NAME}, defaulting to EL7"
     }
   }
+  String new_ci_target = params['CI_' +
+                                result['target'].toString().toUpperCase() +
+                                '_TARGET']
+  if (new_ci_target) {
+    result['ci_target'] = new_ci_target
```
brianjmurrell, line 71: Line 12 still says `result` has a `target` member but doesn't mention `ci_target` member.  Is `ci_target` replacing/obsoleting `target`?  If so, why not re-use it rather than adding additional unnecessary churn to this PR and future confusion about `target` vs. `ci_target`?

vars/skipStage.groovy
```diff
@@ -22,15 +22,33 @@ String run_default_skipped_stage(String stage) {
 }
 
 boolean is_pr() {
+    if (params.CI_MORE_FUNCTIONAL_PR_TESTS) {
```
brianjmurrell, line 25: https://github.com/daos-stack/daos/blob/a49bf5180d0d3d865acbd9d2ad2bc68331a0431c/Jenkinsfile#L128-L130 is unclear to me as to what that parameter is for, why I'd use it, what I'd put in it.

Which makes me wonder why the parameters are not just mirrors (in groovy name, description, etc.) of commit pragmas.

vars/rpmTestVersion.groovy
```diff
@@ -11,5 +11,8 @@
  */
 
 boolean call() {
+    if (params.CI_RPM_TEST_VERSION) {
+        return params.CI_RPM_TEST_VERSION
+    }
```
brianjmurrell, line 16: This doesn't seem to agree with https://github.com/daos-stack/pipeline-lib/blob/27ccb295de9ce63fd80a8752621220bde6974678/vars/skipStage.groovy#L43.

But I'm also not sure that pragmas should have the highest priority. It has been expressed by others that they were surprised that the existing parameters we have did not override commit pragmas.

vars/skipStage.groovy
```diff
@@ -200,7 +248,9 @@ boolean call(Map config = [:]) {
             return skip_stage_pragma('vagrant-test', 'true') &&
                    ! env.BRANCH_NAME.startsWith('weekly-testing')
         case "Coverity on CentOS 7":
-            return skip_stage_pragma('coverity-test') ||
+            return params_value('CI_BUILD_PACKAGES_ONLY', false) ||
+                   params.CI_RPM_TEST_VERSION ||
```
brianjmurrell, line 252: If this is going to be a skip condition, shouldn't `rpmTestVersion() != ''` also be added?

vars/storagePrepTest.groovy
```diff
@@ -0,0 +1,97 @@
...
+
+  if (!fileExists('ci/storage/test_main.sh')) {
+    println("No storage Prep script found!")
+    return
+  }
+
+  Boolean test_rpms = false
+  if (config['test_rpms'] == "true") {
+    test_rpms = true
+  }
```
brianjmurrell, line 83: Is this needed for `ci/storage/test_main.sh`?

vars/rpmTestVersion.groovy
JohnMalmberg: Yeah, that is inconsistent.  Would prefer to make it consistent in a later update if possible.

vars/skipStage.groovy
JohnMalmberg: Currently some behavior of the skips are based on if this is a pr branch or not.

This parameter is used to disable the "is_pr() test so that I can use parameters to enable tests that would normally be skipped in a PR.

It is a boolean parameter.   Unfortunately the help text in the jenkinsfile is not displayed for boolean parameters.

vars/skipStage.groovy
JohnMalmberg: I did not realize that was not already covered.

vars/storagePrepTest.groovy
JohnMalmberg: I am not sure if it is.

vars/rpmTestVersion.groovy
brianjmurrell: How much later?  I just don't want anyone stumbling on to the inconsistency and complaining.  Let's at least have a ticket we can point them at.

But to be clear, I think parameters (where specified, and commit pragams should be honoured where a parameter is not specified) should override commit pragmas as the PoR.

vars/storagePrepTest.groovy
brianjmurrell: I suspect it's not.

vars/skipStage.groovy
brianjmurrell: Ultimately, can coverity be used when doing a run from RPMs built from a previous run?  If not, then this should be added in addition to your parameter addition.  If it can, then your parameter check here shouldn't be here.

JohnMalmberg: Coverity is actually a quickbuild build stage that will build the git commit for the Jenkins job if it is allowed to run.

So any time that you are skipping a normal build, or specifying a pre-built RPM, a coverity build is not wanted or is not going to build what was expected.

The `rpmTestversion() != ''` includes a test for param.RPM_TEST_VERSION so both are not needed in the same conditional and will be removed in the next commit.

vars/storagePrepTest.groovy
JohnMalmberg: Noop code will be removed in the next push.

vars/rpmTestVersion.groovy
JohnMalmberg: Updated comment for skip_ftest.  Effectively the checks need to be done for pragmas to enabled skipped functional tests before checks are done for parameters to enable skipped functional checks.

ashleypittman (approved):

vars/daosPackagesVersion.groovy
```diff
@@ -18,7 +18,7 @@ String daos_latest_version(String next_version, String repo_type='stable') {
     String v = sh(label: "Get RPM packages version",
                   script: 'repoquery --repofrompath=daos,' + env.REPOSITORY_URL +
                           env["DAOS_STACK_EL_7_" + repo_type.toUpperCase() + "_REPO"] +
-                        '''/ --repoid daos --qf %{version}-%{release} --whatprovides 'daos(x86-64) < ''' +
+                        '''/ --repoid daos --qf %{version}-%{release} --whatprovides 'daos-tests(x86-64) < ''' +
```
brianjmurrell, line 21: Just curious why this needed changing.  My only concern is that the `daos` package is probably a more stable name than `daos-tests`.  The latter for example could be ripe to be completely replaced by a bunch of `daos-tests-$subpackages` and then this function will fail due to a missing `daos-tests` package.

PR #238, opened 2021-08-18
DAOS-8161 tests: Enabling VMD testing in CI HW clusters

phender:
Changing the default launch.py --nvme argument for CI HW clusters to
enable testing with both non-VMD NVMe drives and VMD NVMe drives.


phender:
Note this cannot land until a daos-stack/daos PR has been run with `@Library(value="pipeline-lib@PR-238")` in the `Jenkinsfile` to ensure compatibility with existing CI clusters.  An additional PR should also be run on the VMD-enabled cluster to verify that works as intended as well.

phender:
> @Library(value="pipeline-lib@PR-238")

PR to verify compatibility with existing CI clusters: https://github.com/daos-stack/daos/pull/6508

PR #245, opened 2021-09-03
Support stage names with CentOS 8.3.2011in them

brianjmurrell:


vars/parseStageInfo.groovy
```diff
@@ -54,6 +54,8 @@ def call(Map config = [:]) {
       result['target'] = 'centos7'
     } else if (stage_name.contains('CentOS 8')) {
       result['target'] = 'centos8'
+    } else if (stage_name.contains('CentOS 8.3.2011')) {
+      result['target'] = 'centos8.3'
```
JohnMalmberg, line 58: It looks to me that it that the `contains('CentOS 8')` is always going to match before the `contains('CentOS 8.3.2011').

Since contains does a partial match, it only needs the '8.3' not the whole string.

brianjmurrell: Ah yes, you are correct.  I'll push a fix.

JohnMalmberg (approved):

PR #251, opened 2021-09-21
Skip RPM scanning stages by default

brianjmurrell:
Until the Internet I/O operations of it can be made more reliable.


phender (approved):
Ok as a temporary solution.

PR #254, opened 2021-10-11
CORCI-1128 build: Adding pr tests to daily runs

phender:
Hardware tests with 'pr' tags do not get run in landing PRs nor daily
PRs.  THese tests will no be included as part of the master daily run.

### DIFF
--- a/vars/parseStageInfo.groovy
+++ b/vars/parseStageInfo.groovy
@@ -183,7 +183,7 @@ def call(Map config = [:]) {
                 if (env.BRANCH_NAME.startsWith("weekly-testing")) {
                   tag = "full_regression"
                 } else {
-                  tag = "daily_regression"
+                  tag = "pr daily_regression"
                 }
               } else {
                 // Must be a PR run


vars/parseStageInfo.groovy
```diff
@@ -183,7 +183,7 @@ def call(Map config = [:]) {
                 if (env.BRANCH_NAME.startsWith("weekly-testing")) {
                   tag = "full_regression"
                 } else {
-                  tag = "daily_regression"
+                  tag = "pr daily_regression"
```
brianjmurrell, line 186: This will work.  Although I think it's more transparent to add the `daily_regression` tag to the tests that we want to include in the daily testing (i.e. all of the `pr` tagged tests currently) run.  That would not leave somebody who doesn't have the _tribal knowledge_ that daily testing includes the `pr` tag not understanding why a test that does not have the `daily_regression` tag (which does exist, adding to the confusion) is in the daily testing run.

It also allows for future flexibility in removing a particular `pr` tagged test from the daily run.

While I understand my suggestion here has been made and rejected, I just wanted to note it for the record.

ashleypittman:
@brianjmurrell can you comment on the test failure for this PR itself?

brianjmurrell:
> @brianjmurrell can you comment on the test failure for this PR itself?

It's a known failing test that nobody has had any time to look into dealing with.

jolivier23 (commented):
Is the commit comment supposed to say "These tests will now" ?

PR #260, opened 2021-10-27
DAOS-8915 build: Build on Leap 15.3

brianjmurrell:


vars/buildRpm.groovy
```diff
@@ -52,7 +52,8 @@ def call(Map config = [:]) {
   checkoutScm withSubmodules: true
 
   def env_vars = ''
-  env_vars = ' TARGET=' + stage_info['target']
+  env_vars = ' TARGET=' + stage_info['target'] +
+             ' CI_TARGET=' + stage_info['ci_target']
```
JohnMalmberg, line 56: CI_TARGET is being used to select alternate snapshot image names for testing.
I do not think that is compatible for selecting the specific alternate version that we would want to build packages for.
I think we need a new name for that, like "PKG_TARGET"

JohnMalmberg (changes requested):

vars/buildRpm.groovy
brianjmurrell: > CI_TARGET is being used to select alternate snapshot image names for testing.

Understood.  And is used either in a PR by specifying `<distro>-target: <name>` or by using the build parameter.  There is nothing about that use-case that is incompatible with this use case.  This extends that use-case to also allow those methods to specify a non-default distro version, such as when doing builds on alternate versions of the disto.

> I do not think that is compatible for selecting the specific alternate version that we would want to build packages for.

What do you see as incompatible with it?  I have been using it successfully with daos-stack/mercury#67.

But I don't see how this use-case for CI_TARGET is incompatible.

JohnMalmberg (approved):

PR #263, opened 2021-11-11
Can't test RPMs if we don't build them

brianjmurrell:


JohnMalmberg (commented):
What about the case where we specify the RPM-test-version pragma?

brianjmurrell:
> What about the case where we specify the RPM-test-version pragma?

Excellent point!  Let me address that.

PR #287, opened 2022-02-24
CORCI-1161 build: Migrate to more generic EL name

brianjmurrell:
Add support for other EL variants.

PR #289, opened 2022-02-28
CORCI-1162 Build: Fix Ubuntu Artifact Upload

JohnMalmberg:
Test PR to debug/verify repository updating


brianjmurrell (commented):
Should we add a Ubuntu upload test to this repo now?

vars/buildRpmPost.groovy
```diff
@@ -69,20 +69,26 @@ def call(Map config = [:]) {
...
+    if (!target.startsWith('ubuntu')) {
         stash name: target + '-required-mercury-rpm-version',
               includes: target + '-required-mercury-rpm-version'
     }
 
+    String repo_format = 'yum'
+    if (target.startsWith('ubuntu')) {
+        repo_format = 'apt'
+    }
+
```
jolivier23, line 81:
```suggestion
    String repo_format = 'yum'
    if (!target.startsWith('ubuntu')) {
        stash name: target + '-required-mercury-rpm-version',
              includes: target + '-required-mercury-rpm-version'
    } else { 
        repo_format = 'apt'
    }

```

jolivier23 (changes requested):

Jenkinsfile
```diff
@@ -27,6 +28,8 @@ pipeline {
         SSH_KEY_FILE='ci_key'
         SSH_KEY_ARGS="-i$SSH_KEY_FILE"
         CLUSH_ARGS="-o$SSH_KEY_ARGS"
+        ARTIFACTORY_URL = 'https://artifactory.dc.hpdd.intel.com/'
```
brianjmurrell, line 31: Just a note to myself to make sure this is removed before final review/landing.

Jenkinsfile
```diff
@@ -16,6 +16,7 @@
 // Then a second PR submitted to comment out the @Library line, and when it
 // is landed, both PR branches can be deleted.
 //@Library(value="pipeline-lib@my_branch_name") _
+@Library(value="pipeline-lib@corci-1162") _
```
brianjmurrell, line 19: Just a note to myself to make sure this is removed before final review/landing.

brianjmurrell, line 19:
```suggestion
@Library(value="pipeline-lib@corci-1162a") _
```

brianjmurrell (changes requested):

vars/provisionNodesV1.groovy
```diff
@@ -257,7 +257,7 @@ EOF'''
     provision_script += '''\nrm -f /etc/profile.d/openmpi.sh
                              rm -f /tmp/daos_control.log
                              yum -y erase metabench mdtest simul IOR compat-openmpi16
```
brianjmurrell, line 259: Why not use `dnf` here also?  Mixing `yum` and `dnf` means duplicate package libraries including duplicate downloads of repo data.

There might be other uses of YUM that we want to switch to DNF.

brianjmurrell (commented):
Generally looks good.  I do think we need to think about the `yum`->`dnf` change in provisionNodesV1 though.  Hoping you can do some analysis about what the fallout of that change could be.  Is it going to end up getting used (i.e. in some historic release) where DNF is not actually available?

vars/provisionNodesV1.groovy
```diff
@@ -257,7 +257,7 @@ EOF'''
...
-                             if ! yum -y install --exclude=ompi CUnit python36-PyYAML          \
+                             if ! dnf -y install --exclude=ompi CUnit python36-PyYAML          \
```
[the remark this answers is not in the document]
JohnMalmberg: Added making sure DNF is installed, which means that provisionNodesV1 is as backward compatible as currently possible.

vars/provisionNodesV1.groovy
brianjmurrell: >  I think we want look at removing the tests using provisionNodesV1 instead of trying to optimize it.

daos-stack/pipeline-lib#287 does that IIRC.

brianjmurrell (approved):

PR #291, opened 2022-03-03
Convert env[NAME] to env."NAME"

brianjmurrell:
To avoid having to whitelist org.codehaus.groovy.runtime.DefaultGroovyMethods getAt java.lang.Object java.lang.String


SoftBob:
def branch = env.GIT_BRANCH.toUpperCase().replaceAll("-", "_"))  
need modification
def branch = env.GIT_BRANCH.toUpperCase().replaceAll("-", "_")

PR #292, opened 2022-03-15
DAOS-10144 build: Update docker cachebust to be weekly.

ashleypittman:


brianjmurrell:
> Test stage provisionNodes with slurm Leap15 completed with status FAILURE. https://build.hpdd.intel.com/job/daos-stack/job/pipeline-lib/job/PR-292/5/display/redirect

This is looking like a broken repo on Nexus.  I have alerted @JohnMalmberg to apply whatever correction is necessary for this situation.

brianjmurrell:
@ashleypittman Note that I have moved this ticket from DCO-10286 to DAOS-10144 as this really has nothing to do with DCO and everything to do with how DAOS is built within the DAOS project.

brianjmurrell:
The broken test was a Nexus issue and is fixed now and the testing has all passed on this PR now.

ashleypittman:
> The broken test was a Nexus issue and is fixed now and the testing has all passed on this PR now.

Thanks, added @daos-stack/daos-gatekeeper 

As you know this may mean we're using out-of-date software which I don't think matters for OS updates, but for our own repos daily, or 4-8 hour cadene may be more appropriate here.

ashleypittman:
Do we want to land this or shall I abandon it?

It would lead to faster CI times (at least for checkpatch etc) and lower load on the repo servers although maybe that's less important than it was when this was written.

vars/dockerBuildArgs.groovy
```diff
@@ -114,7 +114,7 @@ String call(Map config = [:]) {
                      " --build-arg JENKINS_URL=$env.JENKINS_URL"
     if (cachebust) {
       Calendar current_time = Calendar.getInstance()
-      ret_str += " --build-arg CACHEBUST=${currentBuild.startTimeInMillis}"
+      ret_str += " --build-arg CACHEBUST=" + current_time.get(Calendar.WEEK_OF_YEAR)
```
brianjmurrell, line 117: (Realizing that I +1'd this,) this really should not have been changed here.  `CACHEBUST` was (is) supposed to always guarantee that the remaining portion of a `Dockerfile` will be executed.  While it was being used to force updates, that was not it's specific goal.  There are other use-cases for wanting to force execution in a `Dockerfile`.

If we wanted to create a new trigger/timer for doing updates, a new variable should have been created.

We probably want a number of `CACHEBUST` variables at various granularities named accordingly.  I.e. CB_WEEKLY.

PR #297, opened 2022-04-13
Use a matrix for testing

brianjmurrell:


JohnMalmberg (approved):
Still have concerns if Jenkins fixed their sparse matrix bug.

brianjmurrell:
> Still have concerns if Jenkins fixed their sparse matrix bug.

Which is fair.  But since my test of the suspected issue did not reproduce the issue it seems reasonable to go forward on the hope/assumption that they have fixed the issue.  We can certainly revisit this if the issue does start to occur.

PR #306, opened 2022-05-09
DAOS-9989 test: Adding Test-provider* commit pragma

phender:
Adding a new Test-provider* commit pragam to enable passing a value to
the launch.py --provider argument.  This will override the default
provider specified in the test daos_server configuration file.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -16,7 +16,6 @@
 // Then a second PR submitted to comment out the @Library line, and when it
 // is landed, both PR branches can be deleted.
 //@Library(value="pipeline-lib@my_branch_name") _
-@Library(value="pipeline-lib@DAOS-9989") _
 
 pipeline {
     agent { label 'lightweight' }


vars/parseStageInfo.groovy
```diff
@@ -267,6 +267,26 @@ def call(Map config = [:]) {
...
     }
 
+    String provider
+    // Highest priority is TestProvider parameter
+    if (startedByUser() && params.TestProvider && params.TestProvider != "") {
+      provider = params.TestProvider
+    } else {
+      // Next highest priority is a stage specific Test-provider-*
+      provider = cachedCommitPragma("Test-provider" + result['pragma_suffix'], null)
+      if (!repeat) {
```
brianjmurrell, line 277:
```suggestion
      if (!provider) {
```
?

vars/parseStageInfo.groovy
```diff
@@ -267,6 +267,26 @@ def call(Map config = [:]) {
...
+        // Followed by the more general Test-provider:
+        provider = cachedCommitPragma("Test-provider", null)
```
brianjmurrell, line 279:
```suggestion
      provider = cachedCommitPragma("Test-provider" + result['pragma_suffix'], cachedCommitPragma("Test-provider", null))
```

vars/parseStageInfo.groovy
```diff
@@ -267,6 +267,26 @@ def call(Map config = [:]) {
...
+      }
+    }
+    if (provider) {
+      if (result['ftest_arg']) {
+        result['ftest_arg'] += " --provider='" + provider + "'"
+      } else {
+        result['ftest_arg'] += "--provider='" + provider + "'"
+      }
```
brianjmurrell, line 287:
```suggestion
      if (result['ftest_arg']) {
        result['ftest_arg'] += ' '
      }
      result['ftest_arg'] += "--provider='" + provider + "'"
```
or
```suggestion
      result['ftest_arg'] += result['ftest_arg'] ? '' : ' ' + "--provider='" + provider + "'"
```
but maybe better still would be to just always add the trailing ' ' on every `result['ftest_arg'] += ` and then `.trim()` it at the end.  Or create a temporary list and then join it into a string at the end.

Also wonder how this squares with https://github.com/daos-stack/pipeline-lib/blob/2dca7b4d389dcb74de7ff34ee343b3e817ca46f3/vars/parseStageInfo.groovy#L306-L308.  Is it understood that any `ftest_arg` value passed by the caller will void adding `--provider` (and indeed, `--repeat`) by the pragmas or parameters here?

brianjmurrell (changes requested):
Whatever refactoring you decide to do with `result['ftest_arg']` could/should be applied equally to the `repeat` handling that it was copy'n'pasted from.

vars/parseStageInfo.groovy
phender: Applied to both 'Test-provider' and 'Test-repeat'

vars/parseStageInfo.groovy
brianjmurrell: That's fine.  I just wanted to point out the potential here.  If you are OK with it, it is fine with me also.

brianjmurrell (approved):

vars/skipStage.groovy
```diff
@@ -420,6 +420,18 @@ boolean call(Map config = [:]) {
         case "Bullseye Report on EL 8":
             return env.BULLSEYE == null ||
                    skip_stage_pragma('bullseye', 'true')
+        case "Functional Hardware Small TCP":
+            return ! paramsValue('CI_FUNCTIONAL_HARDWARE_SMALL_TCP', true)
```
brianjmurrell, line 424: We're starting to write Groovy according to coding conventions and standards.  We don't have a linting stage in CI yet, but I have a linter in VS Code and I know that it will tell me that there is not supposed to be a space after the `!` operator.

vars/parseStageInfo.groovy
```diff
@@ -252,19 +252,29 @@ def call(Map config = [:]) {
...
-        result['ftest_arg'] += " --repeat=" + repeat
-      } else {
-        result['ftest_arg'] += "--repeat=" + repeat
-      }
+      result['ftest_arg'] += " --repeat=" + repeat
+    }
+
+    String provider
+    // Highest priority is TestProvider parameter
+    if (startedByUser() && params.TestProvider && params.TestProvider != "") {
```
brianjmurrell, line 265: Does using this parameter **really** want to be limited to jobs started (i.e. manually) by a user?  What if a job wants to run another job with this parameter set?

phender: This isn't any different than what we do with the `TestTag` and `TestRepeat` 'Build with Parameters' options.  It seems like that would be a larger PR.

vars/skipStage.groovy
phender: I can change this for the new code, but I just copied what was already established in the file so there will be a large number of offenses.

vars/parseStageInfo.groovy
brianjmurrell: Since this PR is adding this specific code, I think it's fair to ask the question and possibly change the code depending on the answer and leaving the others to be updated organically if/when a consumer wants to change their behaviour.  This could perhaps happen sooner than later as we are starting to add the running of the daos pipeline as a testing stage of other pipelines.

phender: Done

vars/skipStage.groovy
phender: Done

phender:
Test PR for these changes https://build.hpdd.intel.com/job/daos-stack/job/daos/job/PR-9188/4/

phender:
The vars/*.groovy file changes have been verified by https://build.hpdd.intel.com/job/daos-stack/job/daos/job/PR-9188/6/ - all tests passed.

phender:
All stages passed referencing this pipeline-lib PR in https://build.hpdd.intel.com/blue/organizations/jenkins/daos-stack%2Fpipeline-lib/detail/PR-306/42/pipeline/

JohnMalmberg (approved):

PR #309, opened 2022-05-11
DAOS-10585 test: Disable runnig Functional * w/ Valgrind stage

phender:
The 'Functional on * with Valgrind' stage is now skipped by default and
requires a 'Skip-func-test-vm-valgrind: false' commit pragma to enable
running the stage.

### DIFF
--- a/vars/skipStage.groovy
+++ b/vars/skipStage.groovy
@@ -78,7 +78,9 @@ boolean skip_ftest(String distro, String target_branch) {
 }
 
 boolean skip_ftest_valgrind(String distro, String target_branch) {
-    if (! skip_stage_pragma('func-test-vm-valgrind', 'true')) {
+    // Check if the default for skipping this stage been overriden
+    if (run_default_skipped_stage('func-test-vm-valgrind')) {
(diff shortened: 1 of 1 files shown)


frostedcmos (approved):

sylviachanoiyee (approved):

vars/skipStage.groovy
```diff
@@ -78,7 +78,9 @@ boolean skip_ftest(String distro, String target_branch) {
 }
 
 boolean skip_ftest_valgrind(String distro, String target_branch) {
-    if (! skip_stage_pragma('func-test-vm-valgrind', 'true')) {
```
brianjmurrell, line 81: It's strange.  I would have thought that this already skipped the stage by default by passing `'true'` as the default value:
https://github.com/daos-stack/pipeline-lib/blob/293825867929a6bcb544d82514ef49b1b35ef39e/vars/skipStage.groovy#L14-L16

`cachedCommitPragma()` should return it's second argument if the pragma is not specified.

brianjmurrell (approved):

PR #310, opened 2022-05-12
DAOS-10585 test: Always skip the w/ Valgrind stage

phender:
Only run the the Functional on * with Valgrind test stage if the
Skip-func-test-vm-valgrind pragma is set to false.  All other
conditionals that previously determined if the stage was skipped have
been removed.


sylviachanoiyee (approved):

phender:
Testing change with https://github.com/daos-stack/daos/pull/9003

brianjmurrell (approved):

PR #312, opened 2022-05-12
CORCI-1200 Build: Add Job Status Reporting

JohnMalmberg:
Add methods to allow recording job status while the job runs
and write out a YAML file with the results.


vars/jobStatusUpdate.groovy
```diff
@@ -0,0 +1,19 @@
...
+ *
+ */
+
+def call(String name='', String value='') {
+    if (name == '') {
+        name = env.STAGE_NAME
+    }
+    if (value == '') {
+        value = currentBuild.currentResult
+    }
```
brianjmurrell, line 15: You are effecting parameter defaults here which can be simplified as:

```suggestion
def call(String name=env.STAGE_NAME, String value=currentBuild.currentResult) {
```

vars/jobStatusWrite.groovy
```diff
@@ -0,0 +1,25 @@
...
+    job_name += '_' + env.BUILD_NUMBER
+    String file_name = env.DAOS_STACK_JOB_STATUS_DIR + '/' + job_name
+
+    String job_status_text = writeYaml(data: jobStatus(),
+                                returnText: true)
+    // Need to use shell script for creating files that are not
+    // in the workspace.
+    job_status_text.split("\n").each { String line ->
+      sh(label: "Write jenkins_job_staus ${line}",
+         script: "echo '${line}' >> ${file_name}")
```
brianjmurrell, line 23: Isn't this going to create a job step for every line written?  Can the whole YAML file not just be written all at once?

JohnMalmberg: Currently nothing is working other than it is not crashing from groovy syntax errors.
Having issue with determining the actual scope of a groovy global variable.

Jenkinsfile
```diff
@@ -370,9 +415,16 @@ pipeline {
                                 }
                             }
                         }
+                        job_status_update()
                     } // steps
-                } // stage ('Commit Pragma tests') {
+                } // stage ('Commit Pragma tests') */
```
brianjmurrell, line 420: Where does the comment that this line is ending start?

vars/runTest.groovy
```diff
@@ -185,4 +185,5 @@ def call(Map config = [:]) {
 	    }
     }
 
+    return ['run_test': status]
```
brianjmurrell, line 188: Doesn't the catchError() set `currentBuild.currentResult` to final status of this stage, negating the need to pass it back explicitly?

Jenkinsfile
```diff
@@ -17,6 +17,37 @@
...
+    name = name.replace(' ', '_')
+    name = name.replace('.', '_')
+    job_status_internal[name] = value
+}
+
+def job_step_update(value) {
+    // Wrapper around a pipeline step to obtain a status.
+    job_status_update(env.STAGE_NAME, value)
+}
+
```
brianjmurrell, line 50: How does all of this get into other projects?

Jenkinsfile
```diff
@@ -185,6 +225,7 @@ pipeline {
                                            yum --disablerepo=\\* --enablerepo=build\\* makecache"''',
                                 junit_files: null,
                                 failure_artifacts: env.STAGE_NAME
+                        job_status_update()
```
brianjmurrell, line 228: Doesn't this need to wrap the above runTest() the way it does in other stages?

Jenkinsfile
```diff
@@ -208,6 +249,7 @@ pipeline {
...
```
brianjmurrell, line 252: Ditto.  Won't repeat for further instances.

Jenkinsfile
JohnMalmberg: It needs to be in the Jenkinsfile, that is the only place simple global "variables" now work.

Jenkinsfile
```diff
@@ -58,12 +89,14 @@ pipeline {
                         }
                     }
                     steps {
-                        runTest script: '''set -ex
+                        // jobStatus('JUNIT_failure')
```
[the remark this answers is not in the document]
JohnMalmberg: Yes, will remove on next update.

Jenkinsfile
JohnMalmberg: No, just an example showing two different ways of updating the job status.
For a simple status, just call jobs_status_update() with no parameters.
For a more complex status, job_status_update() must wrap a build step that returns a Map.

Jenkinsfile
JohnMalmberg: Missed edit, will remove on next push.

vars/runTest.groovy
JohnMalmberg: This is mainly a simple example.   This can be a more complex Map with more data if needed.

Jenkinsfile
brianjmurrell: :frowning:

Jenkinsfile
```diff
@@ -17,36 +18,38 @@
...
 
-def job_status_update(String name=env.STAGE_NAME,
+ // groovylint-disable MethodParameterTypeRequired, NoDEF
+ void jobStatusUpdate(String name=env.STAGE_NAME,
                       value=currentBuild.currentResult) {
-    name = name.replace(' ', '_')
-    name = name.replace('.', '_')
-    job_status_internal[name] = value
+    String key = name.replace(' ', '_')
+    key = key.replace('.', '_')
```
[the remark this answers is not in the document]
JohnMalmberg: Fixed in current push

Jenkinsfile
```diff
@@ -17,36 +18,38 @@
 // is landed, both PR branches can be deleted.
 //@Library(value="pipeline-lib@my_branch_name") _
 
-job_status_internal = [:]
+jobStatusInternal = [:]
```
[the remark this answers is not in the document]
JohnMalmberg: Converted to non camel-case, and added groovylint ignores for that.

phender (approved):

brianjmurrell (approved):

kprantis:
This is ready for gatekeeping?

JohnMalmberg:
This is not needed for the job status reporting, it demonstrates how to add additional detail to the job status reporting.
The demonstration does not actually add any significantly more detail.

There is a DAOS PR that actually adds the job status reporting.

kprantis:
Ah, I dont think i grokked that from the commit description.

jolivier23:
Requires force landing without updating.

JohnMalmberg:
This is a demonstration PR on how to add more detail to the job status reporting that is now landed in daos/master branch.
More work would be needed in this PR to actually add more useful data.

JohnMalmberg:
Obsolete

PR #320, opened 2022-06-22
SRE-97 Build: Add support for el-8 unit tests

JohnMalmberg:
Add support for el-8 Unit tests.

### DIFF
--- a/vars/sconsBuild.groovy
+++ b/vars/sconsBuild.groovy
@@ -135,18 +135,11 @@ def call(Map config = [:]) {
         set_cwd = "cd ${config_target}\n"
     }
 
-    String scons_exe = 'scons'
-    int sctest = sh label: 'probe for scons command',
-                    script: 'command -v scons',
-                    returnStatus: true
-    if (sctest != 0) {
-        sctest = sh label: 'probe for scons-3 command',
-                    script: 'command -v scons-3',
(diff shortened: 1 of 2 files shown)


vars/sconsBuild.groovy
```diff
@@ -135,7 +135,19 @@ def call(Map config = [:]) {
...
+                    returnStatus: true
+    if (sctest != 0) {
+        sctest = sh label: 'probe for scons-3 command',
+                    script: 'command -v scons-3',
+                    returnStatus: true
+        if (sctest == 0) {
+            scons_exe = 'scons-3'
+        }
+    }
+
```
ashleypittman, line 150: All the dockerfiles have used 'scons' installed through pip for some months now, no change to support dockerfile building should be considering scons-3 at all.

brianjmurrell (approved):

ashleypittman (changes requested):
@brianjmurrell is correct, but I do not believe we need the scons/scons-3 change at all.

JohnMalmberg:
Current push uses more concise code, which will try the traditional name first and the el-8 distro name second so that if we ever need to make a change to use the distro scons on el-8, we do not have to also modify pipeline-lib.

brianjmurrell (approved):

PR #13, opened 2018-12-13
CORCI-445 Adding node provisioning

JohnMalmberg:
* Add provisionNodes step.
  Allows snapshot option for VMs, and a power cycle only option.
  By default waits for the provisioning or power cycle to complete.

* Update checkPatch step credential ID.

Signed-off-by: John.E.Malmberg <john.e.malmberg@intel.com>

PR #324, opened 2022-07-15
SRE-368 Build: Add test status artifacts

JohnMalmberg:
Test status artifacts are used to track if a test stage has completed.
This is to allow a restart from stage operation to skip completed
tests in that stage.

### DIFF
--- a/vars/skipStage.groovy
+++ b/vars/skipStage.groovy
@@ -43,10 +43,12 @@ boolean already_passed() {
                       selector: specific(old_build)
         try {
             String stage_status = readFile(file: status_file)
-            println("stage ${env.STAGE_NAME} status '${stage_status}'")
             if (stage_status == 'SUCCESS') {
                 return true
             }
+            println("Previous run this stage ended with status " +
+                    "'${stage_status}', so re-running")
+
         } catch (java.nio.file.NoSuchFileException e) {
             // This should not ever fail, so just collecting diagnostics
             // if the code ever gets here.
@@ -64,6 +66,8 @@ boolean already_passed() {
         // Try using httpRequests, which does not require a modified
         // Jenkinsfile, but makes assumptions on where Jenkins is
         // currently storing the artifacts for a job.
+        // This is for transitioning until all Jenkinsfiles are allowing
+        // artifacts to be copied and then can be removed.
         String my_build = "/${env.BUILD_NUMBER}/"
         String prev_build = "/${old_build}/"
         String old_job = env.BUILD_URL.replace(my_build, prev_build)
@@ -112,10 +116,8 @@ boolean is_pr() {
 }
 
 boolean skip_scan_rpms(String distro, String target_branch) {
-    if (already_passed()) {
-        return true
-    }
-    return target_branch == 'weekly-testing' ||
+    return already_passed() ||
+           target_branch == 'weekly-testing' ||
            skip_stage_pragma('scan-rpms', 'true') ||
            (distro == 'centos-7' &&
             (! paramsValue('CI_SCAN_RPMS_el7_TEST', true)) ||
@@ -157,24 +159,20 @@ boolean skip_ftest(String distro, String target_branch) {
 boolean skip_ftest_valgrind(String distro, String target_branch) {
     // Check if the default for skipping this stage been overriden
     // otherwise always skip this stage (DAOS-10585)
-    if (already_passed()) {
-        return true
-    }
-    return ! run_default_skipped_stage('func-test-vm-valgrind')
+    return already_passed() ||
+           !run_default_skipped_stage('func-test-vm-valgrind')
 }
 
 boolean skip_ftest_hw(String size, String target_branch) {
-    if (already_passed()) {
-        return true
-    }
-    return env.DAOS_STACK_CI_HARDWARE_SKIP == 'true' ||
-           ! paramsValue('CI_' + size.replace('-', '_') + '_TEST', true) ||
+    return already_passed() ||
+           env.DAOS_STACK_CI_HARDWARE_SKIP == 'true' ||
+           !paramsValue('CI_' + size.replace('-', '_') + '_TEST', true) ||
            skip_stage_pragma('func-test') ||
            skip_stage_pragma('func-hw-test-' + size) ||
-           ! testsInStage() ||
+           !testsInStage() ||
            ((env.BRANCH_NAME == 'master' ||
              env.BRANCH_NAME.startsWith('release/')) &&
-            ! (startedByTimer() || startedByUser())) ||
+            !(startedByTimer() || startedByUser())) ||
            cachedCommitPragma('Run-daily-stages') == 'true' ||
            (docOnlyChange(target_branch) &&
             prRepos(hwDistroTarget(size)) == '')
@@ -372,25 +370,26 @@ boolean call(Map config = [:]) {
                     docOnlyChange(target_branch) ||
                     skip_build_on_el_gcc(target_branch, '8') ||
                     skip_stage_pragma('unit-tests')
-        case "NLT":
-        case "NLT on CentOS 8":
-        case "NLT on EL 8":
-            return skip_stage_pragma('nlt') || already_passed()
-        case "Unit Test Bullseye":
-        case "Unit Test Bullseye on CentOS 8":
-        case "Unit Test Bullseye on EL 8":
+        case 'NLT':
+        case 'NLT on CentOS 8':
+        case 'NLT on EL 8':
+            return skip_stage_pragma('nlt') ||
+                   already_passed()
+        case 'Unit Test Bullseye':
+        case 'Unit Test Bullseye on CentOS 8':
+        case 'Unit Test Bullseye on EL 8':
             return skip_stage_pragma('bullseye', 'true') ||
                    already_passed()
-        case "Unit Test with memcheck on CentOS 8":
-        case "Unit Test with memcheck on EL 8":
-        case "Unit Test with memcheck":
-            return ! paramsValue('CI_UNIT_TEST_MEMCHECK', true) ||
+        case 'Unit Test with memcheck on CentOS 8':
+        case 'Unit Test with memcheck on EL 8':
+        case 'Unit Test with memcheck':
+            return !paramsValue('CI_UNIT_TEST_MEMCHECK', true) ||
                    skip_stage_pragma('unit-test-memcheck') ||
                    already_passed()
-        case "Unit Test":
-        case "Unit Test on CentOS 8":
-        case "Unit Test on EL 8":
-            return ! paramsValue('CI_UNIT_TEST', true) ||
+        case 'Unit Test':
+        case 'Unit Test on CentOS 8':
+        case 'Unit Test on EL 8':
+            return !paramsValue('CI_UNIT_TEST', true) ||
                    skip_stage_pragma('unit-test') ||
                    skip_stage_pragma('run_test') ||
                    already_passed()
--- a/vars/stageStatusFilename.groovy
+++ b/vars/stageStatusFilename.groovy
@@ -10,8 +10,15 @@ String urlEncode() {
   /**
    * stageStatusFilename step method
    *
+   * This does not create the file, it is so the methods
+   * that create / fetch / read the STAGE status file all use
+   * the same name and it is properly URLEncoded.
+   *
+   * It does make sure that the directory for the stage
+   * status file exists.
+   *
    * @param config Map of parameters passed, Unused
-   * returns: String with Url Encoded filename for stage name status.
+   * returns: String with Url Encoded pathname for the stage status.
    */
 
 String call(Map config= [:]) {


brianjmurrell:
> Test job is https://build.hpdd.intel.com/job/daos-stack/job/daos/view/change-requests/job/PR-9722/ Test job #12 was aborted after the Functional VM el-8 stage passed, and then restarted from the Test stage as #13 where it detected that it did not need to run the functional vm el-8 test stage again.

Did it?  It looks to me like #12 failed with "No such property: True for class: groovy.lang.Binding'.  I think maybe you are off-by-one.

#13 looks like it was aborted after the Functional VM stage and #14 looks like it was the resumption.

A more complete demonstration, I think should run a number of stages in a parallel group, with one failing and the subsequent run shows that only the failed stage reruns.

vars/stageStatusFilename.groovy
```diff
@@ -0,0 +1,19 @@
+// vars/stageStatusFilename.groovy
+
+import java.net.URLEncoder
+
+@NonCPS
+String urlEncode() {
+    return URLEncoder.encode(env.STAGE_NAME, 'UTF-8')
+}
```
brianjmurrell, line 8: Is a function really necessary here for a single line of un-reused code?  Couldn't this just be done in-place in the `call()`?

vars/stageStatusFilename.groovy
```diff
@@ -0,0 +1,19 @@
...
+  /**
+   * stageStatusFilename step method
+   *
+   * @param config Map of parameters passed, Unused
+   * returns: String with Url Encoded filename for stage name status.
+   */
+
+String call(Map config= [:]) {
+    return urlEncode() + ".status"
+}
```
brianjmurrell, line 19: Might as well add a newline here if you have to push again.

vars/skipStage.groovy
```diff
@@ -10,6 +13,34 @@
...
+
+    def response = httpRequest url: art_url,
+                               acceptType: 'TEXT_PLAIN',
+                               httpMode: 'GET',
+                               validResponseCodes: '100:599'
+
+    if (response.status == 200 && response.content == 'SUCCESS') {
+        return true
+    }
+    return false
```
brianjmurrell, line 41: Couldn't you use the more idiomatic [Copy Artifact](https://plugins.jenkins.io/copyartifact/) plugin for this?

vars/skipStage.groovy
```diff
@@ -293,19 +335,31 @@ boolean call(Map config = [:]) {
         case "NLT":
         case "NLT on CentOS 8":
         case "NLT on EL 8":
+            if (already_passed()) {
+                return true
+            }
             return skip_stage_pragma('nlt')
```
brianjmurrell, line 341:
```suggestion
            return skip_stage_pragma('nlt') ||
                   already_passed()
```

Ditto for many others I won't repeat.

brianjmurrell (commented):
No one single thing that warrants a -1 but enough suggestions worth thinking about at least I think,

vars/skipStage.groovy
JohnMalmberg: I have not been able to get the CopyArtifact to work reliably with the Github multibranch pipeline jobs as a target.  The required syntax is not documented for that Plugin.

vars/skipStage.groovy
JohnMalmberg: Will be in the next push

vars/stageStatusFilename.groovy
JohnMalmberg: I thought that the @NonCPS code needed to be in its own function not inline code.
Documentation refers to steps as being CPS transformed and methods can be annotated as @NonCPS.

vars/stageStatusFilename.groovy
JohnMalmberg: Will be in the next push.

JohnMalmberg:
Wanted to get reviews of the general structure done.
Currently do not have a way to inject a failure in the original job step that would not be present on the retry with out doing something like going into a host used for a test and doing something to break the test.

vars/stageStatusFilename.groovy
brianjmurrell: I have not really looked into what all of that [Non]CPS stuff is TBH.  Probably should at some point.  But I wonder if it really matters for one-liner.

vars/skipStage.groovy
brianjmurrell: Work at all or just reliably?  Like it works sometimes and not others?

brianjmurrell:
> Currently do not have a way to inject a failure in the original job step that would not be present on the retry with out doing something like going into a host used for a test and doing something to break the test.

You could create an avocado test that fails on evenly numbered jobs and passes on odd numbered jobs and then you'd be able to demo any time you want -- having it fail, then having the retry succeed.  You could repeat the demo by just running it again and then retrying.

vars/skipStage.groovy
JohnMalmberg: Current code now first tries to use copyArtifact which requires being enabled in the Jenkinsfile to work, and if copyArtifact is not able to find the artifact, it falls back to httpRequest to read the artifact, which allows this change to be used before the Jenkinsfiles are updated.

Also moves the artifact file into a separate directory in the artifacts.

JohnMalmberg:
https://build.hpdd.intel.com/job/daos-stack/job/daos/view/change-requests/job/PR-9722/43/ ran with post-provision modified to fail odd numbered builds.   It is also configured to not to enable copyArtifact to copy artifacts.
https://build.hpdd.intel.com/job/daos-stack/job/daos/view/change-requests/job/PR-9722/44/ was restarted from test stage and is re-running the failed stage.   It successfully ran the SMALL test and was aborted while running the medium test.
https://build.hpdd.intel.com/job/daos-stack/job/daos/view/change-requests/job/PR-9722/45/ was restarted from test stage.
It detected that the "Functional on EL 8" test and "Functional Hardware Small" did not need to be re-run.  As it is an odd numbered test, it is going to fail the Medium and Large tests once it gets a cluster, if the job does not get aborted.

vars/skipStage.groovy
```diff
@@ -10,6 +13,74 @@
...
+    // First try looking up using copyArtifacts, which requires
+    // that the Jenkinsfile give permission for the copy.
+    try {
+        copyArtifacts projectName: env.JOB_NAME,
+                      optional: false,
+                      filter: status_file,
+                      selector: specific(old_build)
+        try {
+            String stage_status = readFile(file: status_file)
+            println("stage ${env.STAGE_NAME} status '${stage_status}'")
```
brianjmurrell, line 46: Maybe make this message a bit more informative?

```suggestion
            println("Previous run, this stage ended with status '${stage_status}'")
```

vars/skipStage.groovy
```diff
@@ -10,6 +13,74 @@
...
+            if (stage_status == 'SUCCESS') {
+                return true
+            }
```
brianjmurrell, line 49: Or maybe better yet, move the message here and add to it that the stage is rerunning:

```suggestion
            }
            println("Previous run, this stage ended with status '${stage_status}'", so rerunning)
```

vars/skipStage.groovy
```diff
@@ -10,6 +13,74 @@
...
+            if (fileExists(status_file)) {
+                println("fileExists found ${status_file}")
+            }
+        }
+    } catch (hudson.AbortException e) {
+        println("Informational: copyArtifact could not get artifact ${e}")
+
+        // Try using httpRequests, which does not require a modified
+        // Jenkinsfile, but makes assumptions on where Jenkins is
+        // currently storing the artifacts for a job.
```
brianjmurrell, line 66:
```suggestion
        // currently storing the artifacts for a job.
        // This should just be transitional until all Jenkinsfiles are allowing artifacts to be copied and should be removed once that has been achieved.
```

vars/skipStage.groovy
```diff
@@ -41,6 +112,9 @@ boolean is_pr() {
 }
 
 boolean skip_scan_rpms(String distro, String target_branch) {
+    if (already_passed()) {
+        return true
+    }
     return target_branch == 'weekly-testing' ||
```
brianjmurrell, line 118:
```suggestion
    return already_passed() ||
           target_branch == 'weekly-testing' ||
```

vars/skipStage.groovy
```diff
@@ -80,10 +157,16 @@ boolean skip_ftest(String distro, String target_branch) {
 boolean skip_ftest_valgrind(String distro, String target_branch) {
     // Check if the default for skipping this stage been overriden
     // otherwise always skip this stage (DAOS-10585)
+    if (already_passed()) {
+        return true
+    }
     return ! run_default_skipped_stage('func-test-vm-valgrind')
```
brianjmurrell, line 163:
```suggestion
    return  already_passed() ||
           ! run_default_skipped_stage('func-test-vm-valgrind') ||
```

vars/skipStage.groovy
```diff
@@ -80,10 +157,16 @@ boolean skip_ftest(String distro, String target_branch) {
...
 }
 
 boolean skip_ftest_hw(String size, String target_branch) {
+    if (already_passed()) {
+        return true
+    }
     return env.DAOS_STACK_CI_HARDWARE_SKIP == 'true' ||
```
brianjmurrell, line 170:
```suggestion
    return already_passed() ||
           env.DAOS_STACK_CI_HARDWARE_SKIP == 'true' ||
```

vars/stageStatusFilename.groovy
```diff
@@ -0,0 +1,23 @@
+// vars/stageStatusFilename.groovy
+
+import java.net.URLEncoder
+
+@NonCPS
+String urlEncode() {
```
brianjmurrell, line 6: Maybe:

```suggestion
String urlEncodedSTAGE_NAME() {
```

Since I actually missed that this is a hard coding of url encoding of the stage name last time.  However, see below...

vars/stageStatusFilename.groovy
```diff
@@ -0,0 +1,23 @@
...
+   * @param config Map of parameters passed, Unused
+   * returns: String with Url Encoded filename for stage name status.
+   */
+
+String call(Map config= [:]) {
+    String directory = 'stage_status'
+    if (!fileExists(directory)) {
+        fileOperations([folderCreateOperation(directory)])
+    }
+    return directory + '/' + urlEncode() + ".status"
```
brianjmurrell, line 22: Still not sure what the advantage of this is over, say:

```suggestion
    return directory + '/' + URLEncoder.encode(env.STAGE_NAME, 'UTF-8') + ".status"
```

Does whatever `@NonCPS` does matter for a single operation like `URLEncoder.encode()` or does it really only apply to functions of some non-trivial length?

brianjmurrell (changes requested):

vars/stepResult.groovy
```diff
@@ -150,5 +150,7 @@ def call(Map config= [:]) {
   }
   param['status'] = result
 
+  writeFile(file: stageStatusFilename(), text: result)
```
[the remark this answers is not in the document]
JohnMalmberg: No, what is written to Jenkins job status by default is the Jenkins stage status, this is the status reported to the github step.
It would take quite a bit of changes to the Jenkins job status procedure that would need to be rolled out to all the Jenkinsfiles to make it both expandable and usable for both purposes, so for now they need to be kept separate.

vars/skipStage.groovy
JohnMalmberg: Will be in next push

vars/skipStage.groovy
JohnMalmberg: Similar text will be in next push

vars/skipStage.groovy
JohnMalmberg: In next push

vars/skipStage.groovy
JohnMalmberg: In next push

vars/skipStage.groovy
JohnMalmberg: in next push

vars/skipStage.groovy
```diff
@@ -335,34 +335,25 @@ boolean call(Map config = [:]) {
         case "NLT":
         case "NLT on CentOS 8":
         case "NLT on EL 8":
-            if (already_passed()) {
-                return true
-            }
-            return skip_stage_pragma('nlt')
+            return skip_stage_pragma('nlt') || already_passed()
```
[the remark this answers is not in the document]
JohnMalmberg: Will be in the next push

vars/stageStatusFilename.groovy
JohnMalmberg: Updated documentation for the function in the next push.

vars/stageStatusFilename.groovy
JohnMalmberg: Since this is a local function, does the name need to be changed?   I did not change it for the current push.

JohnMalmberg:
The failing tests were running on the master branch of this repo and would not be influenced by the changes here.

https://build.hpdd.intel.com/job/daos-stack/job/daos/view/change-requests/job/PR-9722/50/ ran a functional el-8 and Functional Hardware Small test successfully and was then aborted.
https://build.hpdd.intel.com/job/daos-stack/job/daos/view/change-requests/job/PR-9722/51/  has a modification that caused the Hardware Medium test to fail.   This was actually restarted from the "Test" stage, but did not re-run either previously passed tests.
https://build.hpdd.intel.com/job/daos-stack/job/daos/view/change-requests/job/PR-9722/52/ was also restarted from the Test stage and only ran the Hardware Medium test.   This time it passed.

brianjmurrell (approved):

PR #330, opened 2022-08-08
SRE-460 Build: Fix Junit error scanning

JohnMalmberg:
JUnit error scanning missing some errors.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -15,7 +15,8 @@
 // That PR should be landed with out deleting the PR branch.
 // Then a second PR submitted to comment out the @Library line, and when it
 // is landed, both PR branches can be deleted.
-//@Library(value="pipeline-lib@my_branch_name") _
+//@Library(value='pipeline-lib@my_branch_name') _
+@Library(value='pipeline-lib@sre-460') _
 
 pipeline {
     agent { label 'lightweight' }
@@ -67,7 +68,7 @@ pipeline {
                     }
                     // runTest handles SCM notification via stepResult
                 } // stage('grep JUnit results tests failure case')
-                stage('grep JUnit results tests error case') {
+                stage('grep JUnit results tests error case 1') {
                     agent {
                         docker {
                             image 'rockylinux:8'
@@ -83,7 +84,24 @@ pipeline {
                             failure_artifacts: env.STAGE_NAME
                     }
                     // runTest handles SCM notification via stepResult
-                } // stage('grep JUnit results tests error case')
+                } // stage('grep JUnit results tests error case 1')
+                stage('grep JUnit results tests error case 2') {
+                    agent {
+                        docker {
+                            image 'rockylinux:8'
+                            label 'docker_runner'
+                        }
+                    }
+                    steps {
+                        runTest script: '''set -ex
+                                           rm -f *.xml
+                                           echo "<errorDetails bla bla bla/>" > \
+                                             pipeline-test-error.xml''',
+                            junit_files: "*.xml non-exist*.xml",
+                            failure_artifacts: env.STAGE_NAME
+                    }
+                    // runTest handles SCM notification via stepResult
+                } // stage('grep JUnit results tests error case 2')
                 stage('publishToRepository RPM tests') {
                     when {
                         beforeAgent true
--- a/vars/runTest.groovy
+++ b/vars/runTest.groovy
@@ -50,9 +50,9 @@ def call(Map config = [:]) {
     // github expectations at the same time to also include any Matrix
     // environment variables.
 
-    def context = config.get('context', 'test/' + env.STAGE_NAME)
-    def description = config.get('description', env.STAGE_NAME)
-    def flow_name = config.get('flow_name', env.STAGE_NAME)
+    String context = config.get('context', 'test/' + env.STAGE_NAME)
+    String description = config.get('description', env.STAGE_NAME)
+    String flow_name = config.get('flow_name', env.STAGE_NAME)
 
     dir('install') {
         deleteDir()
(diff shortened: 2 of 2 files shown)


JohnMalmberg:
Build # failed because the @Library step was reverted.

vars/runTest.groovy
```diff
@@ -117,33 +117,33 @@ def call(Map config = [:]) {
...
-                       script: 'grep "<error " ' + junit_xml,
+                       script: 'grep "<error.* " ' + junit_xml,
                        returnStatus: true) == 0) {
-                    status = "FAILURE"
-                    println "Found at least one error in the Junit files."
+                    status = 'FAILURE'
+                    println 'Found at least one error in the Junit files.'
                 } else if (sh(label: 'Check junit xml files for failures',
-                              script: 'grep "<failure " ' + junit_xml,
+                              script: 'grep "<failure.* " ' + junit_xml,
```
brianjmurrell, line 139: I'm a little worried that this pattern is overly broad.  Maybe two different more specific patterns using `grep -e` is safer.

Jenkinsfile
```diff
@@ -90,7 +90,24 @@ pipeline {
...
+                    agent {
+                        docker {
+                            image 'rockylinux:8'
+                            label 'docker_runner'
+                        }
+                    }
+                    steps {
+                        runTest script: '''set -ex
+                                           rm -f *.xml
+                                           echo "<errorDetails bla bla bla/>" > \
```
brianjmurrell, line 104: Do you have a link to an example of this new format?  I'm just curious what's all changed.

brianjmurrell (changes requested):
Just the overly broad grep.

Jenkinsfile
JohnMalmberg: New to us, but not to the Junit plugin, it is in their test data.

https://github.com/jenkinsci/junit-plugin/blob/master/src/test/resources/hudson/tasks/junit/junitResult.xml
https://github.com/jenkinsci/junit-plugin/blob/master/src/test/resources/hudson/tasks/junit/junit-report-6700.xml

Still have not found a definitive spec for JUnit.   The definitive code for it is supposed to be in Apache Ant source modules.

vars/runTest.groovy
JohnMalmberg: We do not have a specification for JUnit that we can reference to determine what is safer.
`grep -E ".*<error( |Details>|StackTrace>)" junit.xml` is more precise in the matching what was missed.
Not finding any examples of equivalents suffixes for "failure".

vars/runTest.groovy
```diff
@@ -117,33 +117,33 @@ def call(Map config = [:]) {
...
-            def filesList = []
+            List filesList = []
             config['junit_files'].split().each {
                 filesList.addAll(findFiles(glob: it))
             }
             if (filesList) {
                 String junit_xml = filesList.collect{"'" + it + "'"}.join(' ')
                 if (sh(label: 'Check junit xml files for errors',
-                       script: 'grep "<error " ' + junit_xml,
+                       script: 'grep -E "<error( |Details>|StackTrace>)" ' + junit_xml,
```
brianjmurrell, line 134: Just to be pedantic, this could be further factored down to:

```suggestion
                       script: 'grep -E "<error( |(Details|StackTrace)>)" ' + junit_xml,
```

Jenkinsfile
brianjmurrell: Do you have a link from our tests though?  It's difficult to meaningfully review the changes here without the data on hand that the changes are being made to accomodate.

vars/runTest.groovy
JohnMalmberg: I am sticking with the shorter line.

Jenkinsfile
JohnMalmberg: No links are available, which made this a bit harder to diagnose.
I had to look at the internal Jenkins job storage to find the junit XML file.

vars/cloverReportPublish.groovy
```diff
@@ -83,7 +87,7 @@ def call(Map config = [:]) {
                                     statementCoverage: 80])])
 
   if (cb_result != currentBuild.result) {
-    println "The CloverPublisher plugin changed result to " +
+    println 'The CloverPublisher plugin changed result to ' +
             "${currentBuild.result}."
```
brianjmurrell, line 91: In groovy scripts we usually include the brackets for function calls:

```suggestion
    println('The CloverPublisher plugin changed result to ' +
            "${currentBuild.result}.")
```

vars/functionalTestPostV2.groovy
```diff
@@ -35,6 +34,9 @@ void call(Map config = [:]) {
 
     junit(testResults: junit_results)
 
+    // Save these temporarily
+    archiveArtifacts(artifacts: junit_results)
+
```
brianjmurrell, line 39: So you are going to remove this before landing?  I will -1 here assuming that is a yes to prevent an accidental landing of it.

vars/runTest.groovy
```diff
@@ -108,7 +108,7 @@ def call(Map config = [:]) {
     rc = sh(script: script, label: flow_name, returnStatus: true)
 
     if (cb_result != currentBuild.result) {
-      println 'The runTest script changed build result to ' +
+      println 'Some other stage changed the currentBuild result to ' +
               "${currentBuild.result}."
```
brianjmurrell, line 112:
```suggestion
      println('Some other stage changed the currentBuild result to ' +
              "${currentBuild.result}.")
```

vars/runTest.groovy
```diff
@@ -117,19 +117,31 @@ def call(Map config = [:]) {
     // https://issues.jenkins-ci.org/browse/JENKINS-39203
     // Once that is fixed all of the below should be pushed up into the
     // Jenkinsfile post { stable/unstable/failure/etc. }
+    
+    // Somehow we now have two different names for storing the expected
+    // junit result file names.
```
brianjmurrell, line 122: Where would this be coming from?  _testResults_ is not even a documented parameter.  Probably whatever caller is using that parameter rather than _junit_files_ needs to be fixed.  We shouldn't add hacks here for callers using the wrong parameter.

vars/runTest.groovy
```diff
@@ -117,19 +117,31 @@ def call(Map config = [:]) {
...
+        if (junit_results) {
             List filesList = []
-            config['junit_files'].split().each {
-                filesList.addAll(findFiles(glob: it))
+            junit_results.split().each { junitfile ->
+                filesList.addAll(findFiles(glob: junitfile))
             }
             if (filesList) {
                 String junit_xml = filesList.collect{"'" + it + "'"}.join(' ')
+                println("checking ${junit_xml}")
```
brianjmurrell, line 144: Debug to be removed before landing?

brianjmurrell (changes requested):

JohnMalmberg:
https://build.hpdd.intel.com/job/daos-stack/job/daos/view/change-requests/job/PR-9482/ is testing the Bullseye code changes.
local build will fail as soon as I remove the @Library line from the Jenkinsfile.

JohnMalmberg:
Current build, build 40 will fail because the Jenkinsfile now needs the new pipeline lib for the junit tests to work.

https://build.hpdd.intel.com/job/daos-stack/job/pipeline-lib/view/change-requests/job/PR-330/39/ passed all tests and https://build.hpdd.intel.com/job/daos-stack/job/daos/view/change-requests/job/PR-9482/71/ tested bullseye changes in this library.

JohnMalmberg:
https://build.hpdd.intel.com/job/daos-stack/job/daos/view/change-requests/job/PR-9482/70/artifact/stage_status/ tested that the junit checks are now working.

brianjmurrell (approved):

PR #344, opened 2022-09-26
SRE-519 Build: Fix junit failure lookup

JohnMalmberg:
Junit failures can be "<failure " or "<failure>".

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -17,7 +17,6 @@
 // Then a second PR submitted to comment out the @Library line, and when it
 // is landed, both PR branches can be deleted.
 //@Library(value='pipeline-lib@my_branch_name') _
-@Library(value='pipeline-lib@sre-519') _
 
 String test_branch(String target) {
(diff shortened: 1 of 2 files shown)


JohnMalmberg:
Removed @Library line, so the current run will fail the new checks.

Jenkinsfile
```diff
@@ -68,7 +68,7 @@ pipeline {
                     steps {
                         runTest script: '''set -ex
                                            rm -f *.xml
-                                           echo "<failure bla bla bla/>" > \
+                                           echo "<failure> bla bla bla</failure>" > \
```
brianjmurrell, line 71: Shouldn't this test be expanded to test for both types of failure reporting now that we know it can be in multiple formats?

brianjmurrell (approved):

Jenkinsfile
JohnMalmberg: I considered that, but given that the older test had not been failing, just covering the newly discovered test was good enough instead of adding a new test stage.

PR #347, opened 2022-11-04
DAOS-12023 ci: Add HW Medium Provider stages

phender:
Adding support for the new Functional Hardware Medium <name> Provider stages.  These new functional test stages require a fixed test provider setting and will run any appropraitely sized test with the 'provider' tag.

### DIFF
--- a/vars/parseStageInfo.groovy
+++ b/vars/parseStageInfo.groovy
@@ -214,8 +214,19 @@ void call(Map config = [:]) {
         result['pragma_suffix'] = '-hw-small'
       } else if (stage_name.contains('Medium')) {
         result['node_count'] = 5
-        cluster_size = 'hw,medium'
+        cluster_size = 'hw,medium,-provider'
         result['pragma_suffix'] = '-hw-medium'
+        if (stage_name.contains('Provider')) {
+          cluster_size = 'hw,medium,provider'
+          if (stage_name.contains('Verbs')) {
+            result['pragma_suffix'] += '-verbs-provider'
+            ftest_arg_provider = 'ofi+verbs'
+          }
+          elif (stage_name.contains('UCX')) {
+            result['pragma_suffix'] += '-ucx-provider'
+            ftest_arg_provider = 'ucx+dc_x'
+          }
+        }
       }
     }
     if (stage_name.contains('with Valgrind')) {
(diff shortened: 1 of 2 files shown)


vars/skipStage.groovy
```diff
@@ -523,6 +523,12 @@ boolean call(Map config = [:]) {
         case 'Functional_Hardware_Medium':
         case 'Functional Hardware Medium':
             return skip_ftest_hw('medium', target_branch)
+        case 'Functional_Hardware_Medium_Verbs_Provider':
```
brianjmurrell, line 526: Not necessary.  Was only there for other stages for hysterical raisins.

vars/skipStage.groovy
```diff
@@ -523,6 +523,12 @@ boolean call(Map config = [:]) {
...
+        case 'Functional Hardware Medium Verbs Provider':
+            return skip_ftest_hw('medium-verbs-provider', target_branch)
+        case 'Functional_Hardware_Medium_UCX_Provider':
```
brianjmurrell, line 529: Ditto.

brianjmurrell (commented):
I need to see the correlating daos repo changes before being able to +1.

vars/skipStage.groovy
phender: Removed

vars/skipStage.groovy
phender: Removed

brianjmurrell (approved):

vars/parseStageInfo.groovy
```diff
@@ -214,8 +214,19 @@ void call(Map config = [:]) {
...
+          cluster_size = 'hw,medium,provider'
+          if (stage_name.contains('Verbs')) {
+            result['pragma_suffix'] += '-verbs-provider'
+            ftest_arg_provider = 'ofi+verbs'
+          }
+          elif (stage_name.contains('UCX')) {
+            result['pragma_suffix'] += '-ucx-provider'
+            ftest_arg_provider = 'ucx+dc_x'
+          }
+        }
```
daltonbohning, line 229: A few points
1. Is the syntax `elif` or `else if`? I see `else if` above in this file.
2. Do we want an `else` just in case?
3. Just a thought - I've always thought the Skip-* pragmas would be better served if we used some sort of "subset" naming convention. I.e.
    - `hw-medium`
    - `hw-medium-provider-verbs`
    - `hw-medium-provider-ucx`
    And skip pragmas:
    - `Skip-test-func-hw-medium` - Skips all `hw-medium*`
    - `Skip-test-func-hw-medium-provider` - Skips all `hw-medium-provider*`
    - `Skip-test-func-hw-medium-provider-ucx` - Skips all `hw-medium-provider-ucx*`
    But for this to really shine, we'd want to refactor the existing pragmas. A benefit is the logic for skipping groups of stages becomes MUCH simpler. Not asking you to do this - just something to think about.

brianjmurrell: > A few points
> 
> 1. Is the syntax `elif` or `else if`? I see `else if` above in this file.

Good  catch @daltonbohning.  The syntax is indeed `else if`.

brianjmurrell (changes requested):
Just for the `elif`.

vars/parseStageInfo.groovy
phender: Fixed, thanks guys.

daltonbohning (approved):

PR #350, opened 2022-11-28
SRE-645 test: Fix detecting 'weekly' branches

phender:
Updating the regular expression to fix indetifying branches run on a weekly cadence.

### DIFF
--- a/vars/testBranchRE.groovy
+++ b/vars/testBranchRE.groovy
@@ -5,5 +5,5 @@
    */
 
 def call() {
-    return '^[-0-9A-Za-z]+-testing.*'
+    return '^(weekly|provider|soak)-testing(-[a-z]+|)(-[0-9.]+|)'
 }
--- a/vars/testBranchRE.groovy
+++ b/vars/testBranchRE.groovy
@@ -5,5 +5,5 @@
    */
 
 def call() {
-    return '^(weekly|provider|soak)-testing(-[a-z]+|)(-[0-9.]+|)'
+    return '^[-0-9A-Za-z.]+-testing(-(tcp|ucx))?(-2.2)?$'
 }
--- a/vars/testBranchRE.groovy
+++ b/vars/testBranchRE.groovy
@@ -5,5 +5,5 @@
    */
 
 def call() {
-    return '^[-0-9A-Za-z]+-testing'
+    return '^[-0-9A-Za-z]+-testing.*'
 }


mjean308 (approved):

vars/testBranchRE.groovy
```diff
@@ -5,5 +5,5 @@
    */
 
 def call() {
-    return '^[-0-9A-Za-z]+-testing'
+    return '^[-0-9A-Za-z]+-testing.*'
```
brianjmurrell, line 8: The RE is getting pretty loose and could inadvertently match things we don’t want it to.  I thought I had recalled the proposal was to just end any such branch with `-testing` to try to keep the pattern open but fairly strict.

I realize the toothpaste is out of the tube on these existing branches, so I would advocate adding specific matches for them but then adhering to a more strict pattern for all future branches that want this behaviour.

phender:
Manual testing of the regex:
```
groovy:000> branch = "master"
===> master
groovy:000> branch.matches('^(weekly|provider|soak)-testing(-[a-z]+|)(-[0-9.]+|)')
===> false
groovy:000> branch = "release/2.2"
===> release/2.2
groovy:000> branch.matches('^(weekly|provider|soak)-testing(-[a-z]+|)(-[0-9.]+|)')
===> false
groovy:000> branch = "weekly-testing"
===> weekly-testing
groovy:000> branch.matches('^(weekly|provider|soak)-testing(-[a-z]+|)(-[0-9.]+|)')
===> true
groovy:000> branch = "weekly-testing-2.2"
===> weekly-testing-2.2
groovy:000> branch.matches('^(weekly|provider|soak)-testing(-[a-z]+|)(-[0-9.]+|)')
===> true
groovy:000> branch = "provider-testing-tcp"
===> provider-testing-tcp
groovy:000> branch.matches('^(weekly|provider|soak)-testing(-[a-z]+|)(-[0-9.]+|)')
===> true
groovy:000> branch = "provider-testing-tcp-2.2"
===> provider-testing-tcp-2.2
groovy:000> branch.matches('^(weekly|provider|soak)-testing(-[a-z]+|)(-[0-9.]+|)')
===> true
groovy:000> branch = "provider-testing-ucx"
===> provider-testing-ucx
groovy:000> branch.matches('^(weekly|provider|soak)-testing(-[a-z]+|)(-[0-9.]+|)')
===> true
groovy:000> branch = "provider-testing-ucx-2.2"
===> provider-testing-ucx-2.2
groovy:000> branch.matches('^(weekly|provider|soak)-testing(-[a-z]+|)(-[0-9.]+|)')
===> true
groovy:000> branch = "soak-testing"
===> soak-testing
groovy:000> branch.matches('^(weekly|provider|soak)-testing(-[a-z]+|)(-[0-9.]+|)')
===> true
groovy:000> branch = "yournamehere/DAOS-123456"
===> yournamehere/DAOS-123456
groovy:000> branch.matches('^(weekly|provider|soak)-testing(-[a-z]+|)(-[0-9.]+|)')
===> false

```

vars/testBranchRE.groovy
brianjmurrell: > Would this be sufficient?
> 
> ```
>     return '^(weekly|provider|soak)-testing.*'
> ```

The problem is that negates the entire purpose of using a pattern here which is to avoid having to keep pushing PRs, adding to the pattern when new `*-testing` branches are needed.  We wanted to make one final change here that gives us the flexibility to add new testing branches in the future to cover whatever testing whims may come up.  Imagine, a `daily-testing` requirement comes up, or a 'foo-testing` requirement, etc.  We want to be able to add those without churning a pipeline-lib PR just to update the RE.

> The problem is today we need to match the following branch names:
> 
> ```
> weekly-testing
> weekly-testing-2.2
> provider-testing-tcp
> provider-testing-tcp-2.2
> provider-testing-ucx
> provider-testing-ucx-2.2
> soak-testing
> ```

I am suggesting that those are anomolies/exceptions to the established `*-testing` pattern and be grandfathered in, and that new branches follow the `*-testing` pattern.

> Where eventually `2.2` will be replaced by `2.4`, `2.6`, etc.

Which, if you accept my proposal would be changed to

```
weekly-2.4-testing
provider-tcp-2.4-testing
provider-ucx-2.4-testing
```

As they become required.
 
> If a more specific match is needed we could use:
> 
> ```
>     retrun '^(weekly|provider|soak)-testing(-[a-z]+|)(-[0-9.]+|)'
> ```

Again, that defeats the goal of using a pattern.

mjean308 (approved):

phender:
Updated manual testing https://daosio.atlassian.net/browse/SRE-645?focusedCommentId=107896

brianjmurrell (approved):

PR #355, opened 2022-12-12
DAOS-11624 test: Adding support for the Test-storage-tier pragma

phender:
Adding support for passing the Test-storage-tier build parameter or commit pragma to launch.py via its --storage_tier argument.

### DIFF
--- a/vars/parseStageInfo.groovy
+++ b/vars/parseStageInfo.groovy
@@ -70,6 +70,12 @@ String get_commit_pragma_tags(String pragma_suffix) {
   return pragma_tag
 }
 
+String get_param_or_pragma(param, pragma, arg, predefined) {
+    String res = params.$param ?: cachedCommitPragma(
+      pragma + result['pragma_suffix'], cachedCommitPragma(pragma, predefined))
+
+    return res ? ' ' + arg + '= ' + res : ''
+}
 
 /* groovylint-disable-next-line MethodSize */
 void call(Map config = [:]) {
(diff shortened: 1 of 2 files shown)


vars/parseStageInfo.groovy
```diff
@@ -311,6 +316,9 @@ void call(Map config = [:]) {
     if (ftest_arg_provider) {
       result['ftest_arg'] += ' --provider=' +  ftest_arg_provider
     }
+    if (ftest_arg_storage_tier) {
+      result['ftest_arg'] += ' --storage_tier=' +  ftest_arg_storage_tier
+    }
```
brianjmurrell, line 321: Worth mentioning that this could be made more succinct with:

```suggestion
    result['ftest_arg'] += ftest_arg_storage_tier ? ' --storage_tier=' +  ftest_arg_storage_tier : ''
```

Or that a function could be written to make all of this more DRY, now that it's evident that it's proliferating.  I.e.

```groovy

String get_param_or_pragma(param, pragma, arg) {
    String res = params.$param ?: cachedCommitPragma(
      pragma + result['pragma_suffix'], cachedCommitPragma(pragma, null))

    return res ? '  --' + arg + '= ' + res : ''
}

result['ftest_arg'] += get_param_or_pragma('TestStorageTier', 'Test-storage-tier', '--storage_tier')
```

And of course using the above for all of the other param-or-pragma values that are built into `result['ftest_arg']`.

brianjmurrell (dismissed):

vars/parseStageInfo.groovy
phender: Thanks.  I've added this change.

brianjmurrell (dismissed):

PR #362, opened 2023-01-18
SRE-1682 Build: Report duration of steps

JohnMalmberg:
Add code to pipeline-lib steps to report how long a step took in minutes.

Signed-off-by: John E. Malmberg <john.e.malmberg@intel.com>


Jenkinsfile
```diff
@@ -63,12 +114,13 @@ pipeline {
...
                     }
                     steps {
-                        runTest script: '''set -ex
-                                           rm -f *.xml
-                                           echo "<failure> bla bla bla</failure>" > \
-                                             pipeline-test-failure.xml''',
-                            junit_files: '*.xml non-exist*.xml',
-                            failure_artifacts: env.STAGE_NAME
+                        job_step_update(
+                            runTest(script: '''set -ex
```
brianjmurrell, line 118: Do these **need** the brackets?  We have been using a coding style where we treat these as _step_ names rather than function names in `Jenkinsfile`, and only treat them as function names in pure groovy scripts.

vars/buildRpm.groovy
```diff
@@ -65,10 +66,15 @@ void call(Map config = [:]) {
         error_stage_result = 'UNSTABLE'
         error_build_result = 'SUCCESS'
     }
+    Map runData = ['result': 'FAILURE']
     catchError(stageResult: error_stage_result,
                buildResult: error_build_result) {
         // flow_name used as the label for this step to allow log lookup.
         sh(label: config.get('flow_name', env.STAGE_NAME),
            script: "${env_vars} " + build_script)
+        runData['result'] = 'SUCCESS'
```
brianjmurrell, line 75: If I'm understanding correctly, we are setting a result of `SUCCESS` here even if the stage failed, but was asked to convert to `UNSTABLE`, yes?

If so, I'm not convinced that this is the correct thing to do.  While it's what we need to do to satisfy the developer requirement that package build failures don't stop the pipeline (which I don't think I agree with since a failure to build the EL8 package for example tanks all of the functional testing anyway), it seems like it's unnecessarily masking the stage's real result in our "reliability" data.

IOW, we are tainting the status to satisfy a functional requirement.  Should we also be tainting it for statistical purposes?

It might be useful for example, while looking at the data to know that the EL8 RPM build failed when we see EL8 functional stages failing.  That in fact convinces me further that we should be reporting the real stage result here, not what it had to be to satisfy the functional requirements.

vars/durationMinutes.groovy
```diff
@@ -0,0 +1,24 @@
...
+
+  /**
+   * durationMinutes step method
+   *
+   * This does not create the file, it is so the methods
+   * that create / fetch / read the STAGE status file all use
+   * the same name and it is properly URLEncoded.
+   *
+   * It does make sure that the directory for the stage
+   * status file exists.
```
brianjmurrell, line 11: How is this happening?  Not seeing any directory creation here.

vars/durationMinutes.groovy
```diff
@@ -0,0 +1,24 @@
+// vars/durationMinutes.groovy
+
+  /**
+   * durationMinutes step method
+   *
+   * This does not create the file, it is so the methods
+   * that create / fetch / read the STAGE status file all use
+   * the same name and it is properly URLEncoded.
```
brianjmurrell, line 8: Is this copy'n'pasted from somewhere?  It doesn't seem relevant to this function.

vars/durationMinutes.groovy
```diff
@@ -0,0 +1,24 @@
...
+   * @param startDate, Date object with start time.
+   * @param endDate, Date object with end time, default current date time.
+   * returns: Duration time in minutes.
+   */
+
+int call(Date startDate, Date endDate=null) {
+    if (endDate == null) {
+        endDate = new Date()
+    }
+    int delta = (endDate.getTime() - startDate.getTime())/1000/60
```
brianjmurrell, line 22: It's a pity that we are rounding this up to minutes.  I thought our choice to use minutes was because it was somehow easier.

IMO, input data should always be as precise as it can be, and it's up to the consumer to decide what level of precision they want to distill the data down to.

We might one day come to regret that we don't know how many seconds a stage is taking.

vars/pythonBanditCheck.groovy
```diff
@@ -31,24 +31,25 @@
...
   }
-  def bandit_junit = config.get('junit_files', 'bandit.xml')
-  runTest script: bandit_script,
-          junit_files: bandit_junit,
-          ignore_failure: true
+  String bandit_junit = config.get('junit_files', 'bandit.xml')
+  Map runData = runTest script: bandit_script,
+                junit_files: bandit_junit,
+                ignore_failure: true
+  return runData
```
brianjmurrell, line 54: FWIW, could be simplified into:
```suggestion
  return runTest script: bandit_script,
                junit_files: bandit_junit,
                ignore_failure: true
```

and avoid the intermediate, single-use variable.

brianjmurrell (changes requested):
Only real sticking issue is the masking of the RPM stage results.  I'm willing to be convinced that I am wrong, but as of this moment, it seems wrong to mask that result.

vars/pythonBanditCheck.groovy
JohnMalmberg: I generally prefer the intermediate variable notation, but will change it.

vars/buildRpm.groovy
brianjmurrell: Ah yes.  I was forgetting that processing in the `catchError()` would end on any step that throws an error.  Thanks for the explanation.

brianjmurrell (approved):

PR #363, opened 2023-01-19
SRE-1682 Build: Enable malware scanning

JohnMalmberg:
Enable malware scanning by default

Signed-off-by: John E. Malmberg <john.e.malmberg@intel.com>

### DIFF
--- a/vars/skipStage.groovy
+++ b/vars/skipStage.groovy
@@ -105,8 +105,10 @@ boolean skip_scan_rpms(String distro, String target_branch) {
     return already_passed() ||
            target_branch == 'weekly-testing' ||
            rpmTestVersion() != '' ||
-           skip_stage_pragma('scan-rpms', 'false') ||
+           skip_stage_pragma('scan-rpms') ||
            skip_stage_pragma('scan-' + distro + '-rpms') ||
+           (distro == 'centos7' &&
+            (!paramsValue('CI_SCAN_RPMS_el7_TEST', true)) ||
            docOnlyChange(target_branch) ||
            quickFunctional()
 }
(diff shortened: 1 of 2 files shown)


JohnMalmberg:
Weekly testing job failure does not appear related to this pipeline-lib change

vars/skipStage.groovy
```diff
@@ -104,10 +104,8 @@ boolean skip_build_on_landing_branch(String target_branch) {
 boolean skip_scan_rpms(String distro, String target_branch) {
     return already_passed() ||
            target_branch == 'weekly-testing' ||
-           skip_stage_pragma('scan-rpms', 'true') ||
-           (distro == 'centos-7' &&
-            (!paramsValue('CI_SCAN_RPMS_el7_TEST', true)) ||
-            skip_stage_pragma('scan-centos-rpms')) ||
+           rpmTestVersion() != '' ||
+           skip_stage_pragma('scan-rpms', 'false') ||
```
brianjmurrell, line 108: `false` is already the default value here.  We only supply a second argument to `skip_stage_pragma()` if we are looking for non-default behaviour, for example, `true` when trying to detect if a usually-skipped stage is being requested.

My concern here with supplying `false` here would be some kind of future change in the default value of that argument that is assuming that every caller that wants default behaviour is not supplying a value for the second argument.  We'd have this one outlier that operates differently with the new change and then a hunt to figure out why.

vars/skipStage.groovy
```diff
@@ -104,10 +104,8 @@ boolean skip_build_on_landing_branch(String target_branch) {
 boolean skip_scan_rpms(String distro, String target_branch) {
     return already_passed() ||
            target_branch == 'weekly-testing' ||
-           skip_stage_pragma('scan-rpms', 'true') ||
-           (distro == 'centos-7' &&
-            (!paramsValue('CI_SCAN_RPMS_el7_TEST', true)) ||
-            skip_stage_pragma('scan-centos-rpms')) ||
```
brianjmurrell, line 110: `Skip-scan-centos-rpms` is documented on https://daosio.atlassian.net/wiki/spaces/CI/pages/1389759093/Commit+Pragmas.  So there are two issues.  At minimum, that page will need to be updated, but more concerning is removing existing behaviour that people might be familiar with.  If you are going to remove this, in addition to updating the wiki page, you should probably send out an e-mail informing everyone of the removal of that option.  Probably most won't care but we have done our due diligence in the matter.

But more importantly is the removing of the `CI_SCAN_RPMS_el7_TEST` parameter check.  While that parameter is not currently in master, it is in some historic commit of master (and other branches probably) and by removing this here, you are breaking that functionality if somebody were to try to build such a historic commit and use that parameter.  We should be striving to maintain backward compatibility.  If we wanted to be able to remove these kinds of things eventually, we need to have a statement of backward compatibility guarantee (i.e. how far back in history we will support parameters and pragmas) and then track the end-of-use of options that we want to remove in the future and only remove them once that guarantee has expired.  Just trying to always maintain backward compatibility (where possible of course) seems a lot easier.

JohnMalmberg:
https://build.hpdd.intel.com/job/daos-stack/job/daos/view/change-requests/job/PR-11282/ is master using this PR.

brianjmurrell (approved):

PR #366, opened 2023-02-28
SRE-566 Build: Fix up git command access

JohnMalmberg:
Fix the git commands for looking up commit pragmas to use the GIT_CHECKOUT_DIR environment variable if it exists and the directory it references exists.


vars/archiveArtifacts.groovy
```diff
@@ -1,28 +1,15 @@
...
-                  exit 0\n'''
-    int rc = 0
-    rc = sh(script: script, label: env.STAGE_NAME + '_archiveArtifacts',
-            returnStatus: true)
-    if (rc != 0) {
-        config['allowEmptyArchive'] = true;
+//groovylint-disable DuplicateStringLiteral
+void call(Map config = [:]) {
+    if (env.NO_CI_TESTING != 'true' ||
+            commitPragma('Skip-Test') == 'true') {
```
brianjmurrell, line 11: Any reason not to use [`cachedCommitPragma()`](https://github.com/daos-stack/pipeline-lib/blob/master/vars/cachedCommitPragma.groovy) here?  At least for consistency?

Generally speaking `commitPragma()` should only be used internally and all external users should use `cachedCommitPragma()`.

vars/isPr.groovy
```diff
@@ -10,7 +10,7 @@
  * Method to return whether job is a PR or not
  */
 
-Boolean call() {
+boolean call() {
```
brianjmurrell, line 13: Why the change here?

vars/junit.groovy
```diff
@@ -4,32 +4,18 @@
...
-                  fi
-                  exit 0\n'''
-    int rc = 0
-    rc = sh(script: script, label: env.STAGE_NAME + '_junit',
-            returnStatus: true)
-    if (rc != 0) {
+//groovylint-disable DuplicateStringLiteral
+void call(Map config = [:]) {
+    if (env.NO_CI_TESTING != 'true' ||
+            commitPragma('Skip-Test') == 'true') {
```
brianjmurrell, line 18: Ditto re: `cachedCommitPragma()`.

vars/pragmasToEnv.groovy
```diff
@@ -7,14 +7,15 @@
  * pragmasToEnv variable
  */
 
-String call(String commit_message) {
+Map call(String commit_message) {
```
brianjmurrell, line 10: I'm a little leery of this change.   While it is true that what is being returned is a `Map` are we sure that we are not depending on the cast from `Map` to `String` in the `return`ing of the value?

vars/publishValgrind.groovy
```diff
@@ -1,28 +1,15 @@
...
-                  exit 0\n'''
-    int rc = 0
-    rc = sh(script: script, label: env.STAGE_NAME + '_publishValgrind',
-            returnStatus: true)
-    if (rc != 0) {
-        config['failBuildOnMissingReports'] = false;
+//groovylint-disable DuplicateStringLiteral
+void call(Map config = [:]) {
+    if (env.NO_CI_TESTING != 'true' ||
+            commitPragma('Skip-Test') == 'true') {
```
brianjmurrell, line 11: Ditto `cachedCommitPragma()`.

vars/pragmasToEnv.groovy
brianjmurrell: > That method has always been only returning a Map not a String.

This is not necessarily true.  Groovy could have been casting the `Map` to a `String` before returning the value, given that the function was declared to return a `String`.

> Apparently the return statement will return object with out groovy checking to see if the type is compatible with the declaration.

Is this actually documented somewhere or just your assumption based on this not having caused any problems so far?  I would say it's more likely that Groovy is doing the casting and returning the casted `String` value rather than it's returning values without concern of the function's declared return type.

> The usage of "${env.pragmas}" does conversion of the Map to a string.

Again, or is it Groovy doing the casting in the `return`?

Ultimately, it may not matter whether the casting is being done on value before `return` or before assignment of the returned `Map`.  It just makes me leery to do this without knowing who all the callers are and what they are doing with the returned value.

Did any of your test runs with this change in it use an commit pragams?  If so, and if they were processed as expected this is probably safe to do.  Might be worth a quick check to make sure there are no consumers of this function other than code paths that are always exercised.

I'm ultimately just trying to avoid us landing an embarrassing time bomb that goes off some time in our future.

JohnMalmberg: The Jenkins groovy scripting console confirms that Groovy is doing a Map to String conversion.  It is using a "{foo=bar}" value instead of the "[foo:bar]" conversion that would be done from string interpolation.

Reverting that change until this can be investigated later, as this is slightly different code than is in the DAOS Jenkinsfile for setting the env.pragmas value.

brianjmurrell (approved):

PR #367, opened 2023-03-01
SRE-1729 Build: Fail Unit Test on Valgrind errors

JohnMalmberg:
Check for Valgrind errors after running a Unit test.

### DIFF
--- a/vars/junitSimpleReport.groovy
+++ b/vars/junitSimpleReport.groovy
@@ -30,14 +30,13 @@ void call(Map config = [:]) {
     String jname = config.get('name', simple)
     String jsuite = config.get('suite', sanitizedStageName())
     String jfile = config.get('file', jsuite + '_results.xml')
-    String je = config.get('errors', zero)
-    String jf = config.get('fails', zero)
+    int je = config.get('errors', zero)
+    int jf = config.get('fails', zero)
     String message = config.get('message', '')
     String testdata = config.get('testdata')
     boolean ignoreFailure = config.get('ignoreFailure', false)
     String tresult
     // Enforce consistency
-    println("#### jsr1: jf = ${jf} je=${je} tresult=${tresult} config=${config}")
     if (jf > zero) {
         tresult = 'failure'
         jf = one
@@ -49,7 +48,6 @@ void call(Map config = [:]) {
             je = one
         }
     }
-    println("#### jsr2: jf = ${jf} je=${je} tresult=${tresult}")
     String xml = """
 <testsuite skip="0" failures="${jf}" errors="${je}" tests="1" name="${jsuite}">
   <testcase name="${jname}" classname="${jclass}">
--- a/vars/junitSimpleReport.groovy
+++ b/vars/junitSimpleReport.groovy
@@ -34,8 +34,8 @@ void call(Map config = [:]) {
     String jf = config.get('fails', zero)
     String message = config.get('message', '')
     String testdata = config.get('testdata', '')
-    boolean ignoreFailure = config.get('ingnoreFailure', false)
-    String tresult = None
+    boolean ignoreFailure = config.get('ignoreFailure', false)
+    String tresult
     // Enforce consistency
     if (jf > zero) {
         tresult = 'failure'
--- a/vars/junitSimpleReport.groovy
+++ b/vars/junitSimpleReport.groovy
@@ -25,7 +25,7 @@
 void call(Map config = [:]) {
     int zero = 0
     int one = 1
-    String simple = 'simple'
+    simple = 'simple'
     String jclass = config.get('class', simple)
     String jname = config.get('name', simple)
     String jsuite = config.get('suite', sanitizedStageName())
@@ -63,9 +63,5 @@ void call(Map config = [:]) {
 </testsuite>
 '''
     writeFile(file: jfile, text: "${xml}")
-    String cbr = currentBuild.result
     junit testResults: jfile
-    if (cbr != currentBuild.result) {
-        println "The junit plugin changed result to ${currentBuild.result}."
-    }
 }
--- a/vars/junitSimpleReport.groovy
+++ b/vars/junitSimpleReport.groovy
@@ -34,6 +34,7 @@ void call(Map config = [:]) {
     String jf = config.get('fails', zero)
     String message = config.get('message', '')
     String testdata = config.get('testdata', '')
+    boolean ignoreFailure = config.get('ingnoreFailure', false)
     String tresult = ''
     // Enforce consistency
     if (jf > zero) {
@@ -63,5 +64,12 @@ void call(Map config = [:]) {
 </testsuite>
 '''
     writeFile(file: jfile, text: "${xml}")
-    junit testResults: jfile
+    // groovylint-disable-next-line NoDouble
+    double healthScale = 1.0
+    if (ignoreFailure) {
+        healthScale = 0.0
+    }
+
+    junit testResults: jfile,
+          healthScaleFactor: healthScale
 }
--- a/vars/junitSimpleReport.groovy
+++ b/vars/junitSimpleReport.groovy
@@ -25,7 +25,7 @@
 void call(Map config = [:]) {
     int zero = 0
     int one = 1
-    simple = 'simple'
+    String simple = 'simple'
     String jclass = config.get('class', simple)
     String jname = config.get('name', simple)
     String jsuite = config.get('suite', sanitizedStageName())
--- a/vars/junitSimpleReport.groovy
+++ b/vars/junitSimpleReport.groovy
@@ -64,6 +64,8 @@ void call(Map config = [:]) {
 </testsuite>
 '''
     writeFile(file: jfile, text: "${xml}")
+    // Debug why this is not showing up correctly in Junit output.
+    archiveArtifacts artifacts: jfile
     // groovylint-disable-next-line NoDouble
     double healthScale = 1.0
     if (ignoreFailure) {
--- a/vars/junitSimpleReport.groovy
+++ b/vars/junitSimpleReport.groovy
@@ -1,0 +1,71 @@
+// vars/junitSimpleReport.groovy
+
+  /**
+   * junitSimpleReport step method
+   *
+   * Create and post a simple JUnit test report for a test suite with a
+   * single test.
+   *
+   * @param config Map of parameters passed
+   *
+   * config['class']               Class of test, default 'simple'
+   * config['errors']              count of errors, default 0
+   * config['fails']               count of failures, default 0
+   *                               Only one of errors or fails can be 1.
+   * config['file']                JUnit file name.
+   *                               Default suite + "_results.xml"
+   * config['message']             Error or failure message.  Default ''
+   * config['name']                JUnit testcase name.
+   *                               Default 'simple'
+   * config['suite']               Test suite to for JUnit file.
+   *                               Default sanitizedStageName()
+   * config['testdata']            Test data Default ''.
+   */
+
+void call(Map config = [:]) {
+    int zero = 0
+    int one = 1
+    simple = 'simple'
+    String jclass = config.get('class', simple)
+    String jname = config.get('name', simple)
+    String jsuite = config.get('suite', sanitizedStageName())
+    String jfile = config.get('file', jsuite + '_results.xml')
+    String je = config.get('errors', zero)
+    String jf = config.get('fails', zero)
+    String message = config.get('message', '')
+    String testdata = config.get('testdata', '')
+    String tresult = ''
+    // Enforce consistency
+    if (jf > zero) {
+        tresult = 'failure'
+        jf = one
+    }
+    if (je > zero) {
+        tresult = 'error'
+        jf = zero
+        if (je > one) {
+            je = one
+        }
+    }
+    String xml = """
+<testsuite skip="0" failures="${jf}" errors="${je}" tests="1" name="${jsuite}">
+  <testcase name="${jname}" classname="${jclass}">
+"""
+    if (tresult) {
+        xml += "    <${tresult} message=\"${message}\" type=\"$tresult\">\n"
+        if (testdata) {
+            xml += "      <![CDATA[ ${testdata} ]]>\n"
+        }
+        xml += "    </${tresult}>"
+    }
+    xml += '''
+  </testcase>
+</testsuite>
+'''
+    writeFile(file: jfile, text: "${xml}")
+    String cbr = currentBuild.result
+    junit testResults: jfile
+    if (cbr != currentBuild.result) {
+        println "The junit plugin changed result to ${currentBuild.result}."
+    }
+}
--- a/vars/junitSimpleReport.groovy
+++ b/vars/junitSimpleReport.groovy
@@ -37,6 +37,7 @@ void call(Map config = [:]) {
     boolean ignoreFailure = config.get('ignoreFailure', false)
     String tresult
     // Enforce consistency
+    println("#### jsr1: jf = ${jf} je=${je} tresult=${tresult} config=${config}")
     if (jf > zero) {
         tresult = 'failure'
         jf = one
@@ -48,6 +49,7 @@ void call(Map config = [:]) {
             je = one
         }
     }
+    println("#### jsr2: jf = ${jf} je=${je} tresult=${tresult}")
     String xml = """
 <testsuite skip="0" failures="${jf}" errors="${je}" tests="1" name="${jsuite}">
   <testcase name="${jname}" classname="${jclass}">
--- a/vars/junitSimpleReport.groovy
+++ b/vars/junitSimpleReport.groovy
@@ -35,7 +35,7 @@ void call(Map config = [:]) {
     String message = config.get('message', '')
     String testdata = config.get('testdata', '')
     boolean ignoreFailure = config.get('ingnoreFailure', false)
-    String tresult = ''
+    String tresult = None
     // Enforce consistency
     if (jf > zero) {
         tresult = 'failure'
@@ -64,7 +64,6 @@ void call(Map config = [:]) {
 </testsuite>
 '''
     writeFile(file: jfile, text: "${xml}")
-    // Debug why this is not showing up correctly in Junit output.
     archiveArtifacts artifacts: jfile
     // groovylint-disable-next-line NoDouble
     double healthScale = 1.0
--- a/vars/junitSimpleReport.groovy
+++ b/vars/junitSimpleReport.groovy
@@ -33,7 +33,7 @@ void call(Map config = [:]) {
     String je = config.get('errors', zero)
     String jf = config.get('fails', zero)
     String message = config.get('message', '')
-    String testdata = config.get('testdata', '')
+    String testdata = config.get('testdata')
     boolean ignoreFailure = config.get('ignoreFailure', false)
     String tresult
     // Enforce consistency
--- a/vars/runTest.groovy
+++ b/vars/runTest.groovy
@@ -95,11 +95,15 @@ Map call(Map config = [:]) {
         }
     }
 
-    // We need to pass the rc value to post step.
+    // We need to pass the rc ignore_fairue values to post step.
     // The currentBuild result is for all innerstages, not just this
     // stage, so can not be trusted for this.
     String result_stash = 'result_for_' + sanitizedStageName()
     writeFile(file: result_stash, text: "${rc}")
+    stash name: result_stash,
+          includes: result_stash
+    result_stash = 'ignore_failure_for_' + sanitizedStageName()
+    writeFile(file: result_stash, text: "${ignore_failure}")
     stash name: result_stash,
           includes: result_stash
 
@@ -118,7 +122,9 @@ Map call(Map config = [:]) {
     if (rc != 0) {
         status = 'FAILURE'
     } else if (notify_result) {
-        // Currently only used here for unit testing.
+        // Not used for Unit or Functional testing any more
+        // The junit files are only present in the workspace in the post
+        // section of a stage.
         status = checkJunitFiles(config)
     }
 
--- a/vars/runTest.groovy
+++ b/vars/runTest.groovy
@@ -132,7 +132,7 @@ Map call(Map config = [:]) {
     }
 
     if (status != 'SUCCESS') {
-        String msg = ' See Junit report for details'
+        String msg = ' See Job Test Results report for details'
         if (ignore_failure) {
             catchError(stageResult: 'UNSTABLE',
                        buildResult: 'SUCCESS') {
--- a/vars/runTest.groovy
+++ b/vars/runTest.groovy
@@ -95,7 +95,7 @@ Map call(Map config = [:]) {
         }
     }
 
-    // We need to pass the rc ignore_fairue values to post step.
+    // We need to pass the rc and ignore_failure values to post step.
     // The currentBuild result is for all innerstages, not just this
     // stage, so can not be trusted for this.
     String result_stash = 'result_for_' + sanitizedStageName()
--- a/vars/runTest.groovy
+++ b/vars/runTest.groovy
@@ -95,18 +95,6 @@ Map call(Map config = [:]) {
         }
     }
 
-    // We need to pass the rc and ignore_failure values to post step.
-    // The currentBuild result is for all innerstages, not just this
-    // stage, so can not be trusted for this.
-    String result_stash = 'result_for_' + sanitizedStageName()
-    writeFile(file: result_stash, text: "${rc}")
-    stash name: result_stash,
-          includes: result_stash
-    result_stash = 'ignore_failure_for_' + sanitizedStageName()
-    writeFile(file: result_stash, text: "${ignore_failure}")
-    stash name: result_stash,
-          includes: result_stash
-
     if (cb_result != currentBuild.result) {
         println('Some other stage changed the currentBuild result to ' +
                 "${currentBuild.result}.")
@@ -150,6 +138,17 @@ Map call(Map config = [:]) {
     }
 
     int runTime = durationSeconds(startDate)
-    return ['result': status,
-            'runtest_time': runTime]
+
+    // We need to pass the rc to the post step.
+    Map results = ['result_code': rc,
+                   'result': status,
+                   'runtest_time': runTime]
+
+    String results_map = 'results_map_' + sanitizedStageName()
+    writeYaml file: results_map,
+              data: results,
+              overwrite: true
+    stash name: results_map,
+          includes: results_map
+    return results
 }
--- a/vars/runTest.groovy
+++ b/vars/runTest.groovy
@@ -48,6 +48,8 @@ Map call(Map config = [:]) {
     // github expectations at the same time to also include any Matrix
     // environment variables.
 
+    // Must use Date() in pipeline-lib
+    // groovylint-disable-next-line NoJavaUtilDate
     Date startDate = new Date()
     String context = config.get('context', 'test/' + env.STAGE_NAME)
     String description = config.get('description', env.STAGE_NAME)
@@ -80,8 +82,13 @@ Map call(Map config = [:]) {
     }
 
     String cb_result = currentBuild.result
-    int rc = 0
-    rc = sh(script: script, label: flow_name, returnStatus: true)
+    int rc = 2
+    try {
+        sh(script: script, label: flow_name)
+        rc = 0
+    } catch (hudson.AbortException e) {
+        rc = 1
+    }
 
     // We need to pass the rc value to post step.
     // The currentBuild result is for all innerstages, not just this
--- a/vars/runTest.groovy
+++ b/vars/runTest.groovy
@@ -82,12 +82,17 @@ Map call(Map config = [:]) {
     }
 
     String cb_result = currentBuild.result
-    int rc = 2
+    int rc = 255
     try {
         sh(script: script, label: flow_name)
         rc = 0
     } catch (hudson.AbortException e) {
-        rc = 1
+        // groovylint-disable UnnecessaryGetter
+        // groovylint-disable-next-line NoDef, VariableTypeRequired
+        def rc_val = (e.getMessage() =~ /\d+$/)
+        if (rc_val) {
+            rc = rc_val[0] as Integer
+        }
     }
 
     // We need to pass the rc value to post step.
@@ -127,13 +132,14 @@ Map call(Map config = [:]) {
     }
 
     if (status != 'SUCCESS') {
+        String msg = ' See Junit report for details'
         if (ignore_failure) {
             catchError(stageResult: 'UNSTABLE',
                        buildResult: 'SUCCESS') {
-                error(env.STAGE_NAME + ' failed: ' + rc)
+                error(env.STAGE_NAME + ' failed: ' + rc + msg)
             }
         } else {
-            error(env.STAGE_NAME + ' failed: ' + rc)
+            error(env.STAGE_NAME + ' failed: ' + rc + msg)
         }
     }
 
--- a/vars/unitTestPost.groovy
+++ b/vars/unitTestPost.groovy
@@ -119,7 +119,7 @@ void call(Map config = [:]) {
         }
     }
     if (cbcResult != currentBuild.currentResult &&
-        resultIsWorseOrEqualTo('FAILURE')) {
+        currentBuild.resultIsWorseOrEqualTo('FAILURE')) {
         error 'unitTestPost detected a failure'
     }
 }
--- a/vars/unitTestPost.groovy
+++ b/vars/unitTestPost.groovy
@@ -5,14 +5,9 @@
    *
    * @param config Map of parameters passed
    *
-   * config['always_script']       Script to run after any test.
-   *                               Default 'ci/unit/test_post_always.sh'.
-   *
    * TODO: Always provided, should make required
    * config['artifacts']           Artifacts to archive.
    *                               Default ['run_test.sh/*']
-   *
-   * config['ignore_failure']      Ignore test failures.  Default false.
    *
    * config['referenceJobName']    Reference job name.
    *                               Defaults to 'daos-stack/daos/master'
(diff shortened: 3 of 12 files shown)


vars/runTest.groovy
```diff
@@ -80,8 +82,13 @@ Map call(Map config = [:]) {
...
     String cb_result = currentBuild.result
-    int rc = 0
-    rc = sh(script: script, label: flow_name, returnStatus: true)
+    int rc = 2
+    try {
+        sh(script: script, label: flow_name)
+        rc = 0
+    } catch (hudson.AbortException e) {
+        rc = 1
+    }
```
brianjmurrell, line 91: What's the benefit of using a `try`/`catch` here vs. how it used to be, given that it means that we lose the return value of test script run in the stash?

brianjmurrell (changes requested):
Is there a daos run somewhere demonstrating the results of this change?

vars/runTest.groovy
JohnMalmberg: Here were only checking if the rc value is non-zero.

We get a lot of tickets where the developer filing the ticket only provides the error text from the "error" method called later in the stage.  This makes the step that had the error and potentially useful information highlighted as a place to look, which should make it easier for developers to see where the failure occurred an cut down on the tickets.

JohnMalmberg:
This is a of DAOS with modified to have a valgrind failure.
https://build.hpdd.intel.com/blue/organizations/jenkins/daos-stack%2Fdaos/detail/PR-11595/11/pipeline/544

vars/runTest.groovy
brianjmurrell: > Here were only checking if the rc value is non-zero.

But we stash the return code for future stages to analyze.  This change masks the return code with a pre-defined 2, 0 or 1.
 
> We get a lot of tickets where the developer filing the ticket only provides the error text from the "error" method called later in the stage.

Which ultimately is wrong.  Those should surface up as JUnit results.  Users should never have to look at steps to figure out what went wrong.  Similar to [SRE-652](https://daosio.atlassian.net/browse/SRE-652).

vars/unitTestPost.groovy
```diff
@@ -26,90 +26,104 @@
...
 
-    if (cb_result != currentBuild.result) {
-      println "The recordIssues step changed result to ${currentBuild.result}."
+        if (cb_result != currentBuild.result) {
+            println(
+              "The recordIssues step changed result to ${currentBuild.result}.")
+        }
+    }
+    if (valgrind_found) {
+        error "Valgrind Errors detected in ${env.STAGE_NAME}"
```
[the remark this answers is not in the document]
brianjmurrell: That's fair enough (and I've requested confirmation on that in the ticket as this conflicts directly with the original requirements gathering for our CI).  But we should not be employing `error()`, and confusingly reporting the stage as `FAILED` (i.e. the process of running the test failed), because what it really is is `UNSTABLE` (tests running was successful but some tests had failures) to achieve this.

The subsequent stages already support skipping if the current build result is `UNSTABLE`:
https://github.com/daos-stack/pipeline-lib/blob/master/vars/skipStage.groovy#L175-L188 called from https://github.com/daos-stack/pipeline-lib/blob/master/vars/skipStage.groovy#L415 and https://github.com/daos-stack/pipeline-lib/blob/master/vars/skipStage.groovy#L529.  So you just need to make the stage that has valgrind errors `UNSTABLE` to achieve the goal of stopping the pipeline for valgrind errors.

The above mechanism also allows the behaviour of stopping on `UNSTABLE` stages to be overridden by a commit pragma, so is ultimately more flexible.  It will give us an answer for developers that complain that they don't want valgrind errors to be the end of their pipeline.  It could be that gatekeeping makes this a hard requirement (due to the escape that led to SRE-1729), but we should address that in `skipStage()`.

vars/runTest.groovy
brianjmurrell: > Surfacing as Junit results is just one part of the solution.

I don't think so.  It's the entire solution, along with education about the workflow improvements.

> The text of the error message that fails the stage likely needs to be modified to tell people to check the JUnit test results.

At the point that we are surfacing all possible errors/failures into JUnit results, users will be educated that for any/every failure of a PR, the cause will be in test results and that they can completely ignore the step output.
 
> The changing to use try/catch blocks is https://daosio.atlassian.net/browse/SRE-1683

I'm not sure I'm understanding how one is any different from the other.  Is there a "visual" difference in the step reporting of using try/catch vs. the older implementation?  Is there something I am missing there?

To be clear, I understand what SRE-1683 is saying.  I just don't understand how this code change effects that.
 
> It looks like the actual result code can be obtained by parsing one of 'e.toString()' or 'e.getMessage()' which returns "script returned exit code N'

So that could be parsed to stash the real exit value and restore the previous behaviour.  That only seems worth it if the change from the simple `sh()` call to `try`/`catch`.

vars/unitTestPost.groovy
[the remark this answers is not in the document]
JohnMalmberg: Changed in current push to set the stage to unstable, so a commit pragma can be used as mentioned above.

vars/runTest.groovy
JohnMalmberg: Yes, using a try/catch block causes the "sh" step with a non-zero return to be highlighted in red so that it easily found.

Current push retrieves the rc value from the exception message, which we were stashing to allow its use in post steps so it is there if we ever want to look at specific values.  The default of 255 is for an exception that an exit code could not be parsed from.

This is demonstrated in:
https://build.hpdd.intel.com/blue/organizations/jenkins/daos-stack%2Fdaos/detail/PR-11595/23/pipeline/544

The step that actually ran the test is set to RED to indicate that it was not a SUCCESSFUL test run.
The Unit Test with memcheck on EL 8 step is set to RED as before, but now advises to check the Junit results
The Archive JUnit step is ORANGE for me to indicate that it set the stage to unstable.
The grep for errors in the valgrind memcheck.xml files are RED because they found an issue.
The marking of the stage as unstable because of the valgrind failures is ORANGE for me.

vars/runTest.groovy
```diff
@@ -82,12 +82,17 @@ Map call(Map config = [:]) {
...
-    int rc = 2
+    int rc = 255
     try {
         sh(script: script, label: flow_name)
         rc = 0
     } catch (hudson.AbortException e) {
-        rc = 1
+        // groovylint-disable UnnecessaryGetter
+        // groovylint-disable-next-line NoDef, VariableTypeRequired
+        def rc_val = (e.getMessage() =~ /\d+$/)
```
brianjmurrell, line 92: Couldn't this be `Integer` or `String` even?

vars/runTest.groovy
```diff
@@ -127,13 +132,14 @@ Map call(Map config = [:]) {
     }
 
     if (status != 'SUCCESS') {
+        String msg = ' See Junit report for details'
```
brianjmurrell, line 135: Maybe something more generic like:

```suggestion
        String msg = ' See job Test Results for details'
```

Then a user doesn't have to know that the Test Results are JUnit.

vars/runTest.groovy
brianjmurrell: > Yes, using a try/catch block causes the "sh" step with a non-zero return to be highlighted in red so that it easily found.

I see.  That's interesting and useful.  But I think we need to be careful not to overdo it, per the comments below.

> This is demonstrated in: https://build.hpdd.intel.com/blue/organizations/jenkins/daos-stack%2Fdaos/detail/PR-11595/23/pipeline/544

I see.  I also see that it didn't stop on the  `UNSTABLE` result from *Unit Tests / Unit Test with memcheck on EL 8*.  Is that `UNSTABLE` stage not setting `currentBuild.currentResult` to `UNSTABLE` and thus triggering https://github.com/daos-stack/pipeline-lib/blob/0afbed05f7b7cd6e283cee05bd890be06e76712c/vars/skipStage.groovy#L187?

> The grep for errors in the valgrind memcheck.xml files are RED because they found an issue.

I don't think this one should be red.  I think there is enough indication in the stage['s steps] that there was a failure without adding more confusing red steps.  Moreover this step is an "internal processing" step that we are adding to more clearly indicate that the over job is unsuccessful, which I think we are doing well enough without making this step red. Users really ought not be interested in the details of this step and making it red I think attracts attention to it where they might even be pasting the output from it into bug reports.

brianjmurrell (changes requested):
The UX on this is looking good but I think all of this also needs to create a JUnit result indicating that there were Valgrind errors and that the user should look for them in the *Classic View* page.

While the particular failure you are injecting with https://github.com/daos-stack/daos/blob/b87b80c1f093ed5076d4dcdc7029c74932eaf9b5/src/vos/tests/vts_io.c#L1068 is resulting in an [obvious test failure](https://build.hpdd.intel.com/job/daos-stack/job/daos/job/PR-11595/23/testReport/run_test/memcheck/Unit_Tests___Unit_Test_with_memcheck_on_EL_8___run_test/) often times we get Valgrind errors and no test failing so users look at the Blue Ocean page for a failed test run and see no failed tests and then start asking around "why did this one fail".  I think we need a test result there indicating that it was Valgrind errors and that the user should look for them in the *Classic View* page.

Surely we must have some "create a JUnit" code somewhere already that could be leveraged for this without a ton of effort.

vars/runTest.groovy
JohnMalmberg: Looking at the logs for that run, the Archive JUnit step should have changed the stage status to UNSTABLE from SUCCESSFUL.

There is a check below that which is suppose to report if the stage status was actually changed, and nothing got logged, even though the stage was marked in the UNSTABLE status color.

I need to make some more test diagnostic runs with my test branch to see what is going on.  It is looking like the JUnit report and the "unstable" steps in a post section changed the stage status without changing the currentBuild.result value.

When we first set up the Unit tests, we were doing the JUnit report in the "Steps" section, not the "Post" section, and I wonder if that made a difference in the behavior.

Calling the "error" step instead of the "unstable" step does stop the pipeline though.

JohnMalmberg: The reason that the test errors for "Unit Test with memcheck for EL 8" are not stopping the build is that the DAOS jenkinsfile is specifying "ignore_failure" to allow them to continue.

In the current push, the ignore_failure should also control if junit reporting of the detected valgrind issues will cause the stage to be marked as failed or just marked as unstable.

brianjmurrell: So, if the caller requested `ignore_failure: true` and valgrind errors are found we disregard the caller's request and make the stage an `FAILED` but if the caller did not request `ignore_failure: true`, make it `UNSTABLE` and continue on with the job?

That seems a bit unintuitive to me.  Isn't the entire cause of SRE-1729 in fact simply that the [daos Jenkinsfile](https://github.com/daos-stack/daos/blob/master/Jenkinsfile) should be changed to not request [ignore_failure: true](https://github.com/daos-stack/daos/blob/master/Jenkinsfile#L819)?

I think the rest of what's in this PR is great, especially the part about creating a JUnit result when Valgrind failures are found, but I don't think we need to contort the meaning of `ignore_failure: true` uniquely for Valgrind stages when it's more general (and consistent) meaning works for this stage as well as any other.

vars/unitTestPost.groovy
```diff
@@ -26,90 +26,115 @@
...
+        int vgerr = 0
+        String rcs = sh label: 'Check for Valgrind errors',
+                 script: "grep -E '<error( |>)' ${src_files} || true",
+                 returnStdout: true
+        if (rcs) {
+            if (config['ignore_failure']) {
+                vgfail = 1
+            } else {
+                vgerr = 1
+            }
```
brianjmurrell, line 69: Per the other comments, this seems to unnecessarily contort the meaning of `ignore_failure` for Valgrind stages so that it has a different and inconsistent meaning.  I think we can preserve the meaning and consistency of `ignore_failure` and simply change the caller to not request it for the Valgrind stage if the reporter of SRE-1729 doesn't want failures in that stage ignored.
```suggestion
            vgfail = 1
```

brianjmurrell (changes requested):

vars/junitSimpleReport.groovy
```diff
@@ -0,0 +1,71 @@
...
+   *                               Default 'simple'
+   * config['suite']               Test suite to for JUnit file.
+   *                               Default sanitizedStageName()
+   * config['testdata']            Test data Default ''.
+   */
+
+void call(Map config = [:]) {
+    int zero = 0
+    int one = 1
+    simple = 'simple'
```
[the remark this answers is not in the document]
JohnMalmberg: Yes, will be fixed in the next push

vars/unitTestPost.groovy
JohnMalmberg: With ignore_failure set, a test failure in a stage marks it as unstable, but the pipeline continues, and a test error will stop the pipeline when Junit reports it.

Current behavior is that with ignore_failure clear, a script return of non-zero or a junit test result of "error" will stop the pipeline, but a junit test result of "failure" will not.

The vgfail=1 causes Junit to mark the stage unstable if it is not already unstable.
The vgerr=1 causes Junit to mark the stage as failed if it is not already failed.

With vgfail always = 1 and ignore failure clear, the stage will not stop the pipeline on a valgrind issue.

Will put that in the next push.

vars/unitTestPost.groovy
```diff
@@ -26,90 +26,104 @@
...
+    }
 
-  def artifact_list = config.get('artifacts', ['run_test.sh/*'])
-  def ignore_failure = config.get('ignore_failure', false)
-  artifact_list.each {
-    archiveArtifacts artifacts: it,
+    List artifact_list = config.get('artifacts', ['run_test.sh/*'])
+    boolean ignore_failure = config.get('ignore_failure', false)
+    artifact_list.each { artifactPat ->
+        println("artifactPat = ${artifactPat}")
```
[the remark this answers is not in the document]
JohnMalmberg: Will be in the next push

vars/unitTestPost.groovy
brianjmurrell: > With ignore_failure set, a test failure in a stage marks it as unstable, but the pipeline continues, and a test error will stop the pipeline when Junit reports it.

The pipeline may _continue_ after that `UNSTABLE` stage, however https://github.com/daos-stack/pipeline-lib/blob/master/vars/skipStage.groovy#L415 and https://github.com/daos-stack/pipeline-lib/blob/master/vars/skipStage.groovy#L529 are supposed to skip the subsequent stages if any previous stage ends up `UNSTABLE`.  That's the mechanism that should be used here, not forcing an `UNSTABLE` (yellow) stage into `FAILURE` (red).

> Current behavior is that with ignore_failure clear,

By _clear_ you mean `false`, correct?

> a script return of non-zero or a junit test result of "error" will stop the pipeline, but a junit test result of "failure" will not.

Which is why we introduced skipping on `UNSTABLE` into [`skipStage()`](https://github.com/daos-stack/pipeline-lib/blob/master/vars/skipStage.groovy)

> The vgfail=1 causes Junit to mark the stage unstable if it is not already unstable. The vgerr=1 causes Junit to mark the stage as failed if it is not already failed.

Understood.  But we should not be converting `UNSTABLE` info `FAILURE` just to effect pipeline stoppage.  We have `skipStage()` and it's check for `UNSTABLE` (and even a commit pragma to override that) to achieve that.

> With vgfail always = 1 and ignore failure clear, the stage will not stop the pipeline on a valgrind issue.

Again, looking at the bigger picture, I don't think Valgrind errors should ever make the stage `FAILED` (red) under any conditions.  Only failing to actually execute the test in that stage should make it `FAILED` (red).  After-all, a Valgrind error is not because the test failed to execute.  The test executed successfully.  It just reported (Valgrind, in this case) errors.  That is a test failure and the result should be `UNSTABLE` (yellow).

brianjmurrell:
Could you please push another iteration of your DAOS PR that demonstrates this PR?  I guess that will need to be[come] a PR you want to land, removing the `ignore_failure: true` from the Valgrind stage, yes?

JohnMalmberg:
https://build.hpdd.intel.com/job/daos-stack/job/daos/view/change-requests/job/PR-11595/30/ is running with ignore_failure: true

brianjmurrell:
> https://build.hpdd.intel.com/job/daos-stack/job/daos/view/change-requests/job/PR-11595/30/ is running with ignore_failure: true

`s/with/without/` it seems.

vars/unitTestPost.groovy
```diff
@@ -26,90 +26,111 @@
...
+            vgfail = 1
+        }
+        String suite = sanitizedStageName()
+        junitSimpleReport suite: suite,
+                          file: suite + '_valgrind_results.xml',
+                          fails: vgfail,
+                          errors: vgerr,
+                          name: 'Valgrind_Memcheck',
+                          class: 'Valgrind',
+                          message: 'Valgrind detected',
```
brianjmurrell, line 74:
```suggestion
                          message: 'Valgrind error detected',
```

ashleypittman: Technically they're memcheck warnings.

brianjmurrell: Fair enough.  Whatever the nomenclature is, but _Valgrind detected_ just reads incomplete.

JohnMalmberg:
>> https://build.hpdd.intel.com/job/daos-stack/job/daos/view/change-requests/job/PR-11595/30/ is running with ignore_failure: true

https://build.hpdd.intel.com/job/daos-stack/job/daos/view/change-requests/job/PR-11595/29 was run with ignore_failure: false.

I mis-read what you posted and changed it back to ignore_failure: true, which is the current behavior in master.

With ignore_failure: false, the pipeline stops when the "Unit Test with memcheck on EL 8" hits the failure that has been introduced as seen in build 29.  There was also a change made to the DAOS PR because the unit test scripts were specifically always returning a success status for Unit tests, which was keeping the runTest.
 
> But again, the pipeline did not stop with an `UNSTABLE` _Unit Tests / Unit Test with memcheck on EL 8_ stage. So somehow, that stage being `UNSTABLE` is not causing `currentBuild.currentResult == 'UNSTABLE'` perhaps?
> Which seems odd. That seems like it would be the standard/default Jenkins behaviour. It's not sane that a stage can be `UNSTABLE` but the `currentBuild.currentResult` isn't also `UNSTABLE`.

The currentBuild.result is definitely set to "UNSTABLE" when the stage completes.  One of my earlier commits to pipeline-lib had lots of println() statements in it to see what was going on.

The stage got marked unstable because the change to cause the valgrind detected errors also seems to have cause a test failure, and the test failure caused the "rc" to be non-zero.

Since the Jenkinsfile had ignore_failure: true set, the stage was set to "UNSTABLE" after looking at the "rc" value, but the build result, which is what skipStage looks at is set to 'SUCCESS'.  That code at line 134 in runTest.groovy.

At that point currentBuild.result (Which has the stage value) is "UNSTABLE" and currentBuild.currentResult is "SUCCESS".

Neither the Junit Plugin or the added Valgrind check changed the currentBuild.result from "UNSTABLE", and they could only change it from "UNSTABLE" to a worse value.

So the pipeline proceeded to run the functional VM tests which passed.

> Final/overall build status seems to be `FAILURE`, but that's unfortunately likely due to the multiple tests that timed out. Pity that those are adding noise to the final result here.

The script run by "runTest" did not return a non-zero status.  That may or may not be an issue that needs adjustment in a DAOS PR.  I have not looked as this set of PRs are primarily focusing on the Valgrind behavior.

The pipeline did not get set to error until https://build.hpdd.intel.com/job/daos-stack/job/daos/view/change-requests/job/PR-11595/30/execution/node/1339/log/ when the Junit check detected "<error" in at least one junit xml file.

The only noise added to the final result is that the valgrind publish in the pipeline post step could also have set the final result to failure.   Since the build was already in a FAILED state from the medium test failure, the Valgrind Publish state did not make any changes.  I will need to try to get a run where non of the functional tests fail.

JohnMalmberg:
https://build.hpdd.intel.com/job/daos-stack/job/daos/view/change-requests/job/PR-11595/31/pipeline-graph/

This is an example of a run where the Unit Test with Memcheck marked the stage it was in as unstable with out changing the global build result from Success, and this was because the "ignore_failure: true" is set as it is currently in master.

Since a Valgrind memcheck failure seems to only produce an "<error" that I can detect, and we have that set to be a JUnit failure instead of a warning, it also does not fail the stage.

A junit result containing "<error" currently would override the "ignore_failure" as it would be an error and stop the pipeline.  In this case the junit results were "<failure" so did not change the result.

The over all job is marked as failure because the Valgrind files collected from Unit Test with Memcheck had an "<error" in them which resulted in publishValgrindReport failing the build.

`01:55:36  The publishValgrind step changed result to FAILURE.`

So the changes from the current behavior are:
- Failed Unit tests steps are now highlighted in red.
- The result code from Unit tests now is now actually returned to runTest (DAOS PR change, not pipeline-lib), previously only rc=0 was returned.
- We do not grep JUnit files for "<failure" or "<error" if the result code is non-zero.
- A new JUnit test will always be present for the local analysis of the Valgrind results.
- If a Valgrind issue is found, it will make the stage unstable when that is reported by Junit.

brianjmurrell:
[This](https://jenkins-1.wolf.hpdd.intel.com:9090/blue/organizations/jenkins/daos-stack%2Fdaos/detail/PR-11595/32/pipeline) is (almost) the desired behaviour.  I would think most ideally, the final result should just be `UNSTABLE` as that is the worst result in the entire pipeline but I'd be willing to live with the final result being `FAILURE` given that the *Unit Tests / Unit Test with memcheck on EL 8* is unstable and no subsequent stages were run.

But as an aside and just curiosity, do we understand why?  Is Jenkins doing that as a result of the Valgrind JUnit file?

ashleypittman:
> But as an aside and just curiosity, do we understand why? Is Jenkins doing that as a result of the Valgrind JUnit file?

There is no Valgrind **JUnit** file, there's are valgrind suppression files.  The way that is handled is performed (I believe) through https://github.com/daos-stack/pipeline-lib/blob/076d15c4f5af8d9fdd8cc5149582c8b49ead2e47/vars/valgrindReportPublish.groovy#L63-L71

vars/unitTestPost.groovy
JohnMalmberg: The settings for if Valgrind fails or marks a build unstable are currently controlled by the valgrindReportPublish.groovy code, and that step was not written to allow a DAOS PR to modify them.  I would prefer any changes there be done as a different PR than this one.

And yes, by ignore_failures being clear, I mean false.  However the current daos/master and daos/release never return a non-zero result for unit tests that fail.

https://build.hpdd.intel.com/job/daos-stack/job/daos/view/change-requests/job/PR-11595/48/ is with DAOS modified to  induce a memory leak and return a non-zero result for unit tests that fail.  Since DAOS is now returning a non-zero result, the code to check if if there were failures reported in Junit tests is not run.

I will need to run another DAOS test PR to check that code.

JohnMalmberg:
https://build.hpdd.intel.com/job/daos-stack/job/daos/view/change-requests/job/PR-11595/49/ is with DAOS modified to induce a memory leak, with the current master and release 2.2 behavior of never returning a non-zero result from utils\run_test.sh.

Since it returns a success for the script, there is no way detect in runTest.groovy that there was any test failure, so there is also no way to mark the stage as "UNSTABLE" with out setting the entire pipeline result as UNSTABLE when ignore_failure==true.
There is also no way to fail the PIPELINE here when ignore_failure==false.

With this pipeline-lib PR, the check for of the Junit results is now working in the post stage, and that marks the stage as UNSTABLE.  Since the stage was not already UNSTABLE, this set the entire pipeline as UNSTABLE, so no further stages were run as a default.

This pipeline-lib PR also detected the valgrind failure, which did not change the pipeline status since it was already unstable.
The final status was of the job was set to failure because of the settings in valgrindReportPublish.groovy.

The status of the memcheck stage was reported to Github as unstable.  With out this PR, a status of SUCCESS appears to be what would have been reported to GitHub at this point.


The complete fix would be in https://build.hpdd.intel.com/job/daos-stack/job/daos/view/change-requests/job/PR-11595/48/  with DAOS modified to induce a memory leak and return a non-zero result for unit tests that fail.

Since DAOS is now returning a non-zero result, the memtest failure caused the stage to be marked UNSTABLE and with the pipeline status maintained as SUCCESS.  If ignore_failure==false then the pipeline would have been marked failed at this point and not continued pass this stage, regardless of the pragma settings.  If we want a non-zero script result to have the stage marked UNSTABLE instead of failing the pipeline, then this PR will need another adjustment.

When the junit results and valgrind results were checked by examining the junit files, since stage was already UNSTABLE, these checks did not change the pipeline result.  If the stage was not already UNSTABLE, detection of a failure here would set the build result to UNSTABLE, and by default, the pipeline would stop.

And as with build 49, the vagrindReportPublish step is set to fail the pipeline on detected valgrind failures.

None of the web UI views of a pipeline indicate when the valgrindPublish step changes the status of the pipeline to failed or unstable.  That can only be seen by looking at the console log for the job.

vars/unitTestPost.groovy
brianjmurrell: > The settings for if Valgrind fails or marks a build unstable at the end of the run are currently controlled by the valgrindReportPublish.groovy code, and that step was not written to allow a DAOS PR to modify them. I would prefer any changes there be done as a different PR than this one.

That's fair enough.  Given that it's a defect in the Jenkins plugin (it should be `UNSTABLE` for test failures, not `FAILURE`) this is just another Jenkins bug we are going to have to learn to live with.  That is assuming of course the origin of the XML that we are feeding to the plugin.  What generates that?  Is it valgrind itself?  Maybe it should be distinguishing between `error` and `fail` better?

> Note: The Valgrind Publish step setting the job to failed status does not show up in the Stage View, Blue Ocean, or in the new Pipeline Graph. It only shows up in the console log because we put a check for that in a wrapper to the Valgring Publish step.

Which is exactly why I had requested we generate our own generic JUnit failure to tell the user about it.

> However the current daos/master and daos/release never return a non-zero result for unit tests that fail

If the unit test framework runs successfully (regardless of actual test success/failure) it should return `0`.  It should only return `!0` if it failed to run tests for some reason.  Actual test failures should be reported by JUnit results.

brianjmurrell (dismissed):
Ultimately, visually at least, https://jenkins-1.wolf.hpdd.intel.com:9090/blue/organizations/jenkins/daos-stack%2Fdaos/detail/PR-11595/49/pipeline/542/ looks like what we want.

I'm not entirely sure I agree about the exit code of the test framework that UT is running.  I have left a comment in your daos PR to test this PR about that.

JohnMalmberg:
We want the unit test script exit code to be returned to runTest so that we can at a minimum set that script step to red if it fails.

We can not check the junit or valgrind xml files for unit or functional tests until the post steps.  The post steps can not set the stage to unstable, with out always failing the build.

That exit status is needed the option to "ignore_failures: true" to work as intended.

Build 48 is the run where the "ignore_failure=true" is functioning as intended.  For build 49 it is not.

Having a test script non-zero exit always set the stage to "UNSTABLE" with out stopping the pipeline is the other option, in that case the stage status will be set to unstable, and the we can have the junit checks in the post steps decide to fail the build if ignore_failure!=true.

jolivier23 (approved):

JohnMalmberg:
This is a workflow with that shows this pipeline-lib where a DAOS test has been modified to have a memory leak, and the run_test.sh script used for Unit Testing suppresses returning the script status.
https://build.hpdd.intel.com/job/daos-stack/job/daos/view/change-requests/job/PR-11595/63/pipeline-console/?selected-node=540

It now honors the ignore_failure: true even if the script status is zero, and will properly report the Unit test status to GitHub.

jolivier23 (dismissed):

brianjmurrell:
> https://build.hpdd.intel.com/job/daos-stack/job/daos/view/change-requests/job/PR-11595/64/pipeline-graph/ is a job where the daos run_test.sh passes through the result value.

But the pipeline didn't stop with the Valgrind stage failure.  Isn't that what [Jeff is asking](https://daosio.atlassian.net/browse/SRE-1729?focusedCommentId=111856) for?

JohnMalmberg:
Jeff has approved this PR, as he stated, that he wanted the valgrind issues to be visible before the Valgrind stage failed the job.
The ignore_failure: true can be removed in the Jenkinsfile if they now want failures in this stage to stop the pipeline.

brianjmurrell:
> Jeff has approved this PR, as he stated, that he wanted the valgrind issues to be visible before the Valgrind stage failed the job. The ignore_failure: true can be removed in the Jenkinsfile if they now want failures in this stage to stop the pipeline.

But given that [stopping the pipeline](https://daosio.atlassian.net/browse/SRE-1729?focusedCommentId=111856) is part of SRE-1729, shouldn't daos-stack/daos#11595 remove that `ignore_failure: true` to complete the ticket?  If so, showing the results of that I think would be more illustrative for this related PR also.

JohnMalmberg:
The ignore_failure:true would need to be removed all branches to get the fail on valgrind issue, so the above run shows what will happen with just this PR landed.
Next run of https://build.hpdd.intel.com/job/daos-stack/job/daos/view/change-requests/job/PR-11595/ has ignore_failure:false.

JohnMalmberg:
https://build.hpdd.intel.com/job/daos-stack/job/daos/view/change-requests/job/PR-11595/69/ has ignore_failure:false the default setting.  The utils\run_test.sh for running unit tests is set to return the status code to the job, and that causes the pipeline to stop with a failure status.

It will need to be a future DAOS PR to the utils/run_test.sh to have it only return a non-zero status for test harness failures instead of all test failures for CI so that it will behave the same as Functional tests.

brianjmurrell:
> https://build.hpdd.intel.com/job/daos-stack/job/daos/view/change-requests/job/PR-11595/69/ has ignore_failure:false the default setting.

I don't see any JUnit result in https://jenkins-1.wolf.hpdd.intel.com:9090/blue/organizations/jenkins/daos-stack%2Fdaos/detail/PR-11595/69/tests/ about the valgrind failure though, just 4 NLT test failures.  Such a JUnit result should/would make the stage `UNSTABLE`.

> The utils\run_test.sh for running unit tests is set to return the status code to the job, and that causes the pipeline to stop with a failure status.

But once again, that is not what we want.  The stage should not be a `FAILURE` but rather an `UNSTABLE`.  The tests didn't fail to run.  We just had a test fail.
 
> It will need to be a future DAOS PR to the utils/run_test.sh to have it only return a non-zero status for test harness failures instead of all test failures for CI so that it will behave the same as Functional tests.

It might be future work to make the test exit 0 in CI even if there are test failures, but in the meanwhile, it's return status should be ignored, as it currently is, not as you have commented out in https://github.com/daos-stack/daos/blob/7307c8809001c97edf1c3442196ea2d0bdee0c2c/utils/run_test.sh#L309-L311.  It is currently ignore precisely because test failures should not make a stage `FAILURE` which is what your un-ignoring it is doing.  It is the JUnit result that is supposed to make the stage `UNSTABLE` and the pipeline skip all further stages (assuming the `Allow-unstable: true` commit pragma is not being used).

Ultimately commenting out the ignoring of that `exit 1` when in CI is wrong and is introducing the regression of making that stage a `FAILURE` when the only thing that happened was some tests failed.  That should be `UNSTABLE` due to the JUnit result that is supposed to be logged when that happens.

JohnMalmberg:
Looks like the changing of the run_test.sh script will be handled by work being done in https://github.com/daos-stack/daos/pull/11869.  Until then, its behavior of never returning a non-zero status code is broken;.

https://build.hpdd.intel.com/job/daos-stack/job/daos/view/change-requests/job/PR-11595/70/ is with the script always returning 0, and with ignore_failure:false.   It unexpectedly continued running the functional test stages, so I will need to investigate that again.

JohnMalmberg:
https://jenkins-1.wolf.hpdd.intel.com:9090/blue/organizations/jenkins/daos-stack%2Fdaos/detail/PR-11595/69/tests/ does not have any junit results because all processing in the steps section of a stage stops on an unhandled error condition.
Which means that having utils/run_test.sh or its replacement return a non-zero code can not be done until it is fixed.
And until it is fixed, we will be missing highlighting some types of failures.

Now need to find out why build 70 of the above did not stop on the stage being marked unstable.

JohnMalmberg:
Shouldn't the NLT failure have also caused the pipeline to not execute the Functional test changes?

brianjmurrell:
> Until then, its behavior of never returning a non-zero status code is broken;.

Sure and the way that we deal with it is to ignore the return status code in CI, which your PR is undoing.  That is wrong.  We shouldn't care about the status code, well, other than to know that running the tests failed in some way.  But until that is fixed, we just won't know if running the tests failed and the stage should be a `FAILURE`.  But that is how this has been for years now and there is no reason for that to change now.

> https://build.hpdd.intel.com/job/daos-stack/job/daos/view/change-requests/job/PR-11595/70/ is with the script always returning 0, and with ignore_failure:false. It unexpectedly continued running the functional test stages, so I will need to investigate that again.

That is supposed to be handled by https://github.com/daos-stack/pipeline-lib/blob/dd1d7810c734a4b987d2e733ff005a97fbacdb44/vars/skipStage.groovy#L415 which is https://github.com/daos-stack/pipeline-lib/blob/dd1d7810c734a4b987d2e733ff005a97fbacdb44/vars/skipStage.groovy#L175-L188

So the question is is `currentBuild.currentResult == 'UNSTABLE'` after the end of the _Unit Tests_ stage?

JohnMalmberg:
I will probably have to add some println statements to find out about the currentResult.currentResult status.
Two stages in Unit tests were unstable.
Current push is to fix the valgrind test results being missing.

JohnMalmberg:
https://build.hpdd.intel.com/job/daos-stack/job/daos/view/change-requests/job/PR-11595/77/
Demonstrates expected behavior when a test failure is detected.  The pipeline-lib code will set the stage red and will cause the pipeline to stop.  Master branch of pipelline will in some cases keep on going to the subsequent stages.
In this case it is because the detected valgrind was setup in pipelline-lib to cause the step to be failed.

Next push will be with the detected valgrind setup to cause the step to be unstable.

JohnMalmberg:
Run with current master, modified to have a memory leak, and ignore_failure for that stage set to true:
https://build.hpdd.intel.com/job/daos-stack/job/daos/view/change-requests/job/PR-11595/88/
Expected failures are reported in Junit test results properly and the job continued and is now waiting to run a medium functional test.

PR #370, opened 2023-03-15
DAOS-11421 test: Set the default Test-nvme arg for vos_on_blob

phender:
Moving setting the default Test-Nvme argument in pipeline-lib as opposed to in the Jenkinsfile.

### DIFF
--- a/vars/parseStageInfo.groovy
+++ b/vars/parseStageInfo.groovy
@@ -70,6 +70,16 @@ String get_commit_pragma_tags(String pragma_suffix) {
   return pragma_tag
 }
 
+String get_default_nvme() {
+  // Get the default test nvme setting
+  if (env.BRANCH_NAME.startsWith('feature/vos_on_blob')) {
+      if (startedByTimer()) {
+        return 'auto_md_on_ssd'
+      }
+      return 'auto'
+    }
+    return 'auto:-3DNAND'
+}
 
 /* groovylint-disable-next-line MethodSize */
 void call(Map config = [:]) {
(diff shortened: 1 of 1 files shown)


vars/parseStageInfo.groovy
```diff
@@ -70,6 +70,16 @@ String get_commit_pragma_tags(String pragma_suffix) {
   return pragma_tag
 }
 
+String get_default_nvme() {
+  // Get the default test nvme setting
+  if (env.BRANCH_NAME.startsWith('feature/vos_on_blob')) {
```
brianjmurrell, line 75: I'm a little concerned about `startsWith()` here.  Somebody could create `feature/vos_on_blobulator` not realizing what it will mean in terms of this code.

But maybe you are thinking of a use case where `startsWith()` is useful?

phender: It is a temporary solution until the feature/vos_on_blob branch merges with master.  At that point we will need a different solution (possibly more stages running the same tests with different Test-nvme assignments) which will at least include removing the reference to this specific branch, but if we were to create a feature/vos_on_blob_2_6 it would be supported.

phender:
The `Test / Functional on CentOS 7 / 01-./harness/core_files.py – FTEST_launch.harness-core_files` failure is known: https://daosio.atlassian.net/browse/DAOS-12719

vars/parseStageInfo.groovy
brianjmurrell: Fair enough.

brianjmurrell (approved):

PR #373, opened 2023-05-08
DAOS-13329 test: Enable running ucx provider testing in PR

phender:
Allow the commit pragmas or build with parameter checkboxes to enable running the Functional Hardware Medium UCX Provider stage in a PR.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -18,7 +18,6 @@
 // Then a second PR submitted to comment out the @Library line, and when it
 // is landed, both PR branches can be deleted.
 //@Library(value='pipeline-lib@my_branch_name') _
-@Library(value='pipeline-lib@pahender/DAOS-13329') _
 
 /* groovylint-disable-next-line CompileStatic */
 job_status_internal = [:]
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -18,6 +18,7 @@
 // Then a second PR submitted to comment out the @Library line, and when it
 // is landed, both PR branches can be deleted.
 //@Library(value='pipeline-lib@my_branch_name') _
+@Library(value='pipeline-lib@pahender/DAOS-13329') _
 
 /* groovylint-disable-next-line CompileStatic */
 job_status_internal = [:]
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -424,7 +424,6 @@ pipeline {
                                 println('')
                                 println('  pragmas:        ' + commit.pragmas)
                                 println('  commit message: ' + cm)
-                                // println('expected skips: ' + commit.skips)
                                 println('')
                                 actual_skips = []
                                 i = 0
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -451,7 +451,13 @@ pipeline {
                                         skips: [isPr(), isPr(), false, !isPr(), !isPr(), !isPr(), !isPr()]],
                                        [pragmas: ['Run-daily-stages: true'],
                                         /* groovylint-disable-next-line UnnecessaryGetter */
-                                        skips: [false, false, false, false, false, false, false]]]
+                                        skips: [false, false, false, false, false, false, false]],
+                                       [pragmas: ['Skip-build-el8-rpm: true'],
+                                        /* groovylint-disable-next-line UnnecessaryGetter */
+                                        skips: [isPr(), isPr(), true, true, true, true, true]],
+                                       [pragmas: ['Skip-build-leap15-rpm: true'],
+                                        /* groovylint-disable-next-line UnnecessaryGetter */
+                                        skips: [true, isPr(), false, !isPr(), !isPr(), isPr(), !isPr()]]]
                             errors = 0
                             commits.each { commit ->
                                 cm = 'Test commit\n\n'
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -418,9 +418,9 @@ pipeline {
                                              'COMMIT_MESSAGE=' + cm.stripIndent()]) {
                                         // Useful for debugging since Jenkins'
                                         // assert() is pretty lame
-                                        //println('For stage: ' + stage + ', assert(skipStage(commit_msg: ' +
-                                        //        cm.trim() + ') == ' + commit.skips[i] + ') value is: ' +
-                                        //        skipStage(commit_msg: cm))
+                                        println('For stage: ' + stage + ', assert(skipStage(commit_msg: ' +
+                                               cm.trim() + ') == ' + commit.skips[i] + ') value is: ' +
+                                               skipStage(commit_msg: cm))
                                         assert(skipStage(commit_msg: cm) == commit.skips[i])
                                         i++
                                     }
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -438,8 +438,19 @@ pipeline {
                                     }
                                 }
                                 println('')
-                                println('expected skips: ' + commit.skips)
-                                println('actual skips:   ' + actual_skips)
+                                println('  Expect  Actual  Stage')
+                                println('  ------  ------  ----------------------------------------------')
+                                i = 0
+                                stages.each { stage ->
+                                    expect = 'run '
+                                    if (commit.skips[i]) {expect = 'skip'}
+                                    actual = 'run '
+                                    if (actual_skips[i]) {actual = 'skip'}
+                                    println('  ' + expect + '    ' + actual + '    ' + stage)
+                                    i++
+                                }
+                                // println('  expected skips: ' + commit.skips)
+                                // println('  actual skips:   ' + actual_skips)
                                 println('')
                                 cachedCommitPragma(clear: true)
                             }
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -451,7 +451,7 @@ pipeline {
                                         skips: [isPr(), isPr(), false, !isPr(), !isPr(), !isPr(), !isPr()]],
                                        [pragmas: ['Run-daily-stages: true'],
                                         /* groovylint-disable-next-line UnnecessaryGetter */
-                                        skips: [false, false, false, false, false, false, false]],
+                                        skips: [isPr(), isPr(), false, false, false, false, false]],
                                        [pragmas: ['Skip-build-el8-rpm: true'],
                                         /* groovylint-disable-next-line UnnecessaryGetter */
                                         skips: [isPr(), isPr(), true, true, true, true, true]],
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -425,9 +425,6 @@ pipeline {
                                              'UNIT_TEST=true',
                                              'pragmas=' + env.tmp_pragmas,
                                              'COMMIT_MESSAGE=' + cm.stripIndent()]) {
-                                        // println('  stage:                    ' + stage)
-                                        // println('  skipped (expect==actual): ' +
-                                        //         commit.skips[i] + ' == ' + skipStage(commit_msg: cm))
                                         actual_skips.add(skipStage(commit_msg: cm))
                                         if (skipStage(commit_msg: cm) != commit.skips[i]) {
                                             println('  status: FAIL, stage: ' + stage)
@@ -445,7 +442,7 @@ pipeline {
                                 println('')
                                 cachedCommitPragma(clear: true)
                             }
-                            assert(errors != 0)
+                            assert(errors == 0)
                         }
                         /* tests for all stages:
                               1. Test-tag: datamover
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -428,7 +428,7 @@ pipeline {
                                         // println('  stage:                    ' + stage)
                                         // println('  skipped (expect==actual): ' +
                                         //         commit.skips[i] + ' == ' + skipStage(commit_msg: cm))
-                                        actual_skip.add(skipStage(commit_msg: cm))
+                                        actual_skips.add(skipStage(commit_msg: cm))
                                         if (skipStage(commit_msg: cm) != commit.skips[i]) {
                                             println('  status: FAIL, stage: ' + stage)
                                             errors++
@@ -441,7 +441,7 @@ pipeline {
                                 }
                                 println('')
                                 println('expected skips: ' + commit.skips)
-                                println('actual skips:   ' + commit.skips)
+                                println('actual skips:   ' + actual_skips)
                                 println('')
                                 cachedCommitPragma(clear: true)
                             }
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -412,8 +412,11 @@ pipeline {
                                 println('-------------------------')
                                 println('Unit test commit message:')
                                 println('')
-                                println(commit)
+                                println('  pragmas:        ' + commit.pragmas)
+                                println('  commit message: ' + cm)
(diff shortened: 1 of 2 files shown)


vars/skipStage.groovy
```diff
@@ -156,12 +156,19 @@ boolean skip_ftest_valgrind(String distro, String target_branch) {
...
-           !paramsValue('CI_' + size.replace('-', '_') + '_TEST', true) ||
-           skip_stage_pragma('func-test') ||
+    if (already_passed() || !testsInStage()) {
+        // Stage has already run or there are no tests to run in the stage - always skip
+        return true
+    }
+    if (paramsValue('CI_' + size.replace('-', '_') + '_TEST', false) ||
+        !skip_stage_pragma('func-hw-test-' + size, true) ||
+        !skip_stage_pragma('func-test', true)) {
+        // Checked build with params checkbox or commit pragma forcing stage to run
```
brianjmurrell, line 166: So this is a normally skipped stage that you want to be able to force to run with `Skip-func-hw-test-<size>: false`?

vars/skipStage.groovy
```diff
@@ -156,12 +156,19 @@ boolean skip_ftest_valgrind(String distro, String target_branch) {
...
-    return already_passed() ||
-           env.DAOS_STACK_CI_HARDWARE_SKIP == 'true' ||
-           !paramsValue('CI_' + size.replace('-', '_') + '_TEST', true) ||
-           skip_stage_pragma('func-test') ||
+    if (already_passed() || !testsInStage()) {
+        // Stage has already run or there are no tests to run in the stage - always skip
+        return true
+    }
+    if (paramsValue('CI_' + size.replace('-', '_') + '_TEST', false) ||
+        !skip_stage_pragma('func-hw-test-' + size, true) ||
```
brianjmurrell, line 164: Second argument to `skip_stage_pragma()` is a `String`:
https://github.com/daos-stack/pipeline-lib/blob/79760290f865a55896cdc9a2dee0218537db4f27/vars/skipStage.groovy#L85
and you are passing a Boolean to it here.

```suggestion
        !skip_stage_pragma('func-hw-test-' + size, 'true') ||
```

Indeed, a variant of this function that takes a `Boolean` and returns a `Boolean` so that it can be used in logic statements would indeed be useful.  It has occurred to me many times but never been itchy enough to spend the time writing it.

brianjmurrell (changes requested):

vars/skipStage.groovy
brianjmurrell: Still wondering about this question.

vars/skipStage.groovy
phender: Fixed, thanks.

vars/skipStage.groovy
```diff
@@ -156,20 +197,53 @@ boolean skip_ftest_valgrind(String distro, String target_branch) {
...
+    }
+
+    // Run the stage if its build parameter is either:
+    //   1) enabled by default
+    //   2) selected by the user
+    //   3) not defined
+    // unless it is either a:
+    //   1) PR running the Functional Hardware Medium UCX provider stage
+    //   2) build not started by the user/timer in the master or release branch
+    if (paramsValue('CI_' + size.replace('-', '_') + '_TEST', true) &&
```
brianjmurrell, line 236: Ditto.

phender:
https://build.hpdd.intel.com/blue/organizations/jenkins/daos-stack%2Fpipeline-lib/detail/PR-373/25/pipeline is expected to fail as the new unit tests are not being run against the newer vars/skipStage.groovy.

vars/skipStage.groovy
```diff
@@ -114,33 +116,72 @@ boolean skip_scan_rpms(String distro, String target_branch) {
            quickFunctional()
 }
 
+boolean main_branch() {
```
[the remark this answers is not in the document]
phender: Done.

vars/skipStage.groovy
```diff
@@ -114,33 +116,72 @@ boolean skip_scan_rpms(String distro, String target_branch) {
...
-           (isPr() && distro != 'el8')
+
+    // Run the stage if its build parameter is either:
+    //   1) enabled by default
+    //   2) selected by the user
+    //   3) not defined
+    // unless it is either:
+    //   1) a PR running on any OS besides EL8
+    //   2) running on ubuntu
+    if (paramsValue('CI_FUNCTIONAL_' + distro + '_TEST', true) &&
```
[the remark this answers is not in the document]
phender: Indeed, I've added code to only look at the build parameter for determining if to skip the stage when the build is started manually or via a timer.

vars/skipStage.groovy
```diff
@@ -143,6 +143,12 @@ boolean skip_ftest(String distro, String target_branch) {
         return true
     }
 
+    // When the stage is started manually or via a timer skipping the stage is
+    // controlled by the build parameter and commit pragmas are ignored
+    if (startedByTimer() || startedByUser()) {
```
brianjmurrell, line 148: We also have [startedByUpstream()](https://github.com/daos-stack/pipeline-lib/blob/c80ea0162f5f1bda248544dd6f46eae6d8fe50e5/vars/startedByUpstream.groovy) which sets parameters for the downstream test job.  Should it override commit pragmas also?  Upstream test jobs _can_ take a commit pragma to name (a) branch(es) to use instead of the default of `master` and `release/2.2` for the downstream test job so it's conceivable that it might be asked to use a PR that has commit pragmas.

vars/skipStage.groovy
```diff
@@ -145,7 +145,7 @@ boolean skip_ftest(String distro, String target_branch) {
 
     // When the stage is started manually or via a timer skipping the stage is
     // controlled by the build parameter and commit pragmas are ignored
-    if (startedByTimer() || startedByUser()) {
+    if ((startedByTimer() || startedByUser()) && !startedByUpstream()) {
```
brianjmurrell, line 148: Is `!` the condition we want here?  My  concern was that if an upstream job starts a downstream job from a PR (rather than it's default of `master` or `release/2.2` and that PR had commit pragmas, the upstream job has no ability to override them with the parameters it initiates the downstream job with.

It's a corner-case (or maybe rare is a better word -- that somebody uses a PR/branch for a downstream test job) to be sure, but it something that is currently possible with the given upstream/downstream testing framework that I have put into place.

phender: What commit message does a downstream job read when it is started by an upstream job?  Does it read the commit message from the upstream job?

vars/skipStage.groovy
```diff
@@ -114,33 +116,78 @@ boolean skip_scan_rpms(String distro, String target_branch) {
...
+    return already_passed() || !testsInStage() ||
+           (docOnlyChange(target_branch) && prRepos(distro) == '')
+}
+
+void test_print(message) {
+    // Print the line if this is run in a unit test
+    if (env.UNIT_TEST && env.UNIT_TEST == 'true') {
+        println(message)
+    }
+}
```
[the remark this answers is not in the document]
phender: To provide information when the functions are being used in unit tests in the Jenkninsfile.

vars/skipStage.groovy
phender: I've reworked this fix and its now the changes to `skip_ftest_hw` are modeled after theapproach used in `skip_ftest`.

vars/skipStage.groovy
phender: I've reworked this fix and its now the changes to `skip_ftest_hw` are modeled after theapproach used in `skip_ftest`.

phender:
All Skip unit tests passed in https://[build.hpdd.intel.com/blue/rest/organizations/jenkins/pipelines/daos-stack/pipelines/pipeline-lib/branches/PR-373/runs/32/nodes/50/log/?start=0](https://build.hpdd.intel.com/blue/rest/organizations/jenkins/pipelines/daos-stack/pipelines/pipeline-lib/branches/PR-373/runs/32/nodes/50/log/?start=0)

vars/skipStage.groovy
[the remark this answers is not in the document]
phender: This is no longer used.

vars/skipStage.groovy
```diff
@@ -545,22 +557,22 @@ boolean call(Map config = [:]) {
                    skip_if_unstable()
         case 'Functional_Hardware_Small':
         case 'Functional Hardware Small':
-            return skip_ftest_hw('small', target_branch)
+            return skip_ftest_hw('small', 'el8', target_branch)
```
JohnMalmberg, line 560: Why are we putting a distro here?
While we are currently using el-8, this looks like something that will need to be addressed when we move to el-9.

phender: I've added skipping the HW stages if we skipped building  the RPMs that they require, similar to what we curently do in `skip_ftest()` for the functional VM stages.  While thios isn't much different than what we do when we call `skip_ftest_valgrind('el8', target_branch)` for example, I could see an improvement to use a variable.  The `'target'` from parseStageInfo be the correct value to use here, right?

JohnMalmberg: Yes.  Ideally we would not need a parameter as we should know that the specific rpm build was skipped.

phender:
Unit test results from build 32: https://[build.hpdd.intel.com/blue/rest/organizations/jenkins/pipelines/daos-stack/pipelines/pipeline-lib/branches/PR-373/runs/32/nodes/50/log/?start=0](https://build.hpdd.intel.com/blue/rest/organizations/jenkins/pipelines/daos-stack/pipelines/pipeline-lib/branches/PR-373/runs/32/nodes/50/log/?start=0)

vars/skipStage.groovy
brianjmurrell: Ultimately I think you need to utilize [hwDistroTarget()](https://github.com/daos-stack/pipeline-lib/blob/668ba308eadd4cc9980cf758190c752e1e073546/vars/hwDistroTarget.groovy) here.

phender: I've updated this to use `hwDistroTarget2()`

vars/skipStage.groovy
```diff
@@ -229,6 +229,7 @@ boolean call(Map config = [:]) {
     }
 
     String target_branch = env.CHANGE_TARGET ? env.CHANGE_TARGET : env.BRANCH_NAME
+    String hw_distro = hwDistroTarget2().join("")
```
brianjmurrell, line 232: Ahh.  Disregard my previous comment as you used the `2` variant while I was reviewing.

vars/skipStage.groovy
```diff
@@ -157,17 +155,31 @@ boolean skip_ftest_valgrind(String distro, String target_branch) {
            target_branch =~ branchTypeRE('weekly')
 }
 
-boolean skip_ftest_hw(String size, String target_branch) {
-    return already_passed() ||
+boolean skip_ftest_hw(String size, String distro, String target_branch) {
```
brianjmurrell, line 158: I'm not sure that `distro` needs to be an argument here and cannot just be queried within the function itself.

vars/skipStage.groovy
```diff
@@ -217,6 +229,7 @@ boolean call(Map config = [:]) {
     }
 
     String target_branch = env.CHANGE_TARGET ? env.CHANGE_TARGET : env.BRANCH_NAME
+    String hw_distro = hwDistroTarget2().join("")
```
brianjmurrell, line 232: I think this is what [hwDistroTarget()](https://github.com/daos-stack/pipeline-lib/blob/668ba308eadd4cc9980cf758190c752e1e073546/vars/hwDistroTarget.groovy) does/is for.

But again, I'm not sure I see the need for this to be an argument to `skip_ftest_hw()` and not just determined internally in `skip_ftest_hw()`.

brianjmurrell (changes requested):

vars/skipStage.groovy
phender: Thanks. Agreed.

phender:
All skipStage unit tests passed in https://[build.hpdd.intel.com/blue/rest/organizations/jenkins/pipelines/daos-stack/pipelines/pipeline-lib/branches/PR-373/runs/34/nodes/50/log/?start=0](https://build.hpdd.intel.com/blue/rest/organizations/jenkins/pipelines/daos-stack/pipelines/pipeline-lib/branches/PR-373/runs/34/nodes/50/log/?start=0)

phender:
The skipStage unit tests failed in https://[build.hpdd.intel.com/blue/rest/organizations/jenkins/pipelines/daos-stack/pipelines/pipeline-lib/branches/PR-373/runs/35/nodes/50/log/?start=0](https://build.hpdd.intel.com/blue/rest/organizations/jenkins/pipelines/daos-stack/pipelines/pipeline-lib/branches/PR-373/runs/35/nodes/50/log/?start=0) because they were run against the older skipStage implementation as I removed the reference to this PR from the Jenkinsfile for landing.

vars/skipStage.groovy
```diff
@@ -158,16 +156,31 @@ boolean skip_ftest_valgrind(String distro, String target_branch) {
...
+        return true
+    }
+    // Run the functional hardware test stage if explicitly requested by the user
+    if (run_default_skipped_stage('func-hw-test-' + size) ||
+        run_default_skipped_stage('func-test-hw-' + size) ||
+        cachedCommitPragma('Run-daily-stages').toLowerCase() == 'true') {
+        // Forced to run due to a (Skip) pragma set to false
+        return false
+    }
+    String distro = hwDistroTarget()
```
brianjmurrell, line 171: Since you already have `size` in this function, it would be more optimal to pass it to `hwDistroTarget()` but more importantly, you should be reusing this `distro` variable down on line 185 where `hwDistroTarget(size)` is being called again.

brianjmurrell (changes requested):

vars/skipStage.groovy
phender: Ah right, thanks.

daltonbohning (dismissed):

phender:
New skipTest unit testing passed in https://[build.hpdd.intel.com/blue/rest/organizations/jenkins/pipelines/daos-stack/pipelines/pipeline-lib/branches/PR-373/runs/36/nodes/49/log/?start=0](https://build.hpdd.intel.com/blue/rest/organizations/jenkins/pipelines/daos-stack/pipelines/pipeline-lib/branches/PR-373/runs/36/nodes/49/log/?start=0)

phender:
Ticket for failed downstream release/2.2 and weekly-testing-2.2 branch testing in the PR: https://daosio.atlassian.net/browse/DAOS-14154

Jenkinsfile
```diff
@@ -18,6 +18,7 @@
 // Then a second PR submitted to comment out the @Library line, and when it
 // is landed, both PR branches can be deleted.
 //@Library(value='pipeline-lib@my_branch_name') _
+@Library(value='pipeline-lib@pahender/DAOS-13329') _
```
brianjmurrell, line 21: FWIW, I have come across a way of avoiding having to do this, where a PR uses the branch from the PR.

It's going to take some experimentation on a staging Jenkins to make sure we get correct and don't blow everything up for everyone by not testing.

brianjmurrell (changes requested):
Only -1 for the `@Library()` that needs reverting.

brianjmurrell (approved):

daltonbohning (approved):

PR #374, opened 2023-05-18
Add EL9 support

brianjmurrell:


Jenkinsfile
```diff
@@ -17,7 +17,7 @@
 // That PR should be landed with out deleting the PR branch.
 // Then a second PR submitted to comment out the @Library line, and when it
 // is landed, both PR branches can be deleted.
-//@Library(value='pipeline-lib@my_branch_name') _
+@Library(value="pipeline-lib@bmurrell/el9") _
```
brianjmurrell, line 20: I will remove this on landing.

brianjmurrell:
> Note that you need to remove the pipeline-lib reference from the Jenkinsfile.

Indeed, I [noted as much](https://github.com/daos-stack/pipeline-lib/pull/374#pullrequestreview-1488702066).  I am going to experiment with removing that in the process of landing this PR with the goal that the PR passes test in it's reviewed state and then also passes on landing with the `pipeline-lib` referenced removed.

brianjmurrell:
Landed without `pipeline-lib` self-reference in cab2c82153c088bfb461c367be4c13f5e0995770.

PR #375, opened 2023-05-26
DAOS-623 ci: Add bdev stages to skipStage

jolivier23:
I did this in the simplest way possible.  We really need a pragma that skips all of them (memcheck, bdev memcheck, utest, bdev utest).


JohnMalmberg (approved):

brianjmurrell (approved):
If finer grained skipping is needed in the future, it can be addressed at that time.

phender (approved):

PR #377, opened 2023-06-06
DAOS-13527 test: Support default tag assignment through Jenkinsfile

phender:
The test tags to use the functional test stages should be driven by the method through which the user has started the build. If the build has been built with parameters or started by a timer then the parameters of that build should drive the test tag selection. If the user pushes a PR then any tags defined via commit pragmas should be used followed by the any build parameter tags.

### DIFF
--- a/vars/parseStageInfo.groovy
+++ b/vars/parseStageInfo.groovy
@@ -274,34 +274,42 @@ Map call(Map config = [:]) {
     }
 
     // Determine which tests tags to use
-    String tag
-    if (startedByUser() || startedByUpstream()) {
-      // Test tags defined by the build parameters override all other tags
-      tag = get_build_params_tags()
-    }
-    if (!tag && startedByTimer()) {
-      // Stage defined tags take precedence in timed builds
-      tag = config['test_tag']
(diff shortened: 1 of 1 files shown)


vars/parseStageInfo.groovy
```diff
@@ -274,34 +274,42 @@ Map call(Map config = [:]) {
...
-          tag = config['test_tag']
-          /* groovylint-disable-next-line CouldBeElvis */
+          // Get tags from the build parameters
+          tag = get_build_params_tags()
           if (!tag) {
-            // Otherwise use the default PR tag
-            tag = 'pr'
+            if (startedByUpstream()) {
+              // Default to run simple tests for dowstream builds
+              tag = 'always_passes'
```
brianjmurrell, line 307: Should this reflect https://github.com/daos-stack/pipeline-lib/blob/79f104a6be4dfd33ba4a1adea7826b4ccaad62ac/vars/packageBuildingPipelineDAOSTest.groovy#L743?

TBH, I'm on the fence about whether the default tags for downstream testing belong here or above.

PR #379, opened 2023-06-13
DAOS-13529 test: Add support of + tag group separator

phender:
A '+' string will be used to define different groups of tags to run with different launch.py command lines in the same the test stage.  To enable this the cluster size tags should not be added to this 'tag'.

### DIFF
--- a/vars/parseStageInfo.groovy
+++ b/vars/parseStageInfo.groovy
@@ -343,7 +343,13 @@ Map call(Map config = [:]) {
     // Assemble the stage test tags - add the cluster size to each tag group
     result['test_tag'] = ''
     for (group in tag.split(' ')) {
-      result['test_tag'] += group + ',' + cluster_size + ' '
+      if (group == '+') {
+        // The '+' is a special string in the tag field for defining multiple groups of tags
+        result['test_tag'] += group + ' '
+      }
+      else {
+        result['test_tag'] += group + ',' + cluster_size + ' '
+      }
     }
     result['test_tag'] = result['test_tag'].trim()
 
--- a/vars/parseStageInfo.groovy
+++ b/vars/parseStageInfo.groovy
@@ -300,7 +300,7 @@ Map call(Map config = [:]) {
                 if (!tag) {
                     // Followed by stage defined tags
                     tag = config['test_tag']
-                /* groovylint-disable-next-line CouldBeElvis */
+                    /* groovylint-disable-next-line CouldBeElvis */
                     if (!tag) {
                         // Otherwise use the default PR tag
                         tag = 'pr'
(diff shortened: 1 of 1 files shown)


daltonbohning (dismissed):

phender:
Actual testing of the code change performed in https://github.com/daos-stack/daos/pull/12261

vars/parseStageInfo.groovy
```diff
@@ -344,7 +344,13 @@ Map call(Map config = [:]) {
         // Assemble the stage test tags - add the cluster size to each tag group
         result['test_tag'] = ''
         for (group in tag.split(' ')) {
-            result['test_tag'] += group + ',' + cluster_size + ' '
+            if (group == '+') {
+                // The '+' is a special string in the tag field for defining multiple groups of tags
+                result['test_tag'] += group + ' '
+            }
+            else {
+                result['test_tag'] += group + ',' + cluster_size + ' '
```
brianjmurrell, line 352: FWIW, this could have been done with:

```suggestion
            result['test_tag'] += group + (group != '+' : ',' + cluster_size ? '') + ' '
```

I think.

brianjmurrell:
> Actual testing of the code change performed in [daos-stack/daos#12261](https://github.com/daos-stack/daos/pull/12261) where the cluster-specific tags were not added to the `+` tag:
> 
> ```
> [2023-06-16T14:34:03.411Z] TEST_TAG_ARG: daos_test,hw,large + md_on_ssd,hw,large
> [2023-06-16T14:34:03.411Z] + echo 'TEST_TAG_ARG: daos_test,hw,large + md_on_ssd,hw,large'
> ```

Which commit/jenkins run number was the above from?  I couldn't find one that included the cluster specific tags in the commit message -- which is what I am assuming one needs to do with this new operator, yes?

brianjmurrell (approved):
If I understand what you are trying to achieve, this requires users to add the _cluster_size_ tags themselves if they want to use this new `+` operator, yes?

If so, it was always a goal to remove the need for (particularly non-ftest developer) users to know what the tags are to target specific clusters when trying to add `Test-tag:`s or `Features:` to their PRs.

I'm not sure that I have any bright ideas about how to make that work with this new requirement/feature, but just wanted to throw out there what one of the goals of the _cluster_size_ test tag handling was.

vars/parseStageInfo.groovy
phender: The ':' and '?' where flipped, but it does work, thanks:
```
result['test_tag'] += group + (group != '+' ? ',' + cluster_size : '') + ' '
```

brianjmurrell: Oops. yes.  Sorry.

phender:
Replaced by https://github.com/daos-stack/pipeline-lib/pull/388

PR #381, opened 2023-06-21
DAOS-623 pmdk: mimic how PMDK's CI prepares docker's tmpfs for tests

janekmi:
Required by daos-stack/pmdk and especially by daos-stack/pmdk#26. Tests fail currently because flock(2) allows locking file which is already locked exclusively.

The introduced tmpfs is intended to closely resemble how tmpfs works without docker involved.

Ref: pmem/pmdk#4467
Ref: PMDK-72

### DIFF
--- a/vars/packageBuildingPipeline.groovy
+++ b/vars/packageBuildingPipeline.groovy
@@ -302,7 +302,8 @@ void call(Map pipeline_args) {
                                 label 'docker_runner'
                                 args '--group-add mock' +
                                      ' --cap-add=SYS_ADMIN' +
-                                     ' --privileged=true'
+                                     ' --privileged=true' +
+                                     ' --tmpfs /tmp:rw,relatime,suid,dev,exec,size=6G'
                                 additionalBuildArgs dockerBuildArgs()
                             }
                         }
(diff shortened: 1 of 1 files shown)


grom72 (approved):

brianjmurrell:
I'm not quite sure I understand how/why having a `tmpfs` on `/tmp` is making a difference.  Do I understand that you are saying that `flock(2)` is not working correctly on the `/tmp` that is in the docker container without a `tmpfs` mounted on it?

vars/packageBuildingPipeline.groovy
brianjmurrell, line 1: There are two other `packageBuildingPipeline*()` functions that will need this modification.  It is unfortunate that we now have three of them, indeed.  I personally consider this one and `packageBuildingPipelineDAOS()` to be somewhat deprecated since `packageBuildingPipelineDAOSTest()` provides **much** better testing than the other two.  Not all projects that use the `packageBuildingPipeline*()` pipelines have been converted to `packageBuildingPipelineDAOSTest()`  yet and/or nobody has found the time to DRY these 3 different pipelines out into a single pipeline that can take parameters to act like any of the 3 do.

brianjmurrell (changes requested):

grom72:
More detailed explanation can be found here: https://daosio.atlassian.net/browse/PMDK-72

vars/packageBuildingPipeline.groovy
jolivier23: We should switch pmdk to use the newer one if we are not already.

janekmi:
I am closing this one because the solution on the `docker` level is not sufficient since `mock` is used in this case. For details please see daos-stack/pmdk#31. @brianjmurrell there are also a little bit more details on the `flock(2)` issue if you are interested. Thanks!

PR #391, opened 2023-08-29
DAOS-14252 build: Remove 2.2 from the list of test targets

frostedcmos:
- Remove 2.2 branch for the list of test targets when updating pre-requisites

Signed-off-by: Alexander Oganezov <alexander.a.oganezov@intel.com>

### DIFF
--- a/vars/packageBuildingPipelineDAOS.groovy
+++ b/vars/packageBuildingPipelineDAOS.groovy
@@ -740,8 +740,7 @@ void call(Map pipeline_args) {
                         axis {
                             name 'BRANCH'
                             values 'master',
-                                   'release/2.4',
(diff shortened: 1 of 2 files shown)


brianjmurrell:
This of course cannot land until we have completely EoL'd 2.2, which I don't think we have done yet.

brianjmurrell (changes requested):
Should also remove https://github.com/daos-stack/pipeline-lib/blob/cd0d0a3e2dd66c9a25dc70e6316644889a683ffc/vars/packageBuildingPipelineDAOS.groovy#L744 then I suppose.

brianjmurrell (approved):

PR #396, opened 2023-10-17
DAOS-14449 Always collect Unit Test logs/artifacts

JohnMalmberg:
Change the Unit test code to always collect logs and artifacts even if the test script exits with a failure status.


JohnMalmberg:
The test failures for the weekly 2.4 testing do not use any of the code changed in this PR.

phender (approved):

PR #400, opened 2023-10-27
SRE-574 Build: Cluster offline on failure

JohnMalmberg:
Take the cluster offline if a hardware failure is discovered on a functional test run.

### DIFF
--- a/vars/functionalTestPostV2.groovy
+++ b/vars/functionalTestPostV2.groovy
@@ -64,6 +64,15 @@ void call(Map config = [:]) {
         results['result'] = checkJunitFiles(config)
     }
 
+    // Look for a hardware failure
+    String hardwareJunit = "${env.STAGE_NAME}/hardware_prep/*/results.xml"
+    String hardwareResult = checkJunitFiles(testResults: hardwareJunit)
+    if (hardwareResult == 'FAILURE') {
+        buildAgentControl(action: 'offline',
+                          message: 'Hardware test detected an error',
+                          subject: 'CI Test failure - Hardware issue.')
+    }
+
     String description = config.get('description', env.STAGE_NAME)
     String context = config.get('context', 'test/' + env.STAGE_NAME)
     String flow_name = config.get('flow_name', env.STAGE_NAME)
@@ -107,5 +116,8 @@ void call(Map config = [:]) {
         println(
             "Ignoring failure to record " + env.STAGE_NAME + " tests with launchable: " +
             error.getMessage())
+    }
+    if (!ignore_failure && results['result'] == 'FAILURE') {
+        error "Failure detected with test harness or hardware."
     }
 }
--- a/vars/buildAgentControl.groovy
+++ b/vars/buildAgentControl.groovy
@@ -1,0 +1,31 @@
+/* groovylint-disable DuplicateStringLiteral, DuplicateNumberLiteral
+ */
+// vars/buildAgentControl.groovy
+
+  /**
+   * buildAgentControl step method
+   *
+   * Control a build agent online status.
+   *
+   * @param config Map of parameters passed
+   *
+   * config['action']        'offline' to take a build agent offline.
+   *
+   * config['message']       Reason for taking build agent offline.
+   *
+   * config['subject']       Subject for e-mail notification.
+   */
+   void call(Map config = [:]) {
+        // E-mail needs a newline after the message before
+        // adding build information.
+        message = """${config['message']}
+"""
+        try {
+            emailStatusSystem subject: config['subject'],
+                              body: message
+            buildAgentSystem action: config['action'],
+                             reason: config['message']
+        } catch (java.lang.NoSuchMethodError e) {
+            echo "Did not find emailStatusSystem or buildAgentSystem"
+        }
+   }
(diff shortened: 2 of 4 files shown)


vars/buildAgent.groovy
```diff
@@ -0,0 +1,27 @@
...
+   *
+   * @param config Map of parameters passed
+   *
+   * config['action']        'offline' to take a build agent offline.
+   *
+   * config['message']       Reason for taking build agent offline.
+   *
+   * config['subject']       Subject for e-mail notification.
+   */
+   void buildAgent(Map config = [:]) {
```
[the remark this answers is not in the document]
JohnMalmberg: No it does not work.  was not getting this far because fixes were needed in DAOS also.
Still working on getting a complete test run.

JohnMalmberg:
For full functionality this PR will need a DAOS patches.  The PR for master to implement them is https://github.com/daos-stack/daos/pull/13249 and currently is failing CI due to unrelated DAOS test failures.

vars/functionalTestPostV2.groovy
```diff
@@ -64,6 +64,15 @@ void call(Map config = [:]) {
         results['result'] = checkJunitFiles(config)
     }
 
+    // Look for a hardware failure
+    String hardwareJunit = "${env.STAGE_NAME}/hardware_prep/*/results.xml"
+    String hardwareResult = checkJunitFiles(testResults: hardwareJunit)
+    if (hardwareResult == 'FAILURE') {
```
brianjmurrell, line 70: What about `UNSTABLE`?  Recall `FAILURE` means the test failed to run and `UNSTABLE` means it ran but the test failed.  Is whatever is producing `hardware_prep/*/results.xml` confusing these two results and should be returning `UNSTABLE` when the tests fail (as opposed to running of the tests fails)?

JohnMalmberg: If hardwareResult is 'FAILURE' then the tests did not run because a hardware sanity check reported an error, and the test stage will be failed.

These are the Junit results specifically for the hardware checks and do not contain any application test results, so UNSTABLE will never show up.

brianjmurrell: So if a test runs to determine if PMEM is available and complete/functional but it finds a PMEM module missing is that reported as `FAILURE` or `UNSTABLE`?  If the former, then how is that condition distinguished from the test script to determine PMEM's health from having a previously unknown syntax error that only gets triggered intermittently, for example?  Or perhaps the tools to determine the PMEM health are not available, etc.?

brianjmurrell: > The code that does the hardware check resides in in the DAOS repository not here.

That is orthogonal.  It doesn't matter where it lives.

> Any failure in the hardware test code that is detected will result in marking the test run marked as with an "error" status in the Junit file. An error status in a Junit file will cause the stage to be put in FAILURE state. It is set to the error condition as this is a failure of the test environment, not the something in that is being tested in the DAOS product.

I don't think _DAOS product_ or not is relevant.  You are not testing DAOS at this point.  You need to isolate your thinking about testing hardware from testing DAOS.  Forget the latter.  DAOS testing has not started yet.

So given that shift in thinking, now what you are doing is running tests on the hardware.  Those tests can fail (JUnit `failure`/Jenkins `UNSTABLE`) because something about the expected hardware state was incorrect.  Or the running of the actual test could fail due to syntax error, unchecked error return, missing tools to do the test, etc.  That is JUnit `error`/Jenkins `FAILURE`

> Non-fatal hardware issues are sent out as e-mail messages. They are not put in the junit results so that they do not affect what could be an otherwise successful run. And that code is also in DAOS.

Fair enough.  Also orthogonal.
 
> What this PR does is when that an error result is found in a hardware check junit file, it takes the cluster offline so that we do not have a hardware or configuration error in the cluster cause to quickly go through a backlog waiting that type of cluster.

Understood.  The problem is you could be taking a cluster offline when it doesn't need to be because checking hardware didn't fail but rather the process of checking hardware failed for no fault of the hardware itself (as it was the fault of the checking process/code) and that a subsequent check might pass.

JohnMalmberg: The risks of taking a cluster offline when it should not be are low and that should be limited to PRs in progress and quickly fixed.

The problems for not taking a a cluster offline when it should be is where we run into problems, as it would result in pretty rapid queue draining as has been seen in the past.

The status of any hardware check failing is passed by DAOS as a Junit Failure / Junit error condition as it must both fail the stage and take the resulting code offline.

So is the change you want here is for both Junit failure and junit error for the hardware checks to take the cluster offline?

Or for a junit failure to take the cluster offline, but a junit error condition to not take the cluster offline, which requires a change on DAOS code, including adding adding an additional diagnostic to detect when the hardware script exits with a failure, but did not produce any Junit results.

If we use the junit error condition for syntax/test script failure in this case, and use junit failure for changing reporting if a hardware failure, we are creating a condition where we want the cluster to offline on "unstable" but not on the more severe "error" condition, in that case we are treating "unstable" as a bigger failure than "error" condition.

Right now, the hardware checks only write junit error conditions on failures detected because those are fatal to running tests.   If the script fails due to a syntax error or something that prevents as junit file from being created for that check the script will exit with a non-zero value, which will cause the test stage to fail, but will not take the cluster offline.

brianjmurrell (approved):

PR #402, opened 2023-11-08
DAOS-13529 test: Fix unit tests running in landing PR

phender:
New skipFunctionalTestStage() tests fail when running in a landing PR because when testing the method the run_if_landing argument is set to false. This fix sets it true so the unit tests behave the same in all types of PRs.

Skip-daos-build-and-test: true


Jenkinsfile
```diff
@@ -586,63 +586,63 @@ pipeline {
                             println('------------------------------------------------------------')
                             sequences = [
                                 [description: 'run_if_pr=true',
-                                 kwargs: [tags: 'pr', pragma_suffix: '-hw-medium', distro: null, run_if_pr: true],
+                                 kwargs: [tags: 'pr', pragma_suffix: '-hw-medium', distro: null, run_if_pr: true, run_if_landing=true],
```
brianjmurrell, line 589: I think you want:

```suggestion
                                 kwargs: [tags: 'pr', pragma_suffix: '-hw-medium', distro: null, run_if_pr: true, run_if_landing: true],
```

Many below need the same correction.

brianjmurrell (changes requested):

Jenkinsfile
phender: Yes, thanks.

phender:
Updated unit tests passed in https://build.hpdd.intel.com/job/daos-stack/job/pipeline-lib/job/PR-402/3/

brianjmurrell (approved):

JohnMalmberg (approved):

PR #405, opened 2023-11-17
DAOS-14658 ci: Add support for running tests from alternate branch

phender:
Add support for running tests from a different branch.

### DIFF
--- a/vars/getFunctionalTestStage.groovy
+++ b/vars/getFunctionalTestStage.groovy
@@ -65,7 +65,8 @@ Map call(Map kwargs = [:]) {
                         checkoutScm(
                             url: 'https://github.com/daos-stack/daos.git',
                             branch: base_branch,
-                            withSubmodules: true)
+                            withSubmodules: true,
+                            cleanAfterCheckout: false)
                     } else {
                         checkoutScm(cleanAfterCheckout: false)
                     }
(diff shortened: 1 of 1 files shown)


vars/getFunctionalTestStage.groovy
```diff
@@ -59,7 +61,14 @@ Map call(Map kwargs = [:]) {
                 node(cachedCommitPragma("Test-label${pragma_suffix}", label)) {
                     // Ensure access to any branch provisioning scripts exist
                     echo "[${name}] Check out from version control"
-                    checkoutScm(cleanAfterCheckout: false)
+                    if (base_branch) {
+                        checkoutScm(
+                            url: 'https://github.com/daos-stack/daos.git',
+                            branch: base_branch,
+                            withSubmodules: true)
```
brianjmurrell, line 68: Does this `checkoutScm()` not also need `cleanAfterCheckout: false` like the one on line 70 below does?

brianjmurrell (approved):

vars/getFunctionalTestStage.groovy
phender: I'm not sure.  We don't disable the clean in https://github.com/daos-stack/daos/blob/weekly-testing/Jenkinsfile#L171  Maybe I don't need it in the other line??

brianjmurrell: > I'm not sure. We don't disable the clean in [daos-stack/daos@`weekly-testing`/Jenkinsfile#L171](https://github.com/daos-stack/daos/blob/weekly-testing/Jenkinsfile?rgh-link-date=2023-11-20T14%3A46%3A22Z#L171), so maybe I don't need it in the other line??

Could very easily be oversight on that one.

phender: I've added `cleanAfterCheckout: false`

brianjmurrell (approved):

PR #410, opened 2023-12-12
SRE-1993 test: Enable skip logic for new Test RPMs stages

phender:
Includes a workaround to skip the 'Test RPMs on Leap 15.4' stage if the CI_RPM_TEST_VERSION build parameter is specified.

### DIFF
--- a/vars/skipStage.groovy
+++ b/vars/skipStage.groovy
@@ -484,6 +484,7 @@ boolean call(Map config = [:]) {
                     !run_default_skipped_stage('test-el-8.5-rpms')) ||
                    stageAlreadyPassed()
         case 'Test EL 8.6 RPMs':
+        case 'Test RPMs on EL 8.6':
             return !paramsValue('CI_RPMS_el8.6_TEST', true) ||
                    target_branch =~ branchTypeRE('weekly') ||
                    skip_stage_pragma('test') ||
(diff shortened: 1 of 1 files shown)


vars/skipStage.groovy
```diff
@@ -501,6 +502,14 @@ boolean call(Map config = [:]) {
             return !paramsValue('CI_RPMS_leap15_TEST', true) ||
                    skip_stage_pragma('test-leap-15-rpms', 'true') ||
                    stageAlreadyPassed()
+        case 'Test RPMs on Leap 15.4':
+            return !paramsValue('CI_RPMS_leap15.4_TEST', true) ||
+                   skip_stage_pragma('test') ||
+                   skip_stage_pragma('test-rpms') ||
+                   skip_stage_pragma('test-leap-15.4-rpms', 'true') ||
+                   docOnlyChange(target_branch) ||
+                   (paramsValue('CI_RPM_TEST_VERSION', '') != '') ||
```
brianjmurrell, line 511:
```suggestion
                   (rpmTestVersion() != '') ||
```
covers use of both `params.CI_RPM_TEST_VERSION` and `RPM-test-version:` commit pragma which will both cause the problem.

brianjmurrell (changes requested):

vars/skipStage.groovy
phender: Thanks, good catch!

brianjmurrell (approved):

PR #412, opened 2023-12-12
SRE-1981 Build: Set unstable on test error

JohnMalmberg:
If any test error occurs in a functional test, make sure that the stage is set to unstable, not error, so that it does cause other testing to be aborted.

This will not change a stage from an error state to an unstable state.

### DIFF
--- a/vars/functionalTestPostV2.groovy
+++ b/vars/functionalTestPostV2.groovy
@@ -118,6 +118,6 @@ void call(Map config = [:]) {
             error.getMessage())
     }
     if (!ignore_failure && results['result'] == 'FAILURE') {
-        error "Failure detected with test harness or hardware."
+        unstable "Failure detected with test harness or hardware."
     }
 }


phender (approved):

ryon-jensen (approved):

brianjmurrell:
I'm quite surprised that what I wanted to suggest was not actually possible (yet).  I was going to suggest…

If you wanted to test that this change has the desired effect on daos jobs you can leverage the downstream testing that pipeline-lib does and force the downstream tests to produce failures by adding the following:

```
Test-tag: load_mpi test_core_files test_pool_info_query always_fails
```
to the commit message in this PR.

But that is not actually possible yet.  Again, surprisingly.  However if you land #413 to this PR (this PR is #413's base, so simply merge it as you would any other pipeline-lib PR and it will get merged to this PR, not `master`) you will be able to do the above with `Test-tag:` and be sure your change has the desired effect.  If you want.  Just let me know if you don't want to and I will rebase #413 to master and land it independently.

brianjmurrell (approved):

JohnMalmberg:
As I am about to start another vacation, I would prefer to land this minor change as is.

PR #415, opened 2024-01-10
Get hardware distro version from map

brianjmurrell:


JohnMalmberg (dismissed):

vars/pragmasToEnv.groovy
```diff
@@ -25,6 +25,9 @@ String call(String commit_message) {
         }
     }
 
+    // put the pragms into the environment
+    env.pragmas = pragmas
```
brianjmurrell, line 29: @JohnMalmberg Does it surprise you that this works?  It should, because the best I can determine, it doesn't, unfortunately.  My testing of this was flawed.  :frowning:  It doesn't seem that a shared-lib function can modify the environment like this.  Unless you know otherwise.

Are you aware of any way to achieve what this is trying to achieve?

brianjmurrell: I take that back.  It DOES work!

My original test was successful.  It was a subsequent test that was failing and that's probably due to a bug.  So it was useful testing in the end.

JohnMalmberg (approved):

vars/getFunctionalTestStage.groovy
```diff
@@ -60,7 +60,7 @@ Map call(Map kwargs = [:]) {
             } else {
                 node(cachedCommitPragma("Test-label${pragma_suffix}", label)) {
                     // Ensure access to any branch provisioning scripts exist
-                    echo "[${name}] Check out from version control"
+                    echo "[${name}] Check out '${base_branch}' from version control"
                     if (base_branch) {
                         checkoutScm(
                             url: 'https://github.com/daos-stack/daos.git',
```
JohnMalmberg, line 66: Wouldn't it be better to put that inside the if statement with and else clause explaining why?
Or does that not matter?

brianjmurrell: It could have been.  I thought it would ultimately be obvious which branch it took/why based on the value of `$base_branch`.  Wasn't worth pushing a new iteration and waiting for all of the testing on to make that difference at least.

PR #423, opened 2024-02-27
DAOS-15150 test: To implement bullseye_generate_report and fix error

dinghwah:
Update runTestFunctionalV2.groovy with bullseye test.cov stash

Signed-off-by: Ding Ho ding-hwa.ho@intel.com


vars/functionalTest.groovy
```diff
@@ -137,6 +137,15 @@ Map call(Map config = [:]) {
             error.getMessage())
     }
 
+    if (!env.BULLSEYE) {
```
phender, line 140: We want this code block to be executed if `env.BULLSEYE` is set:
```suggestion
    if (env.BULLSEYE) {
```

phender (changes requested):

PR #425, opened 2024-03-04
DAOS-15232 ci: support Test-tag and Features pragmas together

daltonbohning:
Support both Test-tag and Features pragmas by always looking for Features and appending to the Test-tag

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -546,16 +546,22 @@ pipeline {
                                        tag_template: '@commits.value@,@stages.tag@'],
                                        [tags: [[tag: 'Features', value: 'datamover']],
                                         tag_template: 'pr,@stages.tag@ ' +
+                                                      'pr,@commits.value@,@stages.tag@ ' +
                                                       'daily_regression,@commits.value@,@stages.tag@ ' +
                                                       'full_regression,@commits.value@,@stages.tag@'],
                                        /* groovylint-disable-next-line DuplicateMapLiteral */
                                        [tags: [[tag: 'Test-tag', value: 'datamover'],
                                                [tag: 'Features', value: 'foobar']],
-                                        tag_template: '@commits.value@,@stages.tag@'],
+                                        tag_template: '@commits.value@,@stages.tag@' +
+                                                      'pr,foobar,@stages.tag@ ' +
+                                                      'daily_regression,foobar,@stages.tag@ ' +
+                                                      'full_regression,foobar,@stages.tag@'],
                                        [tags: [[tag: 'Features', value: 'datamover foobar']],
                                         tag_template: 'pr,@stages.tag@ ' +
+                                                      'pr,datamover,@stages.tag@ ' +
                                                       'daily_regression,datamover,@stages.tag@ ' +
                                                       'full_regression,datamover,@stages.tag@ ' +
+                                                      'pr,foobar,@stages.tag@ ' +
                                                       'daily_regression,foobar,@stages.tag@ ' +
                                                       'full_regression,foobar,@stages.tag@'],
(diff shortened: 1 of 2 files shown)


Jenkinsfile
```diff
@@ -17,7 +17,7 @@
 // That PR should be landed with out deleting the PR branch.
 // Then a second PR submitted to comment out the @Library line, and when it
 // is landed, both PR branches can be deleted.
-//@Library(value='pipeline-lib@my_branch_name') _
+@Library(value='pipeline-lib@dbohning/daos-15232') _
```
daltonbohning, line 20: Need to backout before merge.

daltonbohning:
From a high level, this is the expectation
```
Test-tag: foo
Features: bar
-> foo bar,pr bar,daily_regression bar,full_regression
```

```
Features: bar
-> <default-test-tag> bar,pr bar,daily_regression bar,full_regression
```
Where `default-test-tag` is currently either `default_tags` as defined in the Jenkinsfile args, or `pr` to be backward compatible.
In the future, `default-test-tag` could come from the auto-tag calculation.

brianjmurrell (approved):

Jenkinsfile
daltonbohning: Could this work?
```
@Library(value='pipeline-lib@${env.BRANCH_NAME}') _
```

I could test this here, since I need to backout this change and re-push anyway. And if it doesn't work, the commit pragma test will fail. So there's little risk of running too much extra testing if it fails.

brianjmurrell: I am pretty sure we tried that and the result was unfortunately not.  Because it happens so early in the process, I believe it has to be a literal.  No variables or code, etc.

daltonbohning:
The upstream weekly-testing failed
https://build.hpdd.intel.com/blue/organizations/jenkins/daos-stack%2Fdaos/detail/ci-daos-stack-pipeline-lib-PR-425-weekly-testing/1/pipeline/79
But I do see
```
[Functional Hardware Medium] Running functionalTest() on ci_nvme5 with tags=load_mpi,hw,medium,-provider test_core_files,hw,medium,-provider test_pool_info_query,hw,medium,-provider
```
So the tags seem okay

Jenkinsfile
daltonbohning: I'm not at all literate with Jenkins, but this could work for us?
https://stackoverflow.com/questions/76799131/how-to-test-jenkins-shared-library-with-jenkins-itself

daltonbohning: Although I imagine there might be a security reason why it doesn't.

brianjmurrell: > I'm not at all literate with Jenkins, but this couldn't work for us? [stackoverflow.com/questions/76799131/how-to-test-jenkins-shared-library-with-jenkins-itself](https://stackoverflow.com/questions/76799131/how-to-test-jenkins-shared-library-with-jenkins-itself)

So yes, disabling implicit loading is one option that I am aware of.  It needs some **thorough** testing on a non-production Jenkins (because it's disabling the global property of disabling implicit library loading) to understand all of it's ramifications though and I don't think our staging Jenkins has been in any usable shape for being able to do this for a long time.

daltonbohning: Thanks for the explanation. I think probably bigger issues to focus on for now :)

daltonbohning: Another idea for unit test at least is maybe to put them in a separate groovy script so we don't have to modify Jenkinsfile to update the commit pragma tests

phender (changes requested):
Changes approved, but this can't land with the pipeline-lib self reference.

phender (approved):

PR #428, opened 2024-03-12
DAOS-15411 ci: librarify some pragma functions

daltonbohning:
Add envToPragmas.groovy to convert env.pragmas to a Map.
Add pragmasToMap.groovy to convert string commit message to a Map.
Adjust existing usages to use new functions.


daltonbohning:
This will be useful in the `daos` Jenkinsfile for working with `env.pragmas`

brianjmurrell (changes requested):
LGTM.  Would be nice to add a test that both forms of `pragmasToEnv()` both set the environment variable and return the pragmas as a string.

We should also get the master testing to run as that's where all of this stuff is used (the most).

Just -1 until we can get that testing.

daltonbohning:
Testing in a DAOS PR, I see some unexpected behavior. So definitely need tests for this

daltonbohning:
Please feel free to unsubscribe while I work through the unit test

vars/envToPragmas.groovy
```diff
@@ -0,0 +1,18 @@
...
+ * envToPragmas.groovy
+ *
+ * envToPragmas variable
+ */
+
+Map call() {
+    Map pragmas = [:]
+    if (env.pragmas)
+        pragmas = "${env.pragmas}"[1..-2].split(', ').collectEntries { entry ->
+            String[] pair = entry.split('=', 2)
```
daltonbohning, line 14: The previous version did
`entry.split('= ')`
So this version explicitly trims the value

vars/envToPragmas.groovy
```diff
@@ -0,0 +1,18 @@
...
+            [(pair.first().trim()): pair.last().trim()]
+    }
+    return pragmas
+}
```
daltonbohning, line 18: @brianjmurrell The previous version of this was doing
```
Map pragmas = "${env.pragmas}"[1..-2].split(', ').collectEntries { entry ->
            String[] pair = entry.split('= ')
            [(pair.first()): pair.last()]
        }
```
Which implicitly strips the first space from the value. I.e. it splits on `"= "`
This new version should still work similar, except it also allows the value to not start with a space.
But for some reason I've yet to explain, I think some other version of this this function is being used when a `daos` PR runs.

For example, in my `daos` PR, in the get commit message stage:
```
env.pragmas after updating
{skip-build=true, skip-test=true, skip-tests=true, required-githooks=true, signed-off-by=Dalton Bohning <dalton.bohning@intel.com>, test-tag= DfuseMUPerms DmvrDstCreate pool pr}
```

Notice there is NOT a space after the equals sign. That's because `envToPragmas` strips it, and then when the Map is converted back to a string in `env.pragmas`, the space is gone.

The problem is when we then try to parse that `env.pragmas` a second time. I did verify with a unit test that this code can parse and update multiple times, but on my `daos` PR, the `skip-build:true` pragma is being ignored.

I suspect we're actually calling some other version of this function. Maybe the one in `system-pipeline-lib`?

daltonbohning: I've tested this code locally and in the new unit test and both behave as expected.

daltonbohning: @brianjmurrell, @phender and I added more debugging and confirmed that a `daos` PR is in fact using `cachedCommitPragma` from `system-pipeline-lib`, NOT `pipeline-lib`

daltonbohning:
Just the 2.4 NLT is failing due to https://daosio.atlassian.net/browse/DAOS-15416

daltonbohning:
"good" run except NLT failure:
https://build.hpdd.intel.com/blue/organizations/jenkins/daos-stack%2Fpipeline-lib/detail/PR-428/34/pipeline

Runs without the self-reference fail because the `selfUnitTest` does not exist in master

daltonbohning:
With https://github.com/daos-stack/system-pipeline-lib/commit/b7281dc7ec3c11944bee848bf7a8e7338dd971ad
I can re-add the updatePragmas, which works now in a `daos` PR

vars/envToPragmas.groovy
```diff
@@ -0,0 +1,18 @@
...
+ *
+ * envToPragmas variable
+ */
+
+Map call() {
+    Map pragmas = [:]
+    if (env.pragmas)
+        pragmas = "${env.pragmas}"[1..-2].split(', ').collectEntries { entry ->
+            String[] pair = entry.split('=', 2)
+            [(pair.first()): pair.last()]
```
daltonbohning, line 15: The value is now trimmed when parsing the commit message. So no need to trim here. And no need to match on `= ` anymore.

vars/pragmasToMap.groovy
```diff
@@ -0,0 +1,29 @@
...
+    // can't use eachLine() here: https://issues.jenkins.io/browse/JENKINS-46988/
+    commit_message.split('\n').each { line ->
+        String key, value
+        try {
+            (key, value) = line.split(':', 2)
+            if (key.contains(' ')) {
+                // this returns from the .each closure, not the method
+                return
+            }
+            pragmas[key.toLowerCase()] = value.trim()
```
daltonbohning, line 21: Always trim the value before adding to the map. It remains trimmed when converting a string

vars/updatePragmas.groovy
daltonbohning, line 1: It's possible to merge `updatePragmas` and `pragmasToEnv` into a single function, but would be difficult to maintain backwards compatibility that way

daltonbohning:
Latest clean run with self-reference: https://build.hpdd.intel.com/blue/organizations/jenkins/daos-stack%2Fpipeline-lib/detail/PR-428/45/pipeline/1927/

daltonbohning:
Latest run failed unit test as expected because the self-reference is gone:
https://build.hpdd.intel.com/blue/organizations/jenkins/daos-stack%2Fpipeline-lib/detail/PR-428/46/pipeline

daltonbohning:
@brianjmurrell @phender Ready for review again

phender (dismissed):

vars/selfUnitTest.groovy
```diff
@@ -0,0 +1,112 @@
...
+    println("Test pragmasToMap")
+
+    println("  with commit message")
+    result_map = pragmasToMap(commit_message)
+    println("    result_map   = ${result_map}")
+    println("    expected_map = ${expected_map}")
+    assert(result_map == expected_map)
+
+    println("  with empty string")
+    result_map = pragmasToMap("")
```
brianjmurrell, line 50: Maybe ditto here?  So perhaps we need the commit message in parts.  One static part for title and description and we add the pragmas, or no pragmas or garbage to that in each test?

vars/selfUnitTest.groovy
```diff
@@ -0,0 +1,112 @@
...
+    env_pragmas_original = env.pragmas
+    try {
+        _test_pragmas_env()
+    } finally {
+        env.pragmas = env_pragmas_original
+    }
+}
+
+void _test_pragmas_env() {
+    commit_message = '''Skip-build: true
```
brianjmurrell, line 22: We should probably make this commit message look more _real_, with a title and at least one line of description.

Maybe one day we will enforce that commit pragmas are only looked at after a title and description and this test should be ready for that.  It also ensures that a commit message with a title and description (which will be the usual case) also passes this test.

vars/updatePragmas.groovy
```diff
@@ -0,0 +1,24 @@
...
+    Map new_pragmas = pragmasToMap(commit_message)
+
+    new_pragmas.each { entry ->
+        if (overwrite || !current_pragmas.containsKey(entry.key)) {
+            current_pragmas[entry.key] = entry.value.trim()
+        }
+    }
+
+    // put the pragmas into the environment as a String
+    // note this converts the Map to a string in the format "{foo= bar, bat= ball}"
```
brianjmurrell, line 21: This is not true any more is it?

```suggestion
    // note this converts the Map to a string in the format "{foo=bar, bat=ball}"
```

Jenkinsfile
```diff
@@ -690,6 +690,11 @@ pipeline {
                         }
                     } // steps
                 } // stage ('Commit Pragma tests')
+                stage('Self Unit Test') {
+                    steps {
+                        selfUnitTest()
```
brianjmurrell, line 695: Note that as suspected, this still requires the `@Library` self-reference to run [any new code (in the future) in] the `selfUnitTest()` function.  Sadly.

vars/selfUnitTest.groovy
```diff
@@ -0,0 +1,112 @@
...
+    env.pragmas = ""
+    result_map = envToPragmas()
+    println("    result_map   = ${result_map}")
+    println("    expected_map = ${[:]}")
+    assert(result_map == [:])
+    env.pragmas = env_pragmas_tmp
+
+
+    println("Test updatePragmas")
+    println("  with override=true")
```
brianjmurrell, line 89:
```suggestion
    println("  with overwrite=true")
```

vars/selfUnitTest.groovy
```diff
@@ -0,0 +1,112 @@
...
+
+    expected_map["test-tag"] = "foo2"
+    updatePragmas("Test-tag: foo2", true)
+    result_map = envToPragmas()
+    println("    result_map   = ${result_map}")
+    println("    expected_map = ${expected_map}")
+    assert(result_map == expected_map)
+
+    println("  with override=false")
+    updatePragmas("Test-tag: should_not_override", false)
```
brianjmurrell, line 107:
```suggestion
    println("  with overwrite=false")
    updatePragmas("Test-tag: should_not_overwrite", false)
```

brianjmurrell (changes requested):

vars/selfUnitTest.groovy
daltonbohning: Yes, I can do that

vars/selfUnitTest.groovy
daltonbohning: I don't follow here. Do you mean have multiple arguments to the pragmasToMap function?

vars/updatePragmas.groovy
daltonbohning: Thanks, will update

vars/selfUnitTest.groovy
daltonbohning: Or do you mean test cases for each of these, for example?
```
DAOS-XXXX test: my commit title
```

```
DAOS-XXXX test: my commit title

<pragmas>
```

```
some garbage
```

brianjmurrell: > I don't follow here. Do you mean have multiple arguments to the pragmasToMap function?

I just mean that given that pragamsToMap() takes a commit message as it's argument, having it empty here doesn't represent what it will see when being used.

I was just further suggesting that for ease, you might want to build the commit messages passed to these various tests from a static _Tltle/Description_ string you define once and then append another string with the various pragmas you want to test.  Just to reduce the duplication of the Title/Description part.

daltonbohning: > having it empty here doesn't represent what it will see when being used.

In practice, yes. But testing with an empty commit message validates the code path where `pragamsToMap` is given an empty string. Part of that test stems from me seeing different behavior with this Jenkins/groovy vs my local groovy install.

> build the commit messages passed ...

Sorry, yes, I understand what you mean now :)

brianjmurrell: > In practice, yes. But testing with an empty commit message validates the code path where `pragamsToMap` is given an empty string.

But given that `pragmasToMap()` is given a _commit message_, when could that ever be empty?  I don't think you can create a git commit without any message at all.

daltonbohning: In practice, it won't. But the function implementation itself still supports that code path. So having a test for it verifies the function behaves as expected. IMO, it doesn't matter if it's used in practice or not. If the code path is possible, it should ideally be tested. And the fewer assumptions this function has to make about a commit message, the more robust and futureproof it is

brianjmurrell: > In practice, it won't. But the function implementation itself still supports that code path.

That's a good point.  But then we should ask should it?  Given that we can assume that a commit message should never be empty, if that function gets an empty commit message argument, that ought to be an error by the caller, and called out.  Maybe that function should `error()` if the argument is empty.  We probably don't do enough of that kind of argument checking in the library, but we should.

daltonbohning: I've been erring on the side of caution here and trying to minimally change behavior. If we want this to be an error, can we do that separately? TBH I'm not familiar with groovy `error()` so it would probably take me a few cycles to figure out the right code+test

brianjmurrell: > I've been erring on the side of caution here and trying to minimally change behavior.

I don't think that adding validation that should have been there in the first place is necessarily changing behavior.  It's only becoming obvious it's missing as you are adding the testing now and adding test cases for what shouldn't even be considered valid input.  So probably the test case is still valid but just tests that the function does `error()` out.

> If we want this to be an error, can we do that separately? TBH I'm not familiar with groovy `error()` so it would probably take me a few cycles to figure out the right code+test

`error()` is not groovy per-se but is a Jenkins step function.  It's simply `error('This function requires a valid commit message as it's first argument')`.  That's it.  That stops the pipeloine at that point, displays the error and bails.

So for the purposes of testing, you would need to mock it.  I believe you will find an example of that in some of the (probably commented out) U/T hackery that has been added to (some) other function(s).

daltonbohning: > for the purposes of testing, you would need to mock it

That's exactly the kind of rabbit hole I'm trying to avoid here. I've already spent _**a lot more**_ time on this than I thought I needed, and have already fixed multiple existing issues.

brianjmurrell: > That's exactly the kind of rabbit hole I'm trying to avoid here. I've already spent _**a lot more**_ time on this than I thought I needed, and have already fixed multiple existing issues.

I can certainly understand that.  But TBH, the argument validation (for a completely empty commit) and the resulting `error()` if it is empty seems straightforward enough to not really even need a test case, IMO.

daltonbohning: If we're okay with not having a test case, then sure. Just updated

daltonbohning:
Got a clean run with the self-reference:
https://build.hpdd.intel.com/blue/organizations/jenkins/daos-stack%2Fpipeline-lib/detail/PR-428/51/pipeline/1921/

@brianjmurrell is this adequate before I push again to remove the self-reference?

daltonbohning:
Full clean run: https://build.hpdd.intel.com/blue/organizations/jenkins/daos-stack%2Fpipeline-lib/detail/PR-428/51/pipeline

Fun with self-reference removed failed as expected: https://build.hpdd.intel.com/blue/organizations/jenkins/daos-stack%2Fpipeline-lib/detail/PR-428/52/pipeline

phender (approved):

vars/selfUnitTest.groovy
```diff
@@ -19,44 +19,41 @@ void call() {
 }
 
 void _test_pragmas_env() {
-    commit_message = '''Skip-build: true
+    String commit_title = "DAOS-XXXX test: short description about thing"
+    String commit_desc = "Long description about thing\nwith multiple lines"
+    String commit_pragmas = """Skip-build: true
```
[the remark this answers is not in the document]
daltonbohning: I'd rather fix those separately so I can move forward with this

PR #434, opened 2024-04-08
SRE-2155 ci: Fix typo in parameter name to be consistent with main Jenkins file

grom72:
DAOS Jenkins uses `CI_DEB_Ubuntu20_NOBUILD` parameter
https://github.com/daos-stack/daos/blob/ff4375a885a316cefcf6279d752a42eaf9ea197e/Jenkinsfile#L238
It does not work as pipeline-lib expect `CI_RPM_Ubuntu20_NOBUILD`

`Skip-build-ubuntu20-deb` pragma is modified for consistency.

### DIFF
--- a/vars/skipStage.groovy
+++ b/vars/skipStage.groovy
@@ -225,7 +225,7 @@ boolean call(Map config = [:]) {
                    (docOnlyChange(target_branch) &&
                     prRepos('ubuntu20') == '') ||
                    prReposContains('ubuntu20', jobName()) ||
-                   skip_stage_pragma('build-ubuntu20-rpm')
+                   skip_stage_pragma('build-ubuntu20-deb')
         case 'Build on CentOS 8':
         case 'Build on EL 8':
         case 'Build on EL 8.8':
--- a/vars/skipStage.groovy
+++ b/vars/skipStage.groovy
@@ -220,7 +220,7 @@ boolean call(Map config = [:]) {
                    prReposContains('leap15', jobName()) ||
                    skip_stage_pragma('build-leap15-rpm')
         case 'Build DEB on Ubuntu 20.04':
-            return paramsValue('CI_RPM_Ubuntu20_NOBUILD', false) ||
+            return paramsValue('CI_RPM_ubuntu20_NOBUILD', false) ||
                    paramsValue('CI_DEB_Ubuntu20_NOBUILD', false) ||
                    target_branch =~ branchTypeRE('weekly') ||
                    (docOnlyChange(target_branch) &&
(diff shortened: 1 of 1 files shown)


brianjmurrell (approved):

JohnMalmberg (dismissed):

brianjmurrell:
> SRE-2155 ci: fix typo in CI_RPM_ubuntu20_NOBUILD parameters name.

Just wondering @grom72, why the empty commit?

grom72:
> > SRE-2155 ci: fix typo in CI_RPM_ubuntu20_NOBUILD parameters name.
> 
> Just wondering @grom72, why the empty commit?

To add Jira ticket and avoid push -f

vars/skipStage.groovy
```diff
@@ -220,12 +220,12 @@ boolean call(Map config = [:]) {
...
                    skip_stage_pragma('build-leap15-rpm')
         case 'Build DEB on Ubuntu 20.04':
-            return paramsValue('CI_RPM_ubuntu20_NOBUILD', false) ||
+            return paramsValue('CI_DEB_Ubuntu20_NOBUILD', false) ||
                    target_branch =~ branchTypeRE('weekly') ||
                    (docOnlyChange(target_branch) &&
                     prRepos('ubuntu20') == '') ||
                    prReposContains('ubuntu20', jobName()) ||
-                   skip_stage_pragma('build-ubuntu20-rpm')
+                   skip_stage_pragma('build-ubuntu20-deb')
```
brianjmurrell, line 228: This was intentionally left as `rpm` due to the documenting of it as such at https://daosio.atlassian.net/wiki/spaces/CI/pages/1389759093/Commit+Pragmas.  I have no inherent objection to this change but the documentation needs updating at least.

You could also allow for both postfixes here so that you don't trip up anyone that is using the existing documented pragma.

Ultimately this should have probably been `-pkg` to be generic.

brianjmurrell (approved):

vars/skipStage.groovy
grom72: @brianjmurrell Do you think such a change (`-rpm` to `-pkg`) would be acceptable if both documentation and pipeline will be updated?

brianjmurrell: I think we could deprecate `*-rpm` in favour of `-*pkg` by adding (not replacing -- yet) `*-pkg` as an additional condition and then updating the documentation to specify that `*-pkg` is not preferred and that `*-rpm` is deprecated.

grom72: I propose to do this in a separate ticket as it requires more changes in several files.
I roll back the last commit and will create a new one dedicated to replacing `-rpm` with`-pkg`

grom72:
We need this PR on master to finalize work on #439.

grom72:
Already implemented by #439

PR #435, opened 2024-05-06
DAOS-623 ci: Don't change '-' to '_' in variable

jolivier23:
The name in Jenkinsfile has '-'.  Changing it to '_' means the defaults specified in Jenkinsfile don't actually get used for variables with inline dashes.

### DIFF
--- a/vars/skipStage.groovy
+++ b/vars/skipStage.groovy
@@ -102,7 +102,7 @@ boolean skip_ftest_hw(String size, String target_branch, String tags) {
         return false
     }
     String distro = (hwDistroTarget(size) =~ /([a-z]+)(\d+)(\.\d+)?/)[0][1..2].join('')
-    return !paramsValue('CI_' + size.replace('-', '_') + '_TEST', true) ||
+    return !paramsValue('CI_' + size + '_TEST', true) ||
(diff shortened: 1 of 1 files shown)


jolivier23:
https://github.com/daos-stack/daos/blob/76a8b59850739956a8584d0caa43b579ccb64825/Jenkinsfile#L276-L293

vars/skipStage.groovy
```diff
@@ -102,7 +102,7 @@ boolean skip_ftest_hw(String size, String target_branch, String tags) {
         return false
     }
     String distro = (hwDistroTarget(size) =~ /([a-z]+)(\d+)(\.\d+)?/)[0][1..2].join('')
-    return !paramsValue('CI_' + size.replace('-', '_') + '_TEST', true) ||
+    return !paramsValue('CI_' + size + '_TEST', true) ||
```
brianjmurrell, line 105: Has this always been true though?  I suspect to ensure backwards compatibility you want to test both:

```suggestion
    return !paramsValue('CI_' + size + '_TEST', true) ||
           !paramsValue('CI_' + size.replace('-', '_') + '_TEST', true) ||
```

brianjmurrell:
daos-stack/daos#14325

jolivier23:
Going another direction

PR #436, opened 2024-05-08
DAOS-15772 test: Support provider independence from stage name

phender:
Allow defing a UCX test stage with getFunctionalTestStage that will be provisioned correctly w/o the need for a specific stage name format.


phender:
The new parameter is being used/tested in https://github.com/daos-stack/daos/pull/14305

JohnMalmberg (approved):

daltonbohning (approved):

PR #444, opened 2024-08-28
SRE-2453: skip all builds and tests for doc-only PR and unify Build DEB Ubuntu stage names

grom72:
- [x] `@Library(["pipeline-lib@grom72/SRE-2453", "trusted-pipeline-lib@grom72/SRE-2453"]) _` to be removed before landing.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -148,15 +148,9 @@ pipeline {
                 stage('daosLatestVersion() tests') {
                     steps {
                         script {
-                            daosLatestVersionMaster = daosLatestVersion('master', 'el8')
-                            assert(daosLatestVersionMaster.matches(/2.7\.\d+.*/)), 
-                                '"' + daosLatestVersionMaster + '" <> matches(/2.7\\.\\d+.*/)'
-                            daosLatestVersion24 = daosLatestVersion('release/2.4', 'el8')
-                            assert(daosLatestVersion24.matches(/2.[34]\.\d+.*/)),
-                                '"' + daosLatestVersion24 + '" <> matches(/2.[34]\\.\\d+.*/)'
-                            daosLatestVersion26 = daosLatestVersion('release/2.6', 'el8')
-                            assert(daosLatestVersion26.matches(/2.[56]\.\d+.*/)), 
-                                '"' + daosLatestVersion26 + '" <> matches(/2.[56]\\.\\d+.*/)'
+                            assert(daosLatestVersion('master', 'el8').matches(/2.7\.\d+.*/))
+                            assert(daosLatestVersion('release/2.4', 'el8').matches(/2.[34]\.\d+.*/))
+                            assert(daosLatestVersion('release/2.6', 'el8').matches(/2.[56]\.\d+.*/))
                         }
                     }
                 }
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -18,8 +18,7 @@
 // Then a second PR submitted to comment out the @Library line, and when it
 // is landed, both PR branches can be deleted.
 //@Library(value='pipeline-lib@my_branch_name') _
-@Library(value="pipeline-lib@grom72/SRE-2453") _
-@Library(value="trusted-pipeline-lib@grom72/SRE-2453") _
+@Library(["pipeline-lib@grom72/SRE-2453", "trusted-pipeline-lib@grom72/SRE-2453"]) _
 
 /* groovylint-disable-next-line CompileStatic */
 job_status_internal = [:]
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -149,14 +149,11 @@ pipeline {
                     steps {
                         script {
                             assert(daosLatestVersion('master', 'el8').matches(/2.7\.\d+.*/)),
-                                '"' + daosLatestVersion('master', 'el8') + 
-                                '" <> matches(/2.7\.\d+.*/)'
+                                daosLatestVersion('master', 'el8') + ' <> matches(/2.7\.\d+.*/)'
                             assert(daosLatestVersion('release/2.4', 'el8').matches(/2.[34]\.\d+.*/)),
-                                '"' + daosLatestVersion('release/2.4', 'el8') + \
-                                '" <> matches(/2.[34]\.\d+.*/)'
+                                daosLatestVersion('release/2.4', 'el8') + ' <> matches(/2.[34]\.\d+.*/)'
                             assert(daosLatestVersion('release/2.6', 'el8').matches(/2.[56]\.\d+.*/)),
-                                '"' + daosLatestVersion('release/2.6', 'el8') + \
-                                '" <> matches(/2.[56]\.\d+.*/)'
+                                daosLatestVersion('release/2.6', 'el8') + ' <> matches(/2.[56]\.\d+.*/)'
                         }
                     }
                 }
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -148,9 +148,9 @@ pipeline {
                 stage('daosLatestVersion() tests') {
                     steps {
                         script {
+                            assert(daosLatestVersion('master', 'el8').matches(/2.7\.\d+.*/))
                             assert(daosLatestVersion('release/2.4', 'el8').matches(/2.[34]\.\d+.*/))
                             assert(daosLatestVersion('release/2.6', 'el8').matches(/2.[56]\.\d+.*/))
-                            assert(daosLatestVersion('master', 'el8').matches(/2.7\.\d+.*/))
                         }
                     }
                 }
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -19,6 +19,7 @@
 // is landed, both PR branches can be deleted.
 //@Library(value='pipeline-lib@my_branch_name') _
 @Library(value="pipeline-lib@grom72/SRE-2453") _
+@Library(value="trusted-pipeline-lib@grom72/SRE-2453") _
 
 /* groovylint-disable-next-line CompileStatic */
 job_status_internal = [:]
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -148,16 +148,15 @@ pipeline {
                 stage('daosLatestVersion() tests') {
                     steps {
                         script {
-                            assert(daosLatestVersion('master', 'el8').matches(/2.7\.\d+.*/),
-                                '"' + daosLatestVersion('master', 'el8') + \
-                                '" <> matches(/2.7\.\d+.*/)' )
-                            )
-                            assert(daosLatestVersion('release/2.4', 'el8').matches(/2.[34]\.\d+.*/),
+                            assert(daosLatestVersion('master', 'el8').matches(/2.7\.\d+.*/)),
+                                '"' + daosLatestVersion('master', 'el8') + 
+                                '" <> matches(/2.7\.\d+.*/)'
+                            assert(daosLatestVersion('release/2.4', 'el8').matches(/2.[34]\.\d+.*/)),
                                 '"' + daosLatestVersion('release/2.4', 'el8') + \
-                                '" <> matches(/2.[34]\.\d+.*/)' )
-                            assert(daosLatestVersion('release/2.6', 'el8').matches(/2.[56]\.\d+.*/))
+                                '" <> matches(/2.[34]\.\d+.*/)'
+                            assert(daosLatestVersion('release/2.6', 'el8').matches(/2.[56]\.\d+.*/)),
                                 '"' + daosLatestVersion('release/2.6', 'el8') + \
-                                '" <> matches(/2.[56]\.\d+.*/)' )
+                                '" <> matches(/2.[56]\.\d+.*/)'
                         }
                     }
                 }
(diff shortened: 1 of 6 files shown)


brianjmurrell:
Did you test this change against your `Doc-only: true` PR to validate that it does what we want?

grom72:
> Did you test this change against your `Doc-only: true` PR to validate that it does what we want? Since it is changes to make the pipeline bail out early it shouldn't be too long to validate.

Yes, 
https://build.hpdd.intel.com/job/daos-stack/job/pmdk/job/PR-40/4/pipeline-graph/

vars/skipStage.groovy
```diff
@@ -383,12 +384,13 @@ boolean call(Map config = [:]) {
...
-                    !startedByUpstream() &&
-                    !startedByUser()) ||
-                   skip_if_unstable()
+                    skip_stage_pragma('test') ||
+                    docOnlyChange(target_branch ||
+                    (env.BRANCH_NAME =~ branchTypeRE('testing') &&
+                     !startedByTimer() &&
+                     !startedByUpstream() &&
+                     !startedByUser()) ||
+                     skip_if_unstable())
```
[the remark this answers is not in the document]
grom72: By mistake.

brianjmurrell (approved):

grom72:
> Nice. But really, the `Doc-only: true` in the commit message should only be necessary the minority (if any) of the cases and most of the time that state should be automatically detected. Can you push a new empty commit to that PR without the `Doc-only: true` and see if that is automatically determined? Regardless of the outcome, this PR is looking good, but it would just be interesting to know if that auto-detection is (still) working.

Seems that auto-detection does not work.
https://build.hpdd.intel.com/job/daos-stack/job/pmdk/job/PR-40/5/pipeline-graph/

brianjmurrell:
> Seems that auto-detection does not work. [build.hpdd.intel.com/job/daos-stack/job/pmdk/job/PR-40/5/pipeline-graph](https://build.hpdd.intel.com/job/daos-stack/job/pmdk/job/PR-40/5/pipeline-graph/)

Hrm, yes.  I suppose that's due to https://github.com/daos-stack/pipeline-lib/blob/83b1561aeb07524a0aa0a5b99550f28d103980ed/vars/docOnlyChange.groovy#L18 since dependency repos don't typically have that file, nor even a `ci/` sub-directory in them.

Probably [some form of] [ci/doc_only_change.sh](https://github.com/daos-stack/daos/blob/master/ci/doc_only_change.sh) could be added to [vars/docOnlyChange.groovy](https://github.com/daos-stack/pipeline-lib/blob/grom72/SRE-2453/vars/docOnlyChange.groovy) as a fall-back when there is no `ci/doc_only_change.sh`.

grom72:
> > Seems that auto-detection does not work. [build.hpdd.intel.com/job/daos-stack/job/pmdk/job/PR-40/5/pipeline-graph](https://build.hpdd.intel.com/job/daos-stack/job/pmdk/job/PR-40/5/pipeline-graph/)
> 
> Hrm, yes. I suppose that's due to
> 
> https://github.com/daos-stack/pipeline-lib/blob/83b1561aeb07524a0aa0a5b99550f28d103980ed/vars/docOnlyChange.groovy#L18
> 
> since dependency repos don't typically have that file, nor even a `ci/` sub-directory in them.
> Probably [some form of] [ci/doc_only_change.sh](https://github.com/daos-stack/daos/blob/master/ci/doc_only_change.sh) could be added to [vars/docOnlyChange.groovy](https://github.com/daos-stack/pipeline-lib/blob/grom72/SRE-2453/vars/docOnlyChange.groovy) as a fall-back when there is no `ci/doc_only_change.sh`.

Done.

vars/docOnlyChange.groovy
```diff
@@ -16,8 +16,19 @@ boolean call(String target_branch) {
...
+                  script: ["TARGET_BRANCH=${target_branch}",
+                            '''
+                            set -uex
+                            if ! git fetch origin ${TARGET_BRANCH}; then
+                                echo "Hrm.  Got an error fetching the target branch"
+                                exit 0
+                            fi
+                            git diff --no-commit-id --name-only origin/${TARGET_BRANCH} HEAD |
+                               grep -v -e "^docs/" -e "\\.md$" -e "^.*LICENSE.*$"
+                            '''].join(' '),
```
brianjmurrell, line 29: While a ` ` `join()`ed list works, we have typically just gone the route of a single string such as:

```suggestion
                  script: 'TARGET_BRANCH= ' + target_branch + '''
                            set -uex
                            if ! git fetch origin ${TARGET_BRANCH}; then
                                echo "Hrm.  Got an error fetching the target branch"
                                exit 0
                            fi
                            git diff --no-commit-id --name-only origin/${TARGET_BRANCH} HEAD |
                               grep -v -e "^docs/" -e "\\.md$" -e "^.*LICENSE.*$"
                            ''',
```

Although lines 34-36 are not lost on me.  I'm not entirely sure who wrote those and indeed, given that both of these if branches are duplicating the execution of an `sh()` step with the same arguments this whole function could be DRYed down into just setting the `script` variable inside the `if` and moving the duplicated parts after the `if` block.

But most importantly(!), adding the fall-back to the first branch of the `if` clause means needing to separate the or'd conditions in that branch as the `cachedCommitPragma('Doc-only').toLowerCase() == 'false'` is absolute and should not execute the script to determine if there are doc-only changes or not.

So something like:

```groovy
    if (cachedCommitPragma('Doc-only').toLowerCase() == 'false' {
        return false
    }

    if (fileExists('ci/doc_only_change.sh')) {
        script = 'CHANGE_ID= ' + env.CHANGE_ID +
                 ' TARGET_BRANCH= ' + target_branch + 
                 ' ci/doc_only_change.sh'
    } else {
        script = 'TARGET_BRANCH= ' + target_branch + '''
                  set -uex
                  if ! git fetch origin ${TARGET_BRANCH}; then
                      echo "Hrm.  Got an error fetching the target branch"
                      exit 0
                  fi
                  git diff --no-commit-id --name-only origin/${TARGET_BRANCH} HEAD |
                     grep -v -e "^docs/" -e "\\.md$" -e "^.*LICENSE.*$"'''
    }

    return sh(label: "Determine if doc-only change",
              script: script
              returnStatus: true) == 1
```

brianjmurrell (changes requested):
Also note that the `daosLatestVersion()` tests are failing.  Is that new in this PR?  It's not supposed to be failing.

vars/docOnlyChange.groovy
grom72: Done

vars/docOnlyChange.groovy
```diff
@@ -14,25 +14,24 @@ boolean call(String target_branch) {
...
+                      exit 0
+                  fi
+                  git diff --no-commit-id --name-only origin/${TARGET_BRANCH} HEAD |
+                     grep -v -e "^docs/" -e "\\.md$" -e "^.*LICENSE.*$"'''
+    }
     return sh(label: "Determine if doc-only change",
-              script: ["CHANGE_ID=${env.CHANGE_ID}",
-                       "TARGET_BRANCH=${target_branch}",
-                       'ci/doc_only_change.sh'].join(' '),
+              script: script
```
brianjmurrell, line 35: Apologies if it was my example that missed this but you need a comma between the arguments here:

```suggestion
              script: script,
```

brianjmurrell (changes requested):

vars/docOnlyChange.groovy
grom72: Done

brianjmurrell (changes requested):
Code changes look good but you have a failing test that needs to be resolved.

https://jenkins-1.wolf.hpdd.intel.com:9090/blue/organizations/jenkins/daos-stack%2Fpipeline-lib/detail/PR-444/23/pipeline/74

Jenkinsfile
```diff
@@ -148,7 +148,11 @@ pipeline {
                 stage('daosLatestVersion() tests') {
                     steps {
                         script {
+                            testVar4 = daosLatestVersion('release/2.4', 'el8')
+                            testVar6 = daosLatestVersion('release/2.6', 'el8')
                             assert(daosLatestVersion('master', 'el8').matches(/2.7\.\d+.*/))
+                            echo "daosLatestVersion('release/2.4', 'el8')" + testVar4
+                            echo "daosLatestVersion('release/2.6', 'el8')" + testVar6
```
brianjmurrell, line 155: Given the example: https://github.com/daos-stack/pipeline-lib/blob/1623921eebb16ec05a104c97105ed8d9515a9b19/Jenkinsfile#L614-L615 there appears to be an `assert()` signature that takes two arguments, where the first is the test and the second is a string that is displayed when the assertion fails.  This second argument is very useful to put the values being tested (i.e. as a string message) into so that those are displayed with the assertion failure.

It would be good to update these `assert()`s to do this so that this kind of after-the-fact debugging doesn't need to be added just to see why the assertion failed.

Jenkinsfile
```diff
@@ -149,7 +149,6 @@ pipeline {
                     steps {
                         script {
                             assert(daosLatestVersion('master', 'el8').matches(/2.7\.\d+.*/))
-                            assert(daosLatestVersion('release/2.4', 'el8').matches(/2.[34]\.\d+.*/))
```
brianjmurrell, line 151: This is not fixing the root problem however.  Indeed. it is masking it and sweeping that under the rug.

The root problem (beyond this test failing) is that if somebody were to want to try to do a CI run with a 2.4 release for example, it would fail due to the reasons discussed previously about needing to also query the `*-archive-local` repository if a release is not found in the `*-stable-local` repository.  Indeed, if that problem were fixed, nothing would need doing here.

Will we ever need to test with an archived release?  Who knows.  But since a code change is needed regardless, it would be better to spend the time fixing that than simply eliminating a test failure that is highlighting a gap in our CI ability.

brianjmurrell (changes requested):

vars/docOnlyChange.groovy
```diff
@@ -31,7 +31,9 @@ boolean call(String target_branch) {
                   git diff --no-commit-id --name-only origin/${TARGET_BRANCH} HEAD |
                      grep -v -e "^docs/" -e "\\.md$" -e "^.*LICENSE.*$"'''
     }
-    return sh(label: "Determine if doc-only change",
+    boolean docOnlyChange = sh(label: "Determine if doc-only change",
               script: script,
               returnStatus: true) == 1
+    println( 'docOnlyChange: ' + docOnlyChange)
+    return docOnlyChange
```
brianjmurrell, line 38: All of this debugging of course will need to be reverted before landing.

Jenkinsfile
```diff
@@ -18,6 +18,7 @@
 // Then a second PR submitted to comment out the @Library line, and when it
 // is landed, both PR branches can be deleted.
 //@Library(value='pipeline-lib@my_branch_name') _
+@Library(["pipeline-lib@grom72/SRE-2453", "trusted-pipeline-lib@grom72/SRE-2453"]) _
```
brianjmurrell, line 21: This of course needs reverting before landing.

vars/docOnlyChange.groovy
```diff
@@ -14,14 +14,26 @@ boolean call(String target_branch) {
...
         return false
     }
-
-    return sh(label: "Determine if doc-only change",
-              script: ["CHANGE_ID=${env.CHANGE_ID}",
-                       "TARGET_BRANCH=${target_branch}",
-                       'ci/doc_only_change.sh'].join(' '),
+    if (fileExists('ci/doc_only_change.sh')) {
+        script = 'CHANGE_ID= ' + env.CHANGE_ID +
+                 ' TARGET_BRANCH= ' + target_branch +
```
brianjmurrell, line 22: You need to remove the spaces after the `=` in this command line build up.

```suggestion
        script = 'CHANGE_ID=' + env.CHANGE_ID +
                 ' TARGET_BRANCH=' + target_branch +
```

brianjmurrell:
Once you have made the requested changes can you please test this PR again with daos-stack/pmdk#40 to make sure it has the desired effect?

brianjmurrell (changes requested):

vars/docOnlyChange.groovy
grom72: Done

brianjmurrell (changes requested):
Just -1 until the [test for this can be done](https://github.com/daos-stack/pipeline-lib/pull/444#issuecomment-2368142587) and the removal of the pipeline-lib PR reference here.

brianjmurrell:
> I can not make the final test of #444 as it requires modification to the Jenkins file, which is not a documentation file.

Indeed.

> I can only show indirectly that the pipeline works as expected:

Which seems sufficient.

Just need to remove the `@Library` reference here now.

Jenkinsfile
grom72: Done

grom72:
> > I can not make the final test of #444 as it requires modification to the Jenkins file, which is not a documentation file.
> 
> Indeed.
> 
> > I can only show indirectly that the pipeline works as expected:
> 
> Which seems sufficient.
> 
> Just need to remove the `@Library` reference here now.
Done

brianjmurrell (approved):

PR #461, opened 2025-03-16
DAOS-623 ci: Update mock to Fedora 41

jolivier23:
Fedora-37 is EOL

### DIFF
--- a/vars/packageBuildingPipeline.groovy
+++ b/vars/packageBuildingPipeline.groovy
@@ -548,7 +548,7 @@ void call(Map pipeline_args) {
                                      ' --cap-add=SYS_ADMIN' +
                                      ' --privileged=true'   +
                                      ' -v /scratch:/scratch'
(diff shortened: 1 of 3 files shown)


JohnMalmberg (changes requested):
https://github.com/daos-stack/daos/pull/15935 has the correct fix for this.
Until that is landed, we can not remove a hack where fedora:latest is pinned to fedora:37.

The mock build for leap15.5 needed to be pinned to fedora/37.

Pinning Fedora to fedora/41 will cause bad effects as specifying any version of fedora will cause it to use the "vault" repos, and fedora/41 is not in the vault repos yet.

Once the above DAOS PR is landed, and build.hpdd.intel.com fixed, then for leap15.6 the pinned fedora version can be removed.

jolivier23:
that's fair, I will close this.  I was just trying some hackery to get my other PR working.

PR #482, opened 2025-11-18
DAOS-17679 build: Support gcov in functional tests

phender:
Optionally provide *.gcno files to functional test stages and stash any code coverage reports generated by the stage.


JohnMalmberg (commented):
See https://github.com/daos-stack/pipeline-lib/pull/481 for handling doing if a stash does not exist so that a missing stash does not abort post processing or archiving of data from a failed run.

PR #505, opened 2026-04-09
SRE-3711: Fix Fault Injection skipping param value

ryon-jensen:


JohnMalmberg (approved):
Wonder why we are not  just using startsWith() instead of all these ORed tests that only differ at the end when we only run have a test stage like fault injection that that only ever runs for one distro/version for a CI run?

ryon-jensen:
> Wonder why we are not just using startsWith() instead of all these ORed tests that only differ at the end when we only run have a test stage like fault injection that that only ever runs for one distro/version for a CI run?

@JohnMalmberg - I actually thought about that and started to, but then got into refactoring the whole function, so just decided to match current style to get it fixed and maybe refactor later.

PR #510, opened 2026-04-29
fix rpm test skip logic

daltonbohning:
- Do not default to true for the skip pragma
- Do not skip test leap 15 RPMs for landing runs

### DIFF
--- a/vars/skipStage.groovy
+++ b/vars/skipStage.groovy
@@ -526,7 +526,7 @@ boolean call(Map config = [:]) {
                    skip_stage_pragma('build-el9-gcc') ||
                    skip_stage_pragma('test') ||
                    skip_stage_pragma('test-rpms') ||
-                   skip_stage_pragma('test-el-9-rpms', 'true') ||
+                   skip_stage_pragma('test-el-9-rpms') ||
                    docOnlyChange(target_branch) ||
                    (quickFunctional() &&
                     !paramsValue('CI_TEST_EL_RPMs', true) &&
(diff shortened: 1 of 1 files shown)


phender (dismissed):

vars/skipStage.groovy
```diff
@@ -526,7 +526,7 @@ boolean call(Map config = [:]) {
                    skip_stage_pragma('build-el9-gcc') ||
                    skip_stage_pragma('test') ||
                    skip_stage_pragma('test-rpms') ||
-                   skip_stage_pragma('test-el-9-rpms', 'true') ||
+                   skip_stage_pragma('test-el-9-rpms') ||
```
daltonbohning, line 530: This was set to true by #496, but this is a required stage for landing

ryon-jensen (dismissed):

phender (approved):

phender:
The https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos/job/ci-daos-stack-pipeline-lib-PR-510-master/1/ build only encountered NLT and Fault Injection warnings.  It's `Test RPMs on EL 9.6` stage ran and passed.

vars/skipStage.groovy
```diff
@@ -545,18 +545,20 @@ boolean call(Map config = [:]) {
...
                    skip_stage_pragma('test') ||
                    skip_stage_pragma('test-rpms') ||
-                   skip_stage_pragma('test-leap-15-rpms', 'true') ||
+                   skip_stage_pragma('test-leap-15-rpms') ||
                    docOnlyChange(target_branch) ||
                    (quickFunctional() &&
                     !paramsValue('CI_RPMS_leap15_TEST', true) &&
                     !run_default_skipped_stage('test-leap-15-rpms')) ||
                    (rpmTestVersion() != '') ||
+                   !startedByLanding() ||
```
daltonbohning: @phender I think this means we can't enable the stage with a pragma on PRs. I.e. if on a PR we do this
```
Skip-test-leap-15-rpms: false
```

Then we will still skip because
```
skip_stage_pragma('test-leap-15-rpms') ||
!startedByLanding() ||
```
is still `True`

phender: Right, we could do this:
```
        case 'Test RPMs on SLES 15':
            if (startedByLanding()) {
                return true
            }
            return !paramsValue('CI_RPMS_leap15_TEST', true) ||
                   target_branch =~ branchTypeRE('weekly') ||
                   skip_stage_pragma('build-leap15-rpm') ||
                   skip_stage_pragma('test') ||
                   skip_stage_pragma('test-rpms') ||
                   skip_stage_pragma('test-leap-15-rpms') ||
                   docOnlyChange(target_branch) ||
                   (quickFunctional() &&
                    !paramsValue('CI_RPMS_leap15_TEST', true) &&
                    !run_default_skipped_stage('test-leap-15-rpms')) ||
                   (rpmTestVersion() != '') ||
                   stageAlreadyPassed()
```
or add some logic in the Jenkinsfile, or just remove the `!startedByLanding() ||` line and let it work like the Test RPMs on RHEL works today.

phender (changes requested):

vars/skipStage.groovy
daltonbohning:
```
            if (startedByLanding()) {
                return true
            }
```

So always skip for landing builds?

daltonbohning: I think this makes sense for now
```
            if (startedByLanding()) {
                return false
            }
```

ryon-jensen (approved):

ryon-jensen (approved):

phender (approved):

PR #512, opened 2026-04-30
SRE-3764: Revert 2.6 branch to leap15.5

ryon-jensen:


JohnMalmberg (approved):

vars/distroVersion.groovy
```diff
@@ -39,7 +39,7 @@ String call(String distro, String branch) {
                          '2.8':    '9.7'],
             'leap15':   ['master': '15.6',
                          '2.4':    '15.6',
-                         '2.6':    '15.6',
+                         '2.6':    '15.5',
```
phender, line 42: Even though we're moving back to building on 15.5, don't we still want to test on 15.6?  This code chooses the distro for provisioning stages like `Functional on Leap 15.6`.

phender (changes requested):
The distroVersion.groovy script is used to choose

vars/distroVersion.groovy
grom72: I do not know why do we want to push `2.6` to leap 15.6 if the biggest/most important user can not use it.

We can always run tests on leap 15.6 providing full name in the stage name as it is in case of `Functional on Leap 15.6`.

Without moving back to '15.5' here we have to put full information in [daos/Jenkins](https://github.com/daos-stack/daos/blob/a33e2a68e552d7d8a242c0485c937a20bf9562c1/Jenkinsfile#L556) that causes we have to modify again Git's checkers names to include minor version.

phender: Ok, this is for 2.6 only.

phender (approved):

PR #514, opened 2026-05-06
SRE-3773 build: Add Functional on SLES 15.7 to skipStage

ryon-jensen:


JohnMalmberg (approved):
Case statements allow string expressions that could simplify this to two cases that would not need updates in the future.

PR #487, opened 2025-12-13
DAOS-19247 build: Support running Bullseye

phender:
Use groovy script arguments to enable Bullseye code coverage.

PR #516, opened 2026-05-08
SRE-3704 ci: adapt NLT pipeline to cover Fault Injection testing procedure.

grom72:
This PR updates NLT and Fault Injection handling in `pipeline-lib` so stage behavior is
driven more consistently from parsed stage metadata and explicit step configuration.

Requires:
- [x] https://github.com/daos-stack/system-pipeline-lib/pull/23

With the required `system-pipeline-lib` change in place, `Fault Injection testing` can be
wired through the `unitTest` / `unitTestPost` procedures as follows:
```
    steps {
        job_step_update(
                            unitTest(timeout_time: 240,
                                     inst_repos: daosRepos(),
                                     test_script: 'ci/unit/test_nlt.sh --memcheck no' +
                                                  ' --system-ram-reserved 4 --server-debug WARN' +
                                                  ' --log-usage-import nltr.json' +
                                                  ' --log-usage-save nltr.xml' +
                                                  ' --class-name fault-injection fi',
                                     with_valgrind: '',
                                     always_script: 'ci/unit/test_nlt_post.sh',
                                     testResults: 'nlt-junit.xml',
                                     unstash_opt: true,
                                     unstash_tests: false,
                                     inst_rpms: unitPackages(target: 'el9') + ' daos-client-tests',
                                     image_version: 'el9.7',
                                     prov_env_vars: 'VM_CPUS=14'))
    }
    post {
        always {
                            unitTestPost artifacts: ['nlt_logs/'],
                                         testResults: 'nlt-junit.xml',
                                         with_valgrind: '',
                                         FI: true
            ...
        }
    }
```
unitTest.groovy:
- Expand the documented config surface to cover `prov_env_vars`, `testResults`,
  `with_valgrind`, `always_script`, and `valgrind_pattern` along with their defaults.
- Pass `prov_env_vars` through to the provisioning call so stage configuration can
  control provisioning environment variables explicitly.
- Make `with_valgrind` prefer the explicit step config value before falling back to the
  parsed stage metadata.
- Apply the same config-first precedence for `always_script`, `valgrind_pattern`, and
  `testResults` when preparing the `afterTest()` call.
- Remove the `NLT` fallback from the Valgrind error check so Valgrind processing runs only
  when `with_valgrind` is enabled.

unitTestPost.groovy:
- Make `valgrind_pattern` and `testResults` honor explicit config values before falling
  back to stage-derived defaults.
- Limit Valgrind post-processing to the `with_valgrind` setting instead of implicitly
  enabling it for all NLT stages.
- Add explicit `FI` handling and derive the effective `nlt` flag from `FI` or `NLT`.
- Replace the single `tool:` entry in `recordIssues` with a `tools:` list so Fault
  Injection can publish both `vm_test/nlt-errors.json` and `nlt-client-leaks.json`.
- Set the `recordIssues` name dynamically to `Fault injection testing` or `NLT` based on
  whether the stage is running FI-specific reporting.

skipStage.groovy:
- Cache `parseStageInfo()` in a local `stageInfo` map and reuse it for both `test_tag`
  lookup and target-specific decisions.
- Add `skip_build_on_el_gcc(...)` gating for `NLT on EL 8` / `NLT on EL 8.8` using the EL
  version derived from `stageInfo['target']`.
- Update `Fault injection testing` skip logic to use the same target-derived
  `skip_build_on_el_gcc(...)` check.
- Remove the extra `quickFunctional()`, `docOnlyChange(target_branch)`, and
  `skip_stage_pragma('func-test-vm')` checks from the Fault Injection skip path.

### DIFF
--- a/vars/distroVersion.groovy
+++ b/vars/distroVersion.groovy
@@ -39,7 +39,7 @@ String call(String distro, String branch) {
                          '2.8':    '9.7'],
             'leap15':   ['master': '15.6',
                          '2.4':    '15.6',
-                         '2.6':    '15.6',
+                         '2.6':    '15.5',
                          '2.8':    '15.6'],
             'sles15':   ['master': '15.7',
                          '2.4':    '15.7',
@@ -50,7 +50,7 @@ String call(String distro, String branch) {
 
 /* groovylint-disable-next-line CompileStatic */
 assert(call('leap15', '2.4') == '15.6')
-assert(call('leap15', '2.6') == '15.6')
+assert(call('leap15', '2.6') == '15.5')
 assert(call('leap15', 'master') == '15.6')
 assert(call('el8', '2.4') == '8.8')
 assert(call('el8', '2.6') == '8.8')
--- a/vars/distroVersion.groovy
+++ b/vars/distroVersion.groovy
@@ -30,6 +30,7 @@ String call(String distro) {
 }
 
 String call(String distro, String branch) {
+    println("Determining distro version for ${distro} and branch ${branch}")
     return ['el8':      ['master': '8.8',
                          '2.4':    '8.8',
                          '2.6':    '8.8',
--- a/vars/distroVersion.groovy
+++ b/vars/distroVersion.groovy
@@ -30,7 +30,6 @@ String call(String distro) {
 }
 
 String call(String distro, String branch) {
-    println("Determining distro version for ${distro} and branch ${branch}")
     return ['el8':      ['master': '8.8',
                          '2.4':    '8.8',
                          '2.6':    '8.8',
@@ -43,7 +42,6 @@ String call(String distro, String branch) {
                          '2.6':    '15.6',
                          '2.8':    '15.6'],
             'sles15':   ['master': '15.7',
-                         '2.4':    '15.7',
                          '2.6':    '15.7',
                          '2.8':    '15.7'],
             'ubuntu20': ['master': '20.04']][distro][branch]
--- a/vars/distroVersion.groovy
+++ b/vars/distroVersion.groovy
@@ -42,6 +42,7 @@ String call(String distro, String branch) {
                          '2.6':    '15.6',
                          '2.8':    '15.6'],
             'sles15':   ['master': '15.7',
+                         '2.4':    '15.7',
                          '2.6':    '15.7',
                          '2.8':    '15.7'],
             'ubuntu20': ['master': '20.04']][distro][branch]
--- a/vars/parseStageInfo.groovy
+++ b/vars/parseStageInfo.groovy
@@ -285,7 +285,12 @@ Map call(Map config = [:]) {
         result['ftest_arg'] = config['ftest_arg']
     }
 
-    if (stage_name.contains('NLT')) {
+    if (stage_name.toLowerCase().contains('nlt fault injection')) {
+        result['NLT'] = true
+        result['FI'] = true
+        result['always_script'] = config.get('always_script', 'ci/unit/test_nlt_post.sh')
+        result['testResults'] = config.get('testResults', 'nlt-junit.xml')
+    } else if (stage_name.contains('NLT')) {
         result['NLT'] = true
         result['valgrind_pattern'] = config.get('valgrind_pattern', '*memcheck.xml')
         result['always_script'] = config.get('always_script', 'ci/unit/test_nlt_post.sh')
--- a/vars/parseStageInfo.groovy
+++ b/vars/parseStageInfo.groovy
@@ -285,7 +285,7 @@ Map call(Map config = [:]) {
         result['ftest_arg'] = config['ftest_arg']
     }
 
-    if(stage_name.contains('NLT')) {
+    if (stage_name.contains('NLT')) {
         result['NLT'] = true
         result['always_script'] = config.get('always_script', 'ci/unit/test_nlt_post.sh')
         result['testResults'] = config.get('testResults', 'nlt-junit.xml')
--- a/vars/parseStageInfo.groovy
+++ b/vars/parseStageInfo.groovy
@@ -285,17 +285,16 @@ Map call(Map config = [:]) {
         result['ftest_arg'] = config['ftest_arg']
     }
 
-    if (stage_name.toLowerCase().contains('nlt fault injection')) {
+    if(stage_name.contains('NLT')) {
         result['NLT'] = true
-        result['FI'] = true
         result['always_script'] = config.get('always_script', 'ci/unit/test_nlt_post.sh')
         result['testResults'] = config.get('testResults', 'nlt-junit.xml')
-    } else if (stage_name.contains('NLT')) {
-        result['NLT'] = true
-        result['valgrind_pattern'] = config.get('valgrind_pattern', '*memcheck.xml')
-        result['always_script'] = config.get('always_script', 'ci/unit/test_nlt_post.sh')
-        result['testResults'] = config.get('testResults', 'nlt-junit.xml')
-        result['with_valgrind'] = 'memcheck'
+        if (stage_name.toLowerCase().contains('nlt fault injection')) {
+            result['FI'] = true
+        } else {
+            result['valgrind_pattern'] = config.get('valgrind_pattern', '*memcheck.xml')
+            result['with_valgrind'] = 'memcheck'
+        }
     } else {
         result['NLT'] = false
         if (config['valgrind_pattern']) {
--- a/vars/parseStageInfo.groovy
+++ b/vars/parseStageInfo.groovy
@@ -287,14 +287,10 @@ Map call(Map config = [:]) {
 
     if (stage_name.contains('NLT')) {
         result['NLT'] = true
+        result['valgrind_pattern'] = config.get('valgrind_pattern', '*memcheck.xml')
         result['always_script'] = config.get('always_script', 'ci/unit/test_nlt_post.sh')
         result['testResults'] = config.get('testResults', 'nlt-junit.xml')
-        if (stage_name.toLowerCase().contains('nlt fault injection')) {
-            result['FI'] = true
-        } else {
-            result['valgrind_pattern'] = config.get('valgrind_pattern', '*memcheck.xml')
-            result['with_valgrind'] = 'memcheck'
-        }
+        result['with_valgrind'] = 'memcheck'
     } else {
         result['NLT'] = false
         if (config['valgrind_pattern']) {
--- a/vars/rpmDistValue.groovy
+++ b/vars/rpmDistValue.groovy
@@ -21,7 +21,7 @@ String call(String distro) {
                distro.startsWith('rocky9') || distro.startsWith('almalinux9') ||
                distro.startsWith('rhel9')) {
         return '.el9'
-    } else if (distro.startsWith('leap15')) {
+    } else if (distro.startsWith('leap15') || distro.startsWith('sles15')) {
         return '.suse.lp' + parseStageInfo()['distro_version'].replaceAll('\\.', '')
     }
     error("Don't know what the RPM %{dist} is for ${distro}")
--- a/vars/skipStage.groovy
+++ b/vars/skipStage.groovy
@@ -434,6 +434,7 @@ boolean call(Map config = [:]) {
             skip_ftest('ubuntu20', target_branch, tags) */
             return true
         case 'Fault injection testing':
+        case 'NLT Fault injection testing':
         case 'Fault injection testing on CentOS 8':
         case 'Fault injection testing on EL 8':
         case 'Fault injection testing on EL 8.8':
--- a/vars/skipStage.groovy
+++ b/vars/skipStage.groovy
@@ -158,7 +158,8 @@ boolean call(Map config = [:]) {
     }
 
     String target_branch = env.CHANGE_TARGET ? env.CHANGE_TARGET : env.BRANCH_NAME
-    String tags = config['tags'] ?: parseStageInfo()['test_tag']
+    Map stageInfo = parseStageInfo()
+    String tags = config['tags'] ?: stageInfo['test_tag']
 
     switch (env.STAGE_NAME) {
         case 'Cancel Previous Builds':
@@ -353,6 +354,7 @@ boolean call(Map config = [:]) {
         case 'NLT on EL 8':
         case 'NLT on EL 8.8':
             return skip_stage_pragma('nlt') ||
+                   skip_build_on_el_gcc(target_branch, stageInfo['target'].replaceFirst('^el(\\d+).*', '$1')) ||
                    quickBuild() ||
                    stageAlreadyPassed()
         case 'Unit Test Bullseye':
@@ -441,10 +443,8 @@ boolean call(Map config = [:]) {
             return skip_stage_pragma('fault-injection-test') ||
                    !paramsValue('CI_FI_TEST', true) ||
                    !paramsValue('CI_FI_el8_TEST', true) || /* release/2.6 Jenkinsfile still uses CI_FI_el8_TEST */
-                   quickFunctional() ||
-                   docOnlyChange(target_branch) ||
+                   skip_build_on_el_gcc(target_branch, stageInfo['target'].replaceFirst('^el(\\d+).*', '$1')) ||
                    skip_stage_pragma('func-test') ||
-                   skip_stage_pragma('func-test-vm') ||
                    stageAlreadyPassed()
         case 'Test CentOS 7 RPMs':
             return !paramsValue('CI_RPMS_el7_TEST', true) ||
--- a/vars/skipStage.groovy
+++ b/vars/skipStage.groovy
@@ -436,7 +436,6 @@ boolean call(Map config = [:]) {
             skip_ftest('ubuntu20', target_branch, tags) */
             return true
         case 'Fault injection testing':
-        case 'NLT Fault injection testing':
         case 'Fault injection testing on CentOS 8':
         case 'Fault injection testing on EL 8':
         case 'Fault injection testing on EL 8.8':
--- a/vars/unitTest.groovy
+++ b/vars/unitTest.groovy
@@ -209,25 +209,32 @@ Map call(Map config = [:]) {
     p['testResults'] = stage_info.get('testResults', 'test_results/*.xml')
     p['with_valgrind'] = with_valgrind
     p['NLT'] = stage_info['NLT']
-    runTestData = afterTest(p, runData)
-    runTestData.each { resultKey, data -> runData[resultKey] = data }
+
+    // Update the stash inside a finally block so that ignore_failure is always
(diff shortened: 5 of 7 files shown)


daltonbohning (commented):
It appears this goes along with https://github.com/daos-stack/daos/pull/17953.

So we are effectively going to have different handling for this stage based on the release branch? Since the stage name is different, pipeline-lib will treat it differently

JohnMalmberg (dismissed):

vars/distroVersion.groovy
```diff
@@ -42,7 +42,6 @@ String call(String distro, String branch) {
                          '2.6':    '15.6',
                          '2.8':    '15.6'],
             'sles15':   ['master': '15.7',
-                         '2.4':    '15.7',
```
janekmi, line 45: This seems odd. Why only this line got removed?

vars/parseStageInfo.groovy
```diff
@@ -285,12 +285,16 @@ Map call(Map config = [:]) {
         result['ftest_arg'] = config['ftest_arg']
     }
 
-    if (stage_name.contains('NLT')) {
+    if(stage_name.contains('NLT')) {
```
phender: Personally, I think we should move away from relying on parseStageInfo to set arguments for other groovy functions based off of the stage name.  It would be better for this new stage to pass the arguments directly to `unitTest.groovy`.  I have an edit to `unitTest.groovy` in https://github.com/daos-stack/pipeline-lib/pull/487 that allows any of the parseStageInfo values to be overwritten with directly-passed arguments.  That might be useful here.

vars/unitTest.groovy
```diff
@@ -146,7 +148,8 @@ Map call(Map config = [:]) {
                  node_count: stage_info['node_count'],
                  distro: image_version,
                  inst_repos: config.get('inst_repos', ''),
-                 inst_rpms: inst_rpms)
+                 inst_rpms: inst_rpms,
+                 environment: stage_info['NLT']?"VM_CPUS=20":"")
```
phender: Seems like this would be better handled by a new config entry:
```suggestion
                 environment: config.get('prov_env_vars', ''))
```

vars/rpmDistValue.groovy
```diff
@@ -21,7 +21,7 @@ String call(String distro) {
                distro.startsWith('rocky9') || distro.startsWith('almalinux9') ||
                distro.startsWith('rhel9')) {
         return '.el9'
-    } else if (distro.startsWith('leap15')) {
+    } else if (distro.startsWith('leap15') || distro.startsWith('sles15')) {
```
janekmi, line 24: This seems unrelated.

vars/unitTestPost.groovy
```diff
@@ -22,9 +22,7 @@
    * config['valgrind_stash']      Name to stash Valgrind artifacts
    *                               Required if more than one stage is
    *                               creating Valgrind reports.
-   */
-
-// groovylint-disable DuplicateStringLiteral, VariableName
```
janekmi, line 27: 👍

phender (changes requested):

vars/unitTestPost.groovy
janekmi, line 135: One `ignore_failure` access omitted.

janekmi, line 166: One more omitted.

TBH there is a number of them. I am not sure whether to keep it this way or create an intermediary variable and put the default value there.

vars/unitTestPost.groovy
```diff
@@ -112,6 +110,17 @@ void call(Map config = [:]) {
...
+                                name: 'NLT errors',
+                                id: sanitizedStageName() + '_VM_test')]
+        if (stage_info['FI']) {
+            nltTools = [issues(pattern: 'vm_test/nlt-errors.json',
+                                name: 'Fault injection',
+                                id: sanitizedStageName() + '_VM_test')]
+            nltTools << issues(pattern: 'nlt-client-leaks.json',
+                               name: 'Fault injection leaks',
+                               id: 'NLT_client')
+        }
```
janekmi, line 131: 1. I do not like how the `nltTools` list get overwritten and
2. These two are exactly the same except for the `name` value.

```suggestion
        List nltTools = [issues(pattern: 'vm_test/nlt-errors.json',
                                name: stage_info['FI'] ? 'Fault injection' : 'NLT errors',
                                id: sanitizedStageName() + '_VM_test')]
        if (stage_info['FI']) {
            nltTools << issues(pattern: 'nlt-client-leaks.json',
                               name: 'Fault injection leaks',
                               id: 'NLT_client')
        }
```

vars/unitTestPost.groovy
```diff
@@ -142,9 +149,9 @@ void call(Map config = [:]) {
     }
 
     println results['result_code'].getClass()
-    println results['ignore_failure'].getClass()
+    println results.get('ignore_failure', false).getClass()
```
janekmi: Nitpick. These lines seem weirdly debug-y, aren't they? Are they any good on the user-side of things?

vars/unitTest.groovy
```diff
@@ -106,8 +106,10 @@ Map afterTest(Map config, Map testRunInfo) {
                                       flattenFiles: false,
                                       includes: valgrind_pattern,
                                       targetLocation: memcheck_dir)])
-        sh label: 'Create tarball of Valgrind xml files',
-           script: "tar -cjf ${memcheck_dir}.tar.bz2 ${memcheck_dir}"
+        if (fileExists(memcheck_dir)) {
```
janekmi: Out of curiosity. When might this condition not be met?

vars/unitTest.groovy
```diff
@@ -106,8 +106,10 @@ Map afterTest(Map config, Map testRunInfo) {
...
+            sh label: 'Create tarball of Valgrind xml files',
+               script: "tar -cjf ${memcheck_dir}.tar.bz2 ${memcheck_dir}"
+        }
```
janekmi: Should we print a message indicating the creation of a tarball got skipped?

vars/unitTest.groovy
janekmi: No need to employ GStrings.

```suggestion
                 environment: stage_info['NLT']?'VM_CPUS=20':'')
```

vars/parseStageInfo.groovy
JohnMalmberg: The purposes for ParseStageInfo originally:

1. Support use a Matrix format in Jenkinsfiles.   In that format, we could only pass the axis's of the matrix into the stage, and not any additional information.   The matrix format would simplify the structure of the Jenkinsfile, but we ran into some issues with use of a matrix, and have not checked recently Jenkins maintainers have fixed them.

2. To allow for quick cut/paste with minimal edits, and to eliminate human error that was happening, as many of our stages were identical except for information in that we would also want in the stage name.

vars/unitTest.groovy
```diff
@@ -208,26 +211,32 @@ Map call(Map config = [:]) {
                                            'unit-test-*memcheck.xml')
     p['testResults'] = stage_info.get('testResults', 'test_results/*.xml')
     p['with_valgrind'] = with_valgrind
-    p['NLT'] = stage_info['NLT']
-    runTestData = afterTest(p, runData)
-    runTestData.each { resultKey, data -> runData[resultKey] = data }
+
+    // Update the stash inside a finally block so that ignore_failure is always
+    // written even if afterTest() throws.  Without this, unitTestPost() reads
```
janekmi: Realistically, when `afterTest()` can actually throw an exception?

janekmi (changes requested):

vars/unitTest.groovy
grom72: Is it already included in https://github.com/daos-stack/pipeline-lib/pull/487?

vars/unitTest.groovy
grom72: No

vars/distroVersion.groovy
grom72: Done c64934d4bcf82c723a516c2c70b1c0931e1c085f

vars/rpmDistValue.groovy
grom72: PR does not work without this change.

vars/unitTest.groovy
grom72: I have the same idea in my mind, but do not want to extend this PR. I rename variable to `prov_env_vars`. 
c64934d4bcf82c723a516c2c70b1c0931e1c085f

vars/unitTestPost.groovy
grom72: Done c64934d4bcf82c723a516c2c70b1c0931e1c085f

vars/unitTest.groovy
grom72: Done c64934d4bcf82c723a516c2c70b1c0931e1c085f

vars/unitTest.groovy
janekmi: You probably already mentioned this offline but why? 🤔 

The test is configured as `config['with_valgrind']` so I would expect the tarball is there. Otherwise, it would be get a clear message that it was not created so we know everything works just fine.

vars/unitTest.groovy
janekmi: I assume you still test the approach without this check so I leave this comment for now. Again, it seems the directory ought to exist even though it may be empty, right?

vars/unitTestPost.groovy
```diff
@@ -22,9 +22,7 @@
    * config['valgrind_stash']      Name to stash Valgrind artifacts
    *                               Required if more than one stage is
    *                               creating Valgrind reports.
-   */
-
-// groovylint-disable DuplicateStringLiteral, VariableName
+   *
```
janekmi, line 25:
```suggestion
   */
```

janekmi (changes requested):

vars/parseStageInfo.groovy
grom72: @phender please take a look at the latest changes: https://github.com/daos-stack/pipeline-lib/pull/516/commits/fd71e7057a18935b146033375a8362ecc60ea7a4

vars/unitTest.groovy
grom72: Done: 4a52cb2a042586222dd4e32d1e58ecadbc19b271

vars/unitTestPost.groovy
grom72: Done 4d4107ec217c17a19eada6462c883bc3bcfef7d8

vars/unitTestPost.groovy
grom72: Done 4d4107ec217c17a19eada6462c883bc3bcfef7d8

grom72: Done 4d4107ec217c17a19eada6462c883bc3bcfef7d8

vars/unitTest.groovy
```diff
@@ -211,32 +215,26 @@ Map call(Map config = [:]) {
...
-        runData['ignore_failure'] = config.get('ignore_failure', false)
-        writeYaml file: results_map,
-                  data: runData,
-                  overwrite: true
-        stash name: results_map,
-              includes: results_map
-    }
+    p['NLT'] = stage_info['NLT']
+    runTestData = afterTest(p, runData)
+    runTestData.each { resultKey, data -> runData[resultKey] = data }
```
janekmi, line 239:
```suggestion
    runData.putAll(runTestData)
```

Ref: https://docs.groovy-lang.org/latest/html/groovy-jdk/java/util/Map.html#putAll(java.util.Collection)

vars/parseStageInfo.groovy
janekmi:
```suggestion
    if (stage_name.contains('NLT')) {
```

vars/unitTest.groovy
```diff
@@ -218,7 +225,7 @@ Map call(Map config = [:]) {
     }
     int runTime = durationSeconds(startDate)
     runData['unittest_time'] = runTime
-
+
```
janekmi: Please remove this redundant space.

janekmi (changes requested):

vars/unitTest.groovy
grom72: I did not change this line.

janekmi (dismissed):

vars/unitTest.groovy
```diff
@@ -187,7 +202,8 @@ Map call(Map config = [:]) {
                 outputFile: 'bullseye.tar'
     }
 
-    String with_valgrind = stage_info.get('with_valgrind', '')
+    String with_valgrind = config.get('with_valgrind',
+                                      stage_info.get('with_valgrind', ''))
```
janekmi: I don't like how you implemented the precedence of sources by reading `config` and `stage_info` in particular order multiple times.

It looks to me what you are doing here is establishing a subset of keys which behaves a little differently compared to what we had till now. I am not sure whether it is true for all keys in these maps so just to be on the safe side you can calculate a combined map containing only these subset of keys and use it instead of these nested `.get()` calls.

```suggestion
    // subset of keys for which config takes precedence over stage_info
    List config_precedence_keys = ['with_valgrind', 'always_script', 'valgrind_pattern', 'testResults']
    Map econfig = [:] << stage_info[config_precedence_keys] << config[config_precedence_keys]
    String with_valgrind = econfig.get('with_valgrind', ''))
```

If you are absolutely sure what you are doing you can go all the way and just call:

```groovy
    Map econfig = [:] << stage_info[] << config[]
```

**Note**: The second operand overwrites the first one.

vars/skipStage.groovy
```diff
@@ -434,16 +436,15 @@ boolean call(Map config = [:]) {
             skip_ftest('ubuntu20', target_branch, tags) */
             return true
         case 'Fault injection testing':
+        case 'NLT Fault injection testing':
```
phender: Do we need this?  https://github.com/daos-stack/daos/pull/17953 still uses 'Fault injection testing'.

vars/unitTestPost.groovy
```diff
@@ -125,10 +142,8 @@ void call(Map config = [:]) {
                        [threshold: 1, type: 'TOTAL_HIGH'],
                        [threshold: 1, type: 'NEW_NORMAL', unstable: true],
                        [threshold: 1, type: 'NEW_LOW', unstable: true]],
-                     name: 'Node local testing',
-                     tool: issues(pattern: 'vm_test/nlt-errors.json',
-                                  name: 'NLT results',
-                                  id: 'VM_test'),
+                     name: fi?'Fault injection':'NLT',
```
phender: Small suggestion to better match the stage names:
```suggestion
                     name: fi?'Fault injection testing':'NLT',
```

vars/skipStage.groovy
```diff
@@ -434,16 +436,15 @@ boolean call(Map config = [:]) {
...
         case 'Fault injection testing on CentOS 8':
         case 'Fault injection testing on EL 8':
         case 'Fault injection testing on EL 8.8':
             return skip_stage_pragma('fault-injection-test') ||
                    !paramsValue('CI_FI_TEST', true) ||
                    !paramsValue('CI_FI_el8_TEST', true) || /* release/2.6 Jenkinsfile still uses CI_FI_el8_TEST */
-                   quickFunctional() ||
-                   docOnlyChange(target_branch) ||
+                   skip_build_on_el_gcc(target_branch, stageInfo['target'].replaceFirst('^el(\\d+).*', '$1')) ||
```
phender, line 445: I just ran a PR w/ `Skip-test: true` and the current `Fault injection testing` was incorrectly run.  It looks like we also need this added:
```suggestion
                   skip_build_on_el_gcc(target_branch, stageInfo['target'].replaceFirst('^el(\\d+).*', '$1')) ||
                   skip_stage_pragma('test') ||
```

phender (changes requested):

vars/rpmDistValue.groovy
daltonbohning: This was since handled by #515

vars/skipStage.groovy
grom72: The problem is in the DAOS Jenkins file:
```
        stage('Test') {
            when {
                beforeAgent true
                //expression { !paramsValue('CI_FUNCTIONAL_TEST_SKIP', false)  && !skipStage() }
                // Above not working, always skipping functional VM tests.
                expression { !paramsValue('CI_FUNCTIONAL_TEST_SKIP', false) }
...
```
`Skip-test: true` pragma should skip the whole `Test` stage.
I would propose to leave `Fault injection testing` as it is and fix the `Test` stage skip in a separate PR [SRE-3783](https://daosio.atlassian.net/browse/SRE-3783).

vars/unitTestPost.groovy
grom72: Done in https://github.com/daos-stack/pipeline-lib/pull/516/commits/f96665dbda0a43ce45729956ab284d963a969bc8

vars/skipStage.groovy
grom72: Fixed in https://github.com/daos-stack/pipeline-lib/pull/516/commits/129f2d12b5d1784703fcf88e8b8b69815ea510c4

phender (approved):

phender (approved):

vars/unitTestPost.groovy
```diff
@@ -106,12 +113,22 @@ void call(Map config = [:]) {
...
                                   scm: 'daos-stack/daos',
                                   requiredResult: 'UNSTABLE')
+        List nltTools = [issues(pattern: 'vm_test/nlt-errors.json',
+                                name: fi ? 'Fault injection' : 'NLT errors',
+                                id: sanitizedStageName() + '_VM_test')]
+        if (fi) {
+            nltTools << issues(pattern: 'nlt-client-leaks.json',
+                               name: 'Fault injection leaks',
+                               id: 'NLT_client')
+        }
```
daltonbohning, line 131: Maybe I'm missing something, but when would we have both `fi == true` and `nlt == true`?

grom72: See line 117 (`Boolean nlt = fi || ...`), 
We can have `if(nlt || fi) {`, but the way it is implemented shows clearly that the Fault injection is a special type of NLT

daltonbohning: If NLT is a special type of fault injection, then why have
```
if (nlt) {
        ...
        if (fi) {
            nltTools << issues(pattern: 'nlt-client-leaks.json',
                               name: 'Fault injection leaks',
                               id: 'NLT_client')
        }
```

If NLT is always FI, then why do we need the `if (fi)`?

phender: NLT can be run w/o fault injection.  In fact, I need to run it this way in https://github.com/daos-stack/daos/pull/17269 with the https://github.com/daos-stack/pipeline-lib/pull/487 changes.

grom72: FI is a special type of NLT.

daltonbohning (approved):

PR #522, opened 2026-06-04
DAOS-18348 build: Simplify skipping stages

phender:
Predetermine which stages should be run by setting a Map to indicate which stages will be run at the beginning of the build. Update getFunctionalTestStage() to supoort being passed a single boolean to determine iif the Functional test stage should be run.

### DIFF
--- a/vars/unitPackages.groovy
+++ b/vars/unitPackages.groovy
@@ -26,16 +26,8 @@ String call(Map config = [:]) {
     Map stage_info = parseStageInfo(config)
     String target = stage_info['target']
 
-    boolean quick_build = quickBuild()
-
     if (target.startsWith('centos') || target.startsWith('el')) {
-        if (quick_build) {
-            // the script run below will read from this file
-            unstash target + '-required-mercury-rpm-version'
-        }
-
-        return sh(script: "${script} ${target} " +
-                          String.valueOf(quick_build),
+        return sh(script: "${script} ${target}",
                   returnStdout: true)
     }
     error 'unitPackages not implemented for ' + target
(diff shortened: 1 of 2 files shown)


vars/unitPackages.groovy
```diff
@@ -26,16 +26,8 @@ String call(Map config = [:]) {
     Map stage_info = parseStageInfo(config)
     String target = stage_info['target']
 
-    boolean quick_build = quickBuild()
-
     if (target.startsWith('centos') || target.startsWith('el')) {
-        if (quick_build) {
-            // the script run below will read from this file
-            unstash target + '-required-mercury-rpm-version'
-        }
```
daltonbohning, line 35: Why don't we need this anymore?

phender: In my testing with `Quick-functional: true` set in the commit message, if a new build was started by the user that ignored skip logic in the commit message (e.g. the new `CI_IGNORE_SKIP_COMMIT_PRAGMAS` build parameter was checked) the unit tests would fail trying to restore the `el9-required-mercury-rpm-version` stash (e.g. https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos/job/PR-18267/46/stages/?selected-node=455).  

The reason this code path was even followed was because `quickBuild()` references `quickFunctional()` in https://github.com/daos-stack/pipeline-lib/blob/master/vars/quickBuild.groovy#L15.  In this case the `Quick-functional: true` is being ignored and we are running the build stages, so the logic to get a stashed mercury RPM version isn't needed.  When the `Quick-functional: true` is used, we skip all of the unit tests, so this code isn't needed in that case either.

To fix the `el9-required-mercury-rpm-version` stash problem, there are two options:
- remove https://github.com/daos-stack/pipeline-lib/blob/master/vars/quickBuild.groovy#L15 or 
- remove the quick build logic in `unitPackages()`

I chose the latter to avoid breaking any other cases where other calls to `quickBuild()` may actually depend upon `quickFunctional()` and the deprecated message concerning `Quick-build` in https://daosio.atlassian.net/wiki/spaces/DC/pages/11891933554/Commit+Pragmas.

JohnMalmberg (dismissed):

daltonbohning (dismissed):

daltonbohning (approved):

JohnMalmberg (approved):

PR #524, opened 2026-06-16
SRE-3775: Support for Cluster Box

ryon-jensen:
- Add 'cb' to skipFunctionalTestStage logic
- Add optional node count

### DIFF
--- a/vars/getFunctionalTestStage.groovy
+++ b/vars/getFunctionalTestStage.groovy
@@ -25,6 +25,8 @@ import org.jenkinsci.plugins.pipeline.modeldefinition.Utils
  *                      rpmDistValue(distro)
  *      base_branch     if specified, checkout sources from this branch before running tests
  *      other_packages  space-separated string of additional RPM packages to install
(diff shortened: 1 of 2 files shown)


vars/getFunctionalTestStage.groovy
```diff
@@ -25,6 +25,8 @@ import org.jenkinsci.plugins.pipeline.modeldefinition.Utils
  *                      rpmDistValue(distro)
  *      base_branch     if specified, checkout sources from this branch before running tests
  *      other_packages  space-separated string of additional RPM packages to install
+ *      node_count      number of nodes to provision and use for the stage; overrides the count
+ *                      that would otherwise be inferred from the stage name by parseStageInfo()
```
JohnMalmberg, line 29: I thought we already had a way of doing so we can run medium jobs on a large cluster so not sure what this is for.

JohnMalmberg (approved):
Looks good to me, not sure the node count override is needed.

PR #526, opened 2026-07-01
DAOS-19247 test: Support scripted Test stages

phender:
Adding the scriptedTestRpmStage() method to provide a scripted stage definition for the Test RPM
stages and the scriptedUnitTestStage() method to provide a scripted stage definition for the Fault
Injection Testing stage.  Fixing skipping Functional test stages when builds are re-run and the stage
previously passed.


vars/scriptedDockerStage.groovy
daltonbohning, line 1: It seems very pedantic to me to name these functions `scripted*`.
Also, this seems to be specifically for building and not just generally a "docker" stage.

daltonbohning: I don't see this being used in https://github.com/daos-stack/daos/pull/18597 ?

vars/scriptedDockerStage.groovy
```diff
@@ -0,0 +1,142 @@
...
+Map call(Map kwargs = [:]) {
+    String name = kwargs.get('name', 'Unknown Docker Stage')
+    Boolean runStage = kwargs.get('runStage', true)
+    Map jobStatus = kwargs.get('jobStatus', [:])
+    String dockerTag = jobStatusKey("build-${uploadTarget}-${compiler}").toLowerCase()
+    String dockerBuildArgs = kwargs.get('dockerBuildArgs', '')
+
+    // Build stage parameters
+    Boolean buildRpms = kwargs.get('buildRpms', true)
+    String distro = kwargs.get('distro', 'el8')
```
daltonbohning: Maybe we don't need a default distro here? In most cases we should probably specify it so get what we want. Especially for building rpms

vars/scriptedDockerStage.groovy
```diff
@@ -0,0 +1,142 @@
...
+            node('docker_runner') {
+                println("[${name}] Check out from version control")
+                checkoutScm(pruneStaleBranch: true)
+
+                def dockerImage = docker.build(dockerTag, dockerBuildArgs)
+                try {
+                    dockerImage.inside() {
+                        if (buildRpms) {
+                            sh label: 'Install RPMs',
+                                script: "./ci/rpm/install_deps.sh ${rpmDistro} ${release} ${bullseye}"
```
daltonbohning: IMO this is going backwards in architecture and obfuscating the flow. Currently, the Jenkinsfile references `./ci/rpm/install_deps.sh`, which exists in the DAOS repo. Now we are hardcodiing this script reference into pipeline-lib, when the script is not defined in pipeline-lib. So this is creating a circular code dependency. And similar for all the other `.ci/*` references.

daltonbohning: Tying back to my first comment about this not being a generic docker stage - this entire function is not really a "library" function. It's the exact stage definition that the DAOS repo wants to run, and it's hardcoded into this separate repository.

vars/scriptedDockerStage.groovy
phender: That script has been removed and moved to https://github.com/daos-stack/pipeline-lib/pull/528 where it will be tested.

daltonbohning (dismissed):

vars/scriptedTestRpmStage.groovy
```diff
@@ -0,0 +1,102 @@
...
+import org.jenkinsci.plugins.pipeline.modeldefinition.Utils
+
+/**
+ * scriptedTestRpmStage.groovy
+ *
+ * Get a test stage in scripted syntax.
+ *
+ * @param kwargs Map containing the following optional arguments (empty strings yield defaults):
+ *      name                  test stage name
+ *      runStage              whether or not to run the test stage
```
grom72:
```suggestion
 *      runStage              whether or not to run the test stage; defaults to true
```

vars/scriptedTestRpmStage.groovy
```diff
@@ -0,0 +1,102 @@
...
+ *      label                 test stage default cluster label
+ *      testBranch            if specified, checkout sources from this branch before running tests
```
grom72:
```suggestion
 *      testBranch            if specified, checkout sources from this branch before running tests; defaults to ''
```

vars/scriptedUnitTestStage.groovy
grom72, line 1: Can we move this file to e separate PR where we can have also Jenkinsfile in related PR that uses this groovy script?

grom72 (changes requested):

vars/scriptedUnitTestStage.groovy
phender: I'm not sure what you're asking.  The `scriptedUnitTestStage.groovy()` is called in https://github.com/daos-stack/daos/pull/18597 at https://github.com/daos-stack/daos/blob/7b2f7e440949b3b4d60000e68e20e67a8d3cd27d/Jenkinsfile#L1259
```
                        'Fault injection testing': scriptedUnitTestStage(
                            name: 'Fault injection testing',
                            runStage: shouldStageRun('Fault injection testing'),
                            ...
```

vars/scriptedTestRpmStage.groovy
phender: Fixed in https://github.com/daos-stack/pipeline-lib/pull/526/changes/8cf3e29e93628eac27406dde2a8b17fa034a454a

vars/scriptedTestRpmStage.groovy
phender: Fixed in https://github.com/daos-stack/pipeline-lib/pull/526/changes/8cf3e29e93628eac27406dde2a8b17fa034a454a

daltonbohning (dismissed):

vars/scriptedUnitTestStage.groovy
```diff
@@ -0,0 +1,118 @@
...
+    String distro = kwargs.get('distro', '')
+
+    // Unit Test stage parameters
+    Map unitTestArgs = [
+        'timeout_time': kwargs.get('timeoutTime', 60),
+        'inst_repos': kwargs.get('instRepos', daosRepos(distro)),
+        'test_script': kwargs.get('testScript', ''),
+        'with_valgrind': kwargs.get('withValgrind', ''),
+        'always_script': kwargs.get('alwaysScript', ''),
+        'test_results': kwargs.get('testResults', 'nlt-junit.xml'),
```
grom72: Isn't it the case that `unitTest` expects `testResults` instead of `test_results`?
https://github.com/daos-stack/pipeline-lib/blob/cb6b60244bae5758da07e652f09e7c537a84111c/vars/unitTest.groovy#L234
```suggestion
        'testResults': kwargs.get('testResults', 'nlt-junit.xml'),
```

vars/scriptedUnitTestStage.groovy
```diff
@@ -0,0 +1,118 @@
...
+ *      testResults           unitTest() test_results argument; defaults to 'nlt-junit.xml'
+ *      unstashOpt            unitTest() unstash_opt argument; defaults to true
+ *      unstashTests          unitTest() unstash_tests argument; defaults to false
+ *      instRpms              unitTest() inst_rpms argument; defaults to
+ *                              unitPackages(target: distro) + ' daos-client-tests'
+ *      imageVersion          unitTest() image_version argument, e.g 'el9.7'; defaults to ''
+ *      provEnvVars           unitTest() prov_env_vars argument; defaults to ''
+ *      unitTestPostArgs      Map of arguments to pass to unitTestPost() for the stage
+ *      archiveArtifactsArgs  Map of arguments to pass to archiveArtifacts() for the stage
+ * @return a scripted stage to run in a pipeline
```
grom72, line 24: Shouldn't we let the unitTest() function determine the default settings?

vars/scriptedUnitTestStage.groovy
grom72: I see.

vars/scriptedUnitTestStage.groovy
```diff
@@ -0,0 +1,118 @@
...
+ */
+Map call(Map kwargs = [:]) {
+    // General parameters
+    String name = kwargs.get('name', 'Unknown Unit Test Stage')
```
grom72: Should we have name as mandatory parameter and fail fast if not provided?

vars/scriptedUnitTestStage.groovy
```diff
@@ -0,0 +1,118 @@
...
+        'test_script': kwargs.get('testScript', ''),
+        'with_valgrind': kwargs.get('withValgrind', ''),
+        'always_script': kwargs.get('alwaysScript', ''),
+        'test_results': kwargs.get('testResults', 'nlt-junit.xml'),
+        'unstash_opt': kwargs.get('unstashOpt', true),
+        'unstash_tests': kwargs.get('unstashTests', false),
+        'inst_rpms': kwargs.get('instRpms', unitPackages(target: distro) + ' daos-client-tests'),
+        'image_version': kwargs.get('imageVersion', ''),
+        'prov_env_vars': kwargs.get('provEnvVars', '')
+    ]
```
grom72: If we have same names here as in unitTest() we can just iterate over kwargs to create the map automatically:
```suggestion
Set wrapperKeys = [
    'name',
    'runStage',
    'label',
    'testBranch',
    'jobStatus',
    'distro',
    'unitTestPostArgs',
    'archiveArtifactsArgs'
] as Set

Map unitTestArgs = new LinkedHashMap(kwargs)
unitTestArgs.keySet().removeAll(wrapperKeys)
```

If you want to introduce new names the following structure can be easier to read:
```
Map keyMap = [
    timeoutTime : 'timeout_time',
    instRepos   : 'inst_repos',
    testScript  : 'test_script',
    withValgrind: 'with_valgrind',
    alwaysScript: 'always_script',
    testResults : 'test_results',
    unstashOpt  : 'unstash_opt',
    unstashTests: 'unstash_tests',
    instRpms    : 'inst_rpms',
    imageVersion: 'image_version',
    provEnvVars : 'prov_env_vars'
]

Map unitTestArgs = [:]
keyMap.each { inKey, outKey ->
    if (kwargs.containsKey(inKey)) {
        unitTestArgs[outKey] = kwargs[inKey]
    }
}
```

vars/scriptedUnitTestStage.groovy
grom72: Is it a proper default?
The primary goal of this function will be to handle unit tests. NLT and FI is a special case.

grom72 (changes requested):

vars/scriptedUnitTestStage.groovy
daltonbohning: +1

vars/scriptedUnitTestStage.groovy
phender: Thanks for the suggestion. Fixed in https://github.com/daos-stack/pipeline-lib/pull/526/changes/037d0ddc14145bc96d08751797e10cc4b86a0d2a..0162051eb3f92cb52c3abeabeb3404c1c389b1ed.  The arguments for `unitTest.groovy` are now passed through.

vars/scriptedUnitTestStage.groovy
phender: Yes, see https://github.com/daos-stack/pipeline-lib/pull/526/changes/037d0ddc14145bc96d08751797e10cc4b86a0d2a..0162051eb3f92cb52c3abeabeb3404c1c389b1ed

vars/scriptedUnitTestStage.groovy
phender: Good suggestion.  I've applied it to all three groovy scripts in https://github.com/daos-stack/pipeline-lib/pull/526/changes/037d0ddc14145bc96d08751797e10cc4b86a0d2a..0162051eb3f92cb52c3abeabeb3404c1c389b1ed

vars/scriptedUnitTestStage.groovy
phender: Thanks, I've just updated the code to pass the `unitTest.groovy` arguments through in https://github.com/daos-stack/pipeline-lib/pull/526/changes/037d0ddc14145bc96d08751797e10cc4b86a0d2a..0162051eb3f92cb52c3abeabeb3404c1c389b1ed w/o any renaming.

vars/scriptedUnitTestStage.groovy
phender: We'll now just use the `unitTest.groovy` default in https://github.com/daos-stack/pipeline-lib/pull/526/changes/037d0ddc14145bc96d08751797e10cc4b86a0d2a..0162051eb3f92cb52c3abeabeb3404c1c389b1ed

vars/scriptedUnitTestStage.groovy
```diff
@@ -0,0 +1,109 @@
...
+ * @param kwargs Map containing the following optional arguments (empty strings yield defaults):
+ *      name                  test stage name
+ *      runStage              whether or not to run the test stage
+ *      label                 test stage default cluster label
+ *      testBranch            if specified, checkout sources from this branch before running tests
+ *      jobStatus             Map of status for each stage in the job/build
+ *      distro                the distro to use for the test stage, e.g. 'el9'; defaults to ''
+ *      unitTestArgs          Map of arguments to pass to unitTest; defaults to an empty Map.
+ *      unitTestPostArgs      Map of arguments to pass to unitTestPost() for the stage
+ *      archiveArtifactsArgs  Map of arguments to pass to archiveArtifacts() for the stage
```
grom72:
```suggestion
 *      unitTestArgs          Map of arguments to pass to unitTest; defaults to an empty Map.
 *      unitTestPostArgs      Map of arguments to pass to unitTestPost() for the stage; defaults to an empty Map.
 *      archiveArtifactsArgs  Map of arguments to pass to archiveArtifacts() for the stage; defaults to an empty Map.
```

vars/scriptedUnitTestStage.groovy
```diff
@@ -0,0 +1,109 @@
...
+
+            // Add defaults for any missing unitTest() arguments
+            if (!unitTestArgs.containsKey('inst_repos')) {
+                /* groovylint-disable-next-line DuplicateStringLiteral */
+                unitTestArgs['inst_repos'] = daosRepos(distro)
+            }
+            if (!unitTestArgs.containsKey('inst_rpms')) {
+                /* groovylint-disable-next-line DuplicateStringLiteral */
+                unitTestArgs['inst_rpms'] = unitPackages(target: distro) + ' daos-client-tests'
+            }
```
grom72, line 60: Would it be better to fix `unitTest()` in `unitTest.groovy`?

grom72, line 60: Do we have DAOS PR that removes this default from Jenkinsfile?
```                            job_step_update(
                                unitTest(timeout_time: 60,
                                        unstash_opt: true,
                                        // inst_repos: daosRepos(),
                                        // inst_rpms: unitPackages(target: 'el9'),
                                        image_version: 'el9.7',
                                        )
                            )

Comment (grom72) [inline vars/scriptedUnitTestStage.groovy:60]: Why do we add `' daos-client-test'`?
The actual Jenkins file does not require it:
```
                stage('Unit Test') {
...
                    steps {
                            job_step_update(
                                unitTest(timeout_time: 60,
                                        unstash_opt: true,
                                        inst_repos: daosRepos(),
                                        inst_rpms: unitPackages(target: 'el9'),
                                        image_version: 'el9.7',
                                        )
                            )
                    }
...
                }
```
```suggestion
            if (!unitTestArgs.containsKey('inst_rpms')) {
                /* groovylint-disable-next-line DuplicateStringLiteral */
                unitTestArgs['inst_rpms'] = unitPackages(target: distro)
            }
```

vars/scriptedUnitTestStage.groovy
```diff
@@ -0,0 +1,109 @@
...
+            if (!runStage) {
+                println("[${name}] Stage skipped by runStage=false")
+                Utils.markStageSkippedForConditional("${name}")
+                return
+            }
+
+            // Add defaults for any missing unitTest() arguments
+            if (!unitTestArgs.containsKey('inst_repos')) {
+                /* groovylint-disable-next-line DuplicateStringLiteral */
+                unitTestArgs['inst_repos'] = daosRepos(distro)
```
grom72, line 55: Do we have a DAOS PR that removes these defaults from Jenkinsfile?
```                            job_step_update(
                                unitTest(timeout_time: 60,
                                        unstash_opt: true,
                                        // inst_repos: daosRepos(),
                                        // inst_rpms: unitPackages(target: 'el9'),
                                        image_version: 'el9.7',
                                        )
                            )

**Review by grom72 — CHANGES_REQUESTED** [review]

File: vars/scriptedUnitTestStage.groovy (hunk 10 above)
Comment (phender) [reply vars/scriptedUnitTestStage.groovy:60]: This PR is a prerequisite for the https://github.com/daos-stack/daos/pull/18597 where the `scriptedUnitTestStage()` method is being used to run the `Fault Injection Testing stage`.  In this case we want to call the `daosRepos()` and `unitPackages()` methods from inside the stage - not when defining arguments for a stage in the Jenkinsfile.  As such they have been assigned here as defaults if no `inst_repos` or `inst_rpms` arguments are provided.

If in the future we decide to use this for other unit test stages I think we can investigate other options for providing `daosRepos()` as an input for the `unit_test` argument for the `unitTest()` method w/o calling it outside of the stage - which may include passing more arguments to the method and other methods it calls.  It is a bit beyond the scope of this change in my opinion.

Comment (phender) [reply vars/scriptedUnitTestStage.groovy:60]: No, the change is for https://github.com/daos-stack/daos/pull/18597 and the `Fault Injection Testing` stage.

File: vars/scriptedUnitTestStage.groovy (hunk 11 above)
Comment (phender) [reply vars/scriptedUnitTestStage.groovy:55]: No, the change is for https://github.com/daos-stack/daos/pull/18597 and the `Fault Injection Testing` stage.

File: vars/scriptedUnitTestStage.groovy (hunk 10 above)
Comment (phender) [reply vars/scriptedUnitTestStage.groovy:60]: This is needed for the `Fault Injection Testing` stage, where `scriptedUnitTestStage()` is being used.

File: vars/scriptedUnitTestStage.groovy (hunk 9 above)
Comment (phender) [reply vars/scriptedUnitTestStage.groovy]: Fixed in https://github.com/daos-stack/pipeline-lib/pull/526/changes/8130ede3c1904afd9982beb4a56f2ec33b096f59

File: vars/scriptedUnitTestStage.groovy (hunk 10 above)
Comment (grom72) [reply vars/scriptedUnitTestStage.groovy:60]: The problem I see here (and in previous comments) is that we start using the `scriptedUnitTestStage()` method not for default unit test behavior but for an exceptional.
If Fault Injection requires some parameters different than default ones (e.g. additional `daos-client-tests`) they should be provided in Jenkinsfile (or via different wrapper function).

**Review by JohnMalmberg — DISMISSED** [review]

**Review by JohnMalmberg — APPROVED** [review]

File: vars/getFunctionalTestStage.groovy [hunk 12]
```diff
@@ -33,8 +33,9 @@ import org.jenkinsci.plugins.pipeline.modeldefinition.Utils
  *      job_status      Map of status for each stage in the job/build
  * @return a scripted stage to run in a pipeline
  */
+/* groovylint-disable-next-line MethodSize */
 Map call(Map kwargs = [:]) {
-    String name = kwargs.get('name', 'Unknown Functional Test Stage')
+    String name = kwargs.get('name', '')
```
daltonbohning, line 38: Why empty string instead of null?

grom72 (approved):

vars/scriptedTestRpmStage.groovy
```diff
@@ -0,0 +1,113 @@
...
+ *                              testRpmArgs does not specify a daos_pkg_version; defaults to null
+ * @return a scripted stage to run in a pipeline
+ */
+Map call(Map kwargs = [:]) {
+    // General parameters
+    String name = kwargs.get('name', '')
+    Boolean runStage = kwargs.get('runStage', true) as Boolean
+    String label = kwargs.get('label')
+    String testBranch = kwargs.get('testBranch', '')
+    Map jobStatus = kwargs.get('jobStatus', null) ?: [:]
```
daltonbohning, line 34: Is this not the same?
```suggestion
    Map jobStatus = kwargs.get('jobStatus', [:])
```

daltonbohning (approved):

vars/scriptedUnitTestStage.groovy
daltonbohning: I do agree with @grom72 here. IMO it's better to not hardcode this kind of thing into pipeline-lib

vars/scriptedTestRpmStage.groovy
phender: Not exactly:
```
def kwargs = [jobStatus: false]

// #1 Returns [:]    (same result for null, 0, or '' jobStatus value or an missing jobStatus key)
kwargs.get('jobStatus', null) ?: [:]

// #2 Returns false  ([:] is only returned if the 'jobStatus' key is not assigned in kwargs, regardless of its value)
kwargs.get('jobStatus', [:])
```

daltonbohning: Why would we have `[jobStatus: false]` if it is supposed to be a map? This seems like we are masking other issues?

vars/scriptedUnitTestStage.groovy
phender: The tricky part is that methods like `unitPackages()` rely on `parseStageInfo()` and `parseStageInfo` relies on the environment variables like the stage name being set.  Since this method returns the stage definition to be run, the stage name is not yet set to the stage we're going to run, so the return value of `unitPackages()` cannot be passed as an input - in its current form.  We would essentially need to replace all the `parseStageInfo()` outputs with required inputs into a bunch of groovy methods.  Then we could call:
```
scriptedUnitTestStage(
   ...
   inst_rpms: unitPackages(target: distro, script: 'ci/unit/required_packages.sh') + ' daos-client-tests',
   ....
)
```
This is something I think is better tackled in another PR - one where we could also consider using `scriptedUnitTestStage()` for the other stages that use `unitTest()`.

PR #528, opened 2026-07-08
DAOS-19247 build: Support running Bullseye

phender:
In order to support Bullseye code coverage stages, provide additional groovy function parameters including options to override values obtained through the parseStageInfo script.

### DIFF
--- a/vars/buildRpmPost.groovy
+++ b/vars/buildRpmPost.groovy
@@ -49,8 +49,9 @@
    *
    * config['rpmlint']             Whether to run rpmlint on resulting RPMs.
    *                               Default false.
-   * config['new_rpm']             Whether we are using new RPM or not
-   *                               Default false
+   *
+   * config['productArtifacts']    List of product names and artifact directories to publish.
+   *                               Default is to publish the whole target directory.
    *
    * config['unsuccessful_script'] Script to run if build is not successful.
    *                               Default 'ci/rpm/build_unsuccessful.sh'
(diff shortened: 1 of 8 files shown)


vars/buildRpmPost.groovy
```diff
@@ -89,24 +95,37 @@ void call(Map config = [:]) {
...
+            productNames = ['daos', 'deps']
+        }
+
         String product = config.get('product', 'daos-stack')
-        String artdir = 'artifacts/' + target
-        if (config.get('new_rpm', false)) {
-            String deps_dir = 'artifacts/' + target + '/deps'
-            if (fileExists(deps_dir)) {
-                publishToRepository product: 'deps',
+        Map<String, String> productArtifacts = [:]
```
daltonbohning, line 105: So `config['productArtifacts']` is a list but we are also defining this local variable of the same name, but it is a map?
Should we change `config['productArtifacts'] -> config['productNames']` so `productArtifacts` is not overloaded?

vars/buildRpmPost.groovy
```diff
@@ -89,24 +95,37 @@ void call(Map config = [:]) {
                   includes: rpm_version_file
         }
 
+        List<String> productNames = config.get('productArtifacts', [])
```
daltonbohning: This also suggests to me that it should be `config['productNames']`

vars/scriptedDockerStage.groovy
```diff
@@ -0,0 +1,136 @@
...
+ *          buildRpmPostArgs        optional arguments to pass to buildRpmPost(); defaults to [:]
+ *          archiveArtifactsArgs    optional arguments to pass to archiveArtifacts()
+ * @return a scripted stage to run in a pipeline
+ */
+/* groovylint-disable-next-line MethodSize */
+Map call(Map kwargs = [:]) {
+    String name = kwargs.get('name', '')
+    Boolean runStage = kwargs.get('runStage', true)
+    Map jobStatus = kwargs.get('jobStatus', null) ?: [:]
+    String dockerTag = kwargs.get('dockerTag', 'unknown-docker-tag')
```
daltonbohning: Should this instead error? Is it even valid to pass `unknown-docker-tag`?

vars/scriptedDockerStage.groovy
```diff
@@ -0,0 +1,136 @@
...
+ *          configLog               optional config.log name to archive upon exception
+ *          valgrindSconsBuildArgs  optional scons build arguments for valgrind build
+ *          generateRpmsScript      optional script to run to generate rpms; defaults to ''
+ *          buildRpmPostArgs        optional arguments to pass to buildRpmPost(); defaults to [:]
+ *          archiveArtifactsArgs    optional arguments to pass to archiveArtifacts()
+ * @return a scripted stage to run in a pipeline
+ */
+/* groovylint-disable-next-line MethodSize */
+Map call(Map kwargs = [:]) {
+    String name = kwargs.get('name', '')
```
daltonbohning: For string params, why use empty string instead of null?

vars/scriptedDockerStage.groovy
```diff
@@ -0,0 +1,136 @@
...
+                        }
+                        if (generateRpmsScript) {
+                            println("[${name}] Running generateRpmsScript")
+                            sh label: 'Generate RPMs',
+                                script: "${generateRpmsScript}"
+                        }
+                        if (buildRpmPostArgs) {
+                            println("[${name}] Running buildRpmPost()")
+                            buildRpmPost(buildRpmPostArgs)
+                        }
```
daltonbohning: IMO this kind of logic means this is not a "docker" stage. But specifically it is a "build DAOS in docker" stage. And it's not really a library function but more so the entire definition of the stage, which seems more appropriate in the daos repo

daltonbohning (dismissed):

vars/buildRpmPost.groovy
phender: Addressed in https://github.com/daos-stack/pipeline-lib/pull/528/changes/206dda97cfd06612f32d5ae10ca674d2a6988571

PR #529, opened 2026-07-13
SRE-3899 Add SCM checkout retry on git fetch failure

ryon-jensen:
Wrap the checkout() call in checkoutScmInternal with a retry loop to automatically recover from transient GitHub connectivity failures (fetch-pack: invalid index-pack output / git-remote-https signal 15).

Retry count is controlled by, in priority order:
  1. config['checkoutRetries'] passed by the caller
  2. SCM_CHECKOUT_RETRY_COUNT Jenkins environment variable
  3. Default of 1 (no retries, preserving existing behaviour)

Set SCM_CHECKOUT_RETRY_COUNT=3 in Jenkins global env to enable retries across all jobs without a further code change.

### DIFF
--- a/src/com/intel/checkoutScmInternal.groovy
+++ b/src/com/intel/checkoutScmInternal.groovy
@@ -17,10 +17,11 @@ Map checkoutScmInternal(Map config = [:]) {
    * @return Map of any variables the SCM plugin would set in a
    * Freestyle job.
    *
-   * config['scm'] SCM name, defaults to 'GitSCM'.
-   * config['url'] Url for SCM repository.  If not specified,
-   *  defaults will be used from the scm global variable.
-   * config['cleanAfterCheckout'] True for clean after checkout.
+    * config['scm'] SCM name, defaults to 'GitSCM'.
+    * config['url'] Url for SCM repository.  If not specified,
+    *  defaults will be used from the scm global variable.
+    * config['checkoutRetries'] Retry count override. Defaults to $SCM_CHECKOUT_RETRY_COUNT env var, then 1.
+    * config['cleanAfterCheckout'] True for clean after checkout.
    * config['checkoutDir'] Optional directory to checkout into.
    * config['branch'] Optional branch to checkout, defaults to master.
    * config['withSubmodules'] Optional checkout submodules if true.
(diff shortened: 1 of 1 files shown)


src/com/intel/checkoutScmInternal.groovy
```diff
@@ -17,10 +17,11 @@ Map checkoutScmInternal(Map config = [:]) {
...
    * Freestyle job.
    *
-   * config['scm'] SCM name, defaults to 'GitSCM'.
-   * config['url'] Url for SCM repository.  If not specified,
-   *  defaults will be used from the scm global variable.
-   * config['cleanAfterCheckout'] True for clean after checkout.
+    * config['scm'] SCM name, defaults to 'GitSCM'.
+    * config['url'] Url for SCM repository.  If not specified,
+    *  defaults will be used from the scm global variable.
+    * config['checkoutRetries'] Retry count override. Defaults to $SCM_CHECKOUT_RETRY_COUNT env var, then 1.
```
grom72:
```suggestion
    * config['checkoutRetries'] Retry count override. If not specifed
    *  defaults to $SCM_CHECKOUT_RETRY_COUNT env var, then 1.
```

src/com/intel/checkoutScmInternal.groovy
```diff
@@ -17,10 +17,11 @@ Map checkoutScmInternal(Map config = [:]) {
...
+    * config['cleanAfterCheckout'] True for clean after checkout.
```
grom72: Can you please fix the next three lines of comment (identations)?

src/com/intel/checkoutScmInternal.groovy
```diff
@@ -92,5 +93,18 @@ Map checkoutScmInternal(Map config = [:]) {
            trackingSubmodules: false])
     }
 
-    return checkout(params)
+    int maxRetries = (config['checkoutRetries'] != null
+                      ? config['checkoutRetries']
+                      : (env.SCM_CHECKOUT_RETRY_COUNT ?: '1').toInteger())
```
grom72: I thing we should ignore 0 as invalid value and use default in this case.
This gives possibility to use shorter form of assignment.
```suggestion
    int maxRetries = config.checkoutRetries ?: (env.SCM_CHECKOUT_RETRY_COUNT ?: '1').toInteger()
```

grom72 (changes requested):
Do we have any PR in DAOS to check how this will behave?
Especially the retry path?

src/com/intel/checkoutScmInternal.groovy
ryon-jensen: Fixed

src/com/intel/checkoutScmInternal.groovy
ryon-jensen: Updated

src/com/intel/checkoutScmInternal.groovy
ryon-jensen: Good idea. I'll make sure it's not 0.

